PR #23, opened 2020-05-21
update to use new fips compliant images

briandowns:
Replaces default Kubernetes and EtcD registries and repositories with the FIPS 140-2 compliant images.


ibuildthecloud:
What are the other changes in this PR? Are they just cosmetic or did something change?

briandowns:
> What are the other changes in this PR? Are they just cosmetic or did something change?

Some cosmetic. I added rke2 to the .gitignore since it complied and left a bin at the root. The string concat reads cleaner and removes a few allocations and reflection from Sprintf call.

briandowns:
@ibuildthecloud are we good to go for this one?

PR #31, opened 2020-06-12
CIS 1.5 compliance

briandowns:
Update component arguments to be in line with CIS 1.5 Benchmarks and the Rancher Hardening Guide.


pkg/podexecutor/staticpod.go
```diff
@@ -64,7 +64,13 @@ func (s *StaticPod) APIServer(ctx context.Context, etcdReady <-chan struct{}, ar
 	if err := images.Pull(s.PullImages, "kube-apiserver", s.Images.KubeAPIServer); err != nil {
 		return nil, nil, err
 	}
-
+	args = append(args,
+		`--basic-auth-file=""`,
```
dweomer, line 68: `basic-auth-file` needs to be explicitly set to empty? if so, please comment as to why

pkg/podexecutor/staticpod.go
```diff
@@ -121,7 +127,9 @@ func (s *StaticPod) ControllerManager(apiReady <-chan struct{}, args []string) e
 		return staticpod.Run(s.Manifests, staticpod.Args{
 			Command: "kube-controller-manager",
 			Args: append(args,
-				"/usr/libexec/kubernetes/kubelet-plugins/volume/exec"),
+				"--flex-volume-plugin-dir=/usr/libexec/kubernetes/kubelet-plugins/volume/exec",
```
dweomer, line 130: this will break on systems (such as k3OS) with read-only `/usr` but I think this only emits a warning in the logs (over and over again).

briandowns: This changes is a bug fix to add the argument that was missing. This may need a larger conversation to address this concern.

pkg/podexecutor/staticpod.go
briandowns: Updated. Thanks!

pkg/podexecutor/staticpod.go
```diff
@@ -64,7 +64,13 @@ func (s *StaticPod) APIServer(ctx context.Context, etcdReady <-chan struct{}, ar
 	if err := images.Pull(s.PullImages, "kube-apiserver", s.Images.KubeAPIServer); err != nil {
 		return nil, nil, err
 	}
-
+	args = append(args,
+		`--basic-auth-file=""`, // explicitly setting to empty value per CIS 1.5.
```
galal-hussein, line 68: this will break how rke2 works i think, if i understand correctly rke2 process uses this basic auth file generated by k3s

briandowns: @ibuildthecloud had mentioned that this is required for k3s but not for RKE2 and can be skipped here.

PR #32, opened 2020-06-16
set secrets-encrpytion flag by default and set default values

briandowns:
This PR sets the `--secrets-encryption` flag by default.


pkg/cli/cmds/server.go
```diff
@@ -10,77 +10,77 @@ var (
...
+		"log":               hide,
+		"alsologtostderr":   hide,
 		"bind-address":      nil,
-		"https-listen-port": Drop,
+		"https-listen-port": drop,
 		"advertise-address": nil,
-		"advertise-port":    Drop,
+		"advertise-port":    drop,
 		"tls-san":           nil,
 		"data-dir": {
```
ibuildthecloud, line 22: I wonder if we can set this to `nil`.  This commit https://github.com/rancher/k3s/commit/7e59c0801e905f9f574264326ba42143c665b3b5 was supposed to make all instances of the string "k3s" be changed to "rke2" at compile time. So either setting this flag is not needed or the referenced commit does not do all it should.  So ideally this should be `nil`.

pkg/cli/cmds/server.go
```diff
@@ -75,8 +75,7 @@ var (
 		"agent-token-file":           nil,
 		"server":                     nil,
 		"secrets-encryption": {
```
ibuildthecloud, line 77: Should this just be nil? It's probably not obvious from the code what is going on here.  `nil` just means "take the flag as is."  The intention here is that rke2 hides/removes/changes flags of k3s and we wanted it such that if k3s changed a flag rke2 would panic if it didn't have an explicit action for the flag.  The fear is that k3s adds a flag that is not appropriate for rke2 and we didn't notice and slips through the cracks.  So `nil` is forcing us to explicitly confirm the flag, but it takes not action to modify it.  To make the code clearer possibly a `const copy = nil` could be created and we replace `nil` with `copy`.

ibuildthecloud:
I appreciate the cosmetic code hygiene changes and that they are in separate commits.  Really helps reviewing when they can be separated from functional changes.  Thanks.

pkg/cli/cmds/k3sopts.go
```diff
@@ -29,6 +30,7 @@ func mustCmdFromK3S(cmd cli.Command, flagOpts map[string]*K3SFlagOption) cli.Com
 	if err != nil {
 		panic(err)
 	}
+	fmt.Printf("%#v\n", cmd)
```
galal-hussein, line 33: can this be removed

pkg/cli/cmds/agent.go
```diff
@@ -8,36 +8,33 @@ import (
...
-		"kube-proxy-arg":             nil,
-		"rootless":                   Drop,
-		"server":                     nil,
-		"no-flannel":                 Drop,
-		"cluster-secret":             Drop,
+		"v":                          hide,
+		"vmodule":                    hide,
+		"log":                        hide,
+		"alsologtostderr":            hide,
+		"data-dir":                   copy,
```
galal-hussein, line 15: shouldnt this be overriden to /var/lib/rancher/rke2 instead of /var/lib/rancher/k3s

briandowns: I think this qualifies per @ibuildthecloud 's comment from above about this commit: https://github.com/rancher/k3s/commit/7e59c0801e905f9f574264326ba42143c665b3b5

pkg/cli/cmds/k3sopts.go
```diff
@@ -9,13 +9,14 @@ import (
...
-	Drop = &K3SFlagOption{
+	drop = &K3SFlagOption{
 		Drop: true,
 	}
-	Hide = &K3SFlagOption{
+	hide = &K3SFlagOption{
 		Hide: true,
 	}
 )
+var copy *K3SFlagOption = nil
```
dweomer, line 19: @briandowns I appreciate what you've done with this change although some call-out as to why you are moving to package scope in a PR comment might be helpful. Also, this change, overall, makes the PR much noisier and harder to asses re: "secrets-encryption flag".

briandowns: Understood. I'll spit this out in the future. As for the "why", it's not used outside of the package.

PR #47, opened 2020-06-26
Add installation script

briandowns:
Add an installation script based off of the k3s install script and account for installing an rke2 and etcd user, updating permissions, etc. Side note, bash-isms have been removed for the closest I can get the script to posix compatible. I tested in dash and busybox so bash no longer is a requirement.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -390,15 +390,13 @@ download_hash() {
     info "downloading hash ${HASH_URL}"
     download "${TMP_HASH}" "${HASH_URL}"
     HASH_EXPECTED=$(awk -F ' ' '{print $1}' ${TMP_HASH})
-    HASH_EXPECTED=${HASH_EXPECTED%%[[:blank:]]*}
 }
 
 # installed_hash_matches checks hash against 
 # installed version.
 installed_hash_matches() {
     if [ -x ${BIN_DIR}/rke2 ]; then
-        HASH_INSTALLED=$(sha256sum ${BIN_DIR}/rke2)
-        HASH_INSTALLED=${HASH_INSTALLED%%[[:blank:]]*}
+        HASH_INSTALLED=$(sha256sum ${BIN_DIR}/rke2 | awk -F ' ' '{print $1}')
         if [ "${HASH_EXPECTED}" = "${HASH_INSTALLED}" ]; then
             return
         fi
@@ -413,7 +411,7 @@ download_binary() {
     else
         BIN_URL=${GITHUB_URL}/download/${VERSION_RKE2}/rke2-${VERSION_RKE2}.linux-${ARCH}
     fi
-    info "downloading binary ${BIN_URL}"
+    info "downloading binary at ${BIN_URL}"
     download "${TMP_BIN}" "${BIN_URL}"
 }
 
--- a/install.sh
+++ b/install.sh
@@ -87,6 +87,7 @@ INSTALL_PATH="/usr/local/bin"
 GITHUB_URL=https://github.com/rancher/rke2/releases
 STORAGE_URL=https://storage.googleapis.com/rke2-ci-builds
 DOWNLOADER=
+
 USING_RKE2_USER=0
 USING_ETCD_USER=0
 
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,877 @@
+#!/bin/sh
+
+set -e
+
+if [ "${DEBUG}" = 1 ]; then 
+    set -x
+fi
+
+# Usage:
+#   curl ... | ENV_VAR=... sh -
```
galal-hussein, line 10: Wondering if we should remove any curl mention, in this script since its not uploaded yet

install.sh
```diff
@@ -0,0 +1,877 @@
...
+#       or
+#   ENV_VAR=... ./install.sh
+#
+# Example:
+#   Installing a server without traefik:
+#     curl ... | INSTALL_RKE2_EXEC="--disable=traefik" sh -
+#   Installing an agent to point at a server:
+#     curl ... | RKE2_TOKEN=xxx RKE2_URL=https://server-url:6443 sh -
```
galal-hussein, line 18: so in rke2 the supervisor port is set to 9345 by default, so this command will fail if its using 6443

install.sh
```diff
@@ -0,0 +1,877 @@
...
+#
+#   - INSTALL_RKE2_ETCD_USER
+#     Create a user 'etcd'. If this value is set, the installation
+#     will chown the etcd data-dir to this user and update the etcd
+#     pod manifest.
+
+BASE_DIR="/var/lib/rancher/rke2"
+INSTALL_PATH="/usr/local/bin"
+GITHUB_URL=https://github.com/rancher/rke2/releases
+STORAGE_URL=https://storage.googleapis.com/rke2-ci-builds
```
galal-hussein, line 88: There is no ci builds storage sat up for rke2 yet so this wont work properly

install.sh
briandowns: I'll remove it and add it back when that feature is possible.

install.sh
briandowns: I left this in, in the hopes we have that soon. I can remove.

install.sh
```diff
@@ -0,0 +1,877 @@
...
+        arm64)
+            ARCH=arm64
+            SUFFIX=-${ARCH}
+            ;;
+        aarch64)
+            ARCH=arm64
+            SUFFIX=-${ARCH}
+            ;;
+        arm*)
+            ARCH=arm
```
galal-hussein, line 289: there is no support for arm yet for rke2

install.sh
```diff
@@ -0,0 +1,877 @@
...
+        rm -rf "${TMP_DIR}"
+        exit $code
+    }
+    trap cleanup INT EXIT
+}
+
+# --- use desired rke2 version if defined or find version from channel ---
+get_release_version() {
+    if [ -n "${INSTALL_RKE2_COMMIT}" ]; then
+        VERSION_RKE2="commit ${INSTALL_RKE2_COMMIT}"
```
galal-hussein, line 332: I guess we should either set up the same commit storage space for rke2 like k3s or skip this option

install.sh
briandowns: Was leaving that as well as that's on the way.

install.sh
briandowns: I agree. :D

install.sh
```diff
@@ -0,0 +1,871 @@
...
+
+# re-evaluate args to include env command
+eval set -- $(escape "${INSTALL_RKE2_EXEC}") $(quote "$@")
+
+# main
+{
+    if [ ! -z "${INSTALL_RKE2_USER}" ]; then
+        mkdir -p "${BASE_DIR}"
+        create_user "${INSTALL_RKE2_USER}" "RKE2 Service User Account"
+	chown -R "${INSTALL_RKE2_USER}":"${INSTALL_RKE2_USER}" "$(dirname ${BASE_DIR})"
```
galal-hussein, line 849: nit: indentation

briandowns: Ugh. Good catch!

PR #56, opened 2020-07-02
Have ETCD Run as the Etcd User

briandowns:
This pull request sets up the ability for etcd in RKE2 to run as the etcd user by updating the necessary directories to being owned by the etcd user and adding a security context to the pod specification for etcd.

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -3,6 +3,7 @@ package podexecutor
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -161,6 +164,17 @@ func (s *StaticPod) CurrentETCDOptions() (opts executor.InitialOptions, err erro
 	return
 }
 
+const (
+	basePath    = "/var/lib/rancher/rke2"
```
galal-hussein, line 168: I dont think this should be hard coded to /var/lib/rancher/rke2, the datadir for rke2 can be changed with an option

pkg/podexecutor/staticpod.go
```diff
@@ -197,5 +211,43 @@ func (s *StaticPod) ETCD(args executor.ETCDConfig) error {
 		HealthPort:  2381,
 		HealthPath:  "/health",
 		HealthProto: "HTTP",
+	}
+
+	if s.CISMode {
+		etcdUser, err := user.Lookup("etcd")
```
galal-hussein, line 217: just a question here, should we consider creating the etcd user if it doesnt exist, or at least in the install script?

galal-hussein: oh I see now that this mode has been added to the install script, nvm

pkg/podexecutor/staticpod.go
briandowns: Good point. I'll look to see how this can be avoided.

briandowns: Resolved. Thanks!

PR #58, opened 2020-07-03
Update extend install script

briandowns:
Update includes setting kernel parameters as well as adjusting how users are created and accounted for in CIS mode and updates the service files for CIS mode.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -875,10 +875,10 @@ setup_etcd_user() {
 # update_kernel_params adjusts the necessary kernel parameters
 # to allow RKE2 to run in CIS mode.
 update_kernel_params() {
-    sysctl -w vm.panic_on_oom=0
-    sysctl -w kernel.panic=10
-    sysctl -w kernel.panic_on_oops=1
-    sysctl -w kernel.keys.root_maxbytes=25000000
+    for param in vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000; do
+        sysctl -w ${param}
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -75,6 +75,15 @@ fi
 #     Create a user 'etcd'. If this value is set, the installation
 #     will chown the etcd data-dir to this user and update the etcd
 #     pod manifest.
+#
+#   - INSTALL_RKE2_CIS_MODE
+#     Enable all options to allow RKE2 to run in CIS mode if set to true. This 
+#     will change add an "etcd" system user and will update the following kernel
```
galal-hussein, line 81: nit: typo here i think: "change add"

briandowns: Thanks! Resolved.

install.sh
```diff
@@ -841,19 +853,47 @@ create_user() {
...
+    USING_ETCD_USER=1
+}
+
+# update_kernel_params adjusts the necessary kernel parameters
+# to allow RKE2 to run in CIS mode.
+update_kernel_params() {
+    sysctl -w vm.panic_on_oom=0
+    sysctl -w kernel.panic=10
+    sysctl -w kernel.panic_on_oops=1
+    sysctl -w kernel.keys.root_maxbytes=25000000
```
dweomer, line 881: These persist until the next reboot. Should we not attempt to drop-in (and apply) an `/etc/sysctl.d` entry?

briandowns: I think that'd actually be preferred. I'll do that. Thanks for bringing that up!

Issue #64, opened 2020-07-08
RKE2 Fails on Execution Looking for ETCD User in Non CIS Mode


ShylajaDevadiga:
Version:
rke2 version v0.0.1-alpha.6

Node OS:
Ubuntu 20.04

Issue:
1. When NSTALL_RKE2_CIS_MODE is set to true installation is successful.
2. When INSTALL_RKE2_CIS_MODE is not passed to install script as below, installation fails with message  "unknown user etcd"
INSTALL_RKE2_VERSION=v0.0.1-alpha.6 ./install.sh

```
Jul 07 23:07:16 ip-172-31-15-215 systemd[1]: Failed to start Rancher Kubernetes Engine v2.
Jul 07 23:07:21 ip-172-31-15-215 systemd[1]: rke2.service: Scheduled restart job, restart counter is at 6.
Jul 07 23:07:21 ip-172-31-15-215 systemd[1]: Stopped Rancher Kubernetes Engine v2.
Jul 07 23:07:21 ip-172-31-15-215 systemd[1]: Starting Rancher Kubernetes Engine v2...
Jul 07 23:07:21 ip-172-31-15-215 rke2[2299]: time="2020-07-07T23:07:21Z" level=warning msg="not running in CIS 1.5 mode"
Jul 07 23:07:21 ip-172-31-15-215 rke2[2299]: time="2020-07-07T23:07:21Z" level=info msg="Starting rke2 v0.0.1-alpha.6 (HEAD)"
Jul 07 23:07:21 ip-172-31-15-215 rke2[2299]: time="2020-07-07T23:07:21Z" level=fatal msg="starting kubernetes: preparing server: start cluster and https: user: unknown user etcd"
```

briandowns:
PR #65 has been merged. This is ready to be tested again.

ShylajaDevadiga:
Verified in rke2 version v0.0.1-alpha.7

PR #65, opened 2020-07-08
update flag value parse logic

briandowns:
A situation arose where the code looking at the flag wasn't checking to see if there was actuall a value set and was creating false positives. This remediates issue #64 .


brandond:
LGTM

PR #69, opened 2020-07-08
Add Ability to Remove Etcd & RKE2 User

briandowns:
Adds the ability for the rke2 uninstall script to remove the rke2 and etcd user, if they exist, as well as groups.

This can be verified by running the following commands:

```sh
grep etcd /etc/passwd
grep rke2 /etc/passwd
grep etcd /etc/group
grep rke2 /etc/group
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -637,6 +637,8 @@ rm -rf "${BASE_DIR}"
 rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/rke2
 rm -f ${KILLALL_RKE2_SH}
+
+userdel etcd
 EOF
     ${SUDO} chmod 755 "${UNINSTALL_RKE2_SH}"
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -637,6 +637,14 @@ rm -rf "${BASE_DIR}"
 rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/rke2
 rm -f ${KILLALL_RKE2_SH}
+
+for u in etcd rke2; do
+    if id -u \${u} 2>/dev/null; then
+        userdel \${u}
```
dweomer, line 643: If these fail for any reason they uninstall will return non-zero. Is that on purpose?
I would submit that uninstalls should rarely, if ever, exit non-zero.

briandowns: Do we want to account for that sort of edge case? The failure scnearios would be be not being able to update the passwd or group file (we're running as root) , can't remove home dir (home dir is already remvoed by this point), and user doesn't exist which we check for before attempting the removal.

Issue #49, opened 2020-06-30
Support for other linux distros and versions


ShylajaDevadiga:
RKE2 version:
v0.0.1-alpha.4

Node OS:
Ubuntu 18.04:

Describe the issue:
rke2 installation errors on ubuntu18.04.

Logs:
```
Using binary for rke2. now seeing this error `rke2: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.28' not found (required by rke2)
```

briandowns:
A binary that's statically linked can be made by issuing: `BUILDTAGS="netgo" make build`. Do we want to add this to the `BUILDTAGS` variable as a default value in the Makefile and append on additonal things when the `make` command is run? @cjellick @ibuildthecloud @galal-hussein

cjellick:
go for it

ShylajaDevadiga:
Using rke2 v0.0.1-alpha.7
On CentOS 7.8 we have the same error.
```
# cat /etc/redhat-release 
CentOS Linux release 7.8.2003 (Core)
# ./rke2 -v
./rke2: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by ./rke2)
```

briandowns:
PR #73 submitted to resolve issue.

ShylajaDevadiga:
With rke2 version v1.18.4-alpha9-rke2 

Centos 7, rpm does not exist
```
rpm -i https://rpm.rancher.io/rke2-selinux-0.1.1-rc1.el7.noarch.rpm
curl: (22) The requested URL returned error: 404 Not Found
error: skipping https://rpm.rancher.io/rke2-selinux-0.1.1-rc1.el7.noarch.rpm - transfer failed
```

Oats87:
> With rke2 version v1.18.4-alpha9-rke2
> 
> Centos 7, rpm does not exist
> 
> ```
> rpm -i https://rpm.rancher.io/rke2-selinux-0.1.1-rc1.el7.noarch.rpm
> curl: (22) The requested URL returned error: 404 Not Found
> error: skipping https://rpm.rancher.io/rke2-selinux-0.1.1-rc1.el7.noarch.rpm - transfer failed
> ```

RPM's are unavailable at this point, but will be available soon

briandowns:
This issue is different than the original reported here. I think this should be tracked  separately and tag @Oats87 .

ShylajaDevadiga:
rke2 version: v1.18.4-alpha9-rke2
[...]

PR #73, opened 2020-07-10
Have Makefile Create Static Binaries

briandowns:
Resolves issue #49 .  This can be tested by creating a new binary `make build` and copying it to a CentOS 7 machine or an Ubuntu 18.04 machine.


Makefile
```diff
@@ -47,13 +47,16 @@ endif
 REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .dirty; fi)
 RELEASE=${PROG}-$(VERSION).${GOOS}-${GOARCH}
 
-ifdef BUILDTAGS
```
ibuildthecloud, line 50: This logic was copied from containerd's Makefile so I don't even get what is was doing in the first place really.

PR #90, opened 2020-07-15
add missing logic for cis mode and kubelet option

briandowns:
This fixes and oversight from #87 . This doesn't need to be tested directly and will be caught in the tests defined in the test ticket for the original feature.


brandond:
LGTM

PR #95, opened 2020-07-18
Flag values not being set

briandowns:
For some reason flag values were not being set and propagated through the application. The behavior was noticed when RKE2 would unpack the runtime image contents to PWD rather than the "data-dir". This PR updates the ability to set the flag values and propagate them through as well as some minor updates.

This issue was causing kubelet to fail as it was unable to find containerd and other dependencies.

### DIFF
--- a/pkg/cli/cmds/k3sopts.go
+++ b/pkg/cli/cmds/k3sopts.go
@@ -62,19 +62,13 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
 		}
 
 		if opt.Usage != "" {
-			if err := flagSetUsage(flag, opt); err != nil {
-				return nil, err
-			}
+			flagSetUsage(flag, opt)
 		}
 		if opt.Default != "" {
-			if err := flagSetDefault(flag, opt); err != nil {
-				return nil, err
-			}
+			flagSetDefault(flag, opt)
 		}
 		if opt.Hide {
-			if err := flagSetHide(flag, opt); err != nil {
-				return nil, err
-			}
+			flagSetHide(flag, opt)
 		}
 		newFlags = append(newFlags, flag)
 	}
@@ -93,7 +87,7 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
 // flagSetUsage receives a flag and a K3S flag option, parses
 // both and sets the necessary fields based on the underlying
 // flag type.
-func flagSetUsage(flag cli.Flag, opt *K3SFlagOption) error {
+func flagSetUsage(flag cli.Flag, opt *K3SFlagOption) {
 	v := reflect.ValueOf(flag).Elem()
 	if v.CanSet() {
 		switch t := flag.(type) {
@@ -103,18 +97,14 @@ func flagSetUsage(flag cli.Flag, opt *K3SFlagOption) error {
 			t.Usage = opt.Usage
 		case *cli.BoolFlag:
 			t.Usage = opt.Usage
-		case *cli.IntFlag:
-		default:
-			return fmt.Errorf("error: %T is unsupported", t)
 		}
 	}
-	return nil
 }
 
 // flagSetDefault receives a flag and a K3S flag option, parses
 // both and sets the necessary fields based on the underlying
 // flag type.
-func flagSetDefault(flag cli.Flag, opt *K3SFlagOption) error {
+func flagSetDefault(flag cli.Flag, opt *K3SFlagOption) {
 	v := reflect.ValueOf(flag).Elem()
 	if v.CanSet() {
 		switch t := flag.(type) {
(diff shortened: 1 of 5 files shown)


pkg/cli/cmds/k3sopts.go
```diff
@@ -94,3 +83,61 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
...
 	return cmd, errs.Err()
 }
+
+// flagSetUsage receives a flag and a K3S flag option, parses
+// both and sets the necessary fields based on the underlying
+// flag type.
+func flagSetUsage(flag cli.Flag, opt *K3SFlagOption) {
+	v := reflect.ValueOf(flag).Elem()
+	if v.CanSet() {
+		switch t := flag.(type) {
```
brandond, line 93: Do we need to switch on type if all types support Usage the same way?

pkg/cli/cmds/k3sopts.go
```diff
@@ -94,3 +83,61 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
...
+
+// flagSetDefault receives a flag and a K3S flag option, parses
+// both and sets the necessary fields based on the underlying
+// flag type.
+func flagSetDefault(flag cli.Flag, opt *K3SFlagOption) {
+	v := reflect.ValueOf(flag).Elem()
+	if v.CanSet() {
+		switch t := flag.(type) {
+		case *cli.StringFlag:
+			x := opt.Default
```
brandond, line 112: Do we have to create a new variable for the destination vs just passing through to the upstream destination, or is there some other reason (such as destination being nil by default)?

pkg/cli/cmds/k3sopts.go
```diff
@@ -94,3 +83,61 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
...
+			t.DefaultText = x
+			t.Destination = &x
+			t.Value = opt.Default
+		case *cli.StringSliceFlag:
+			t.DefaultText = opt.Default
+			t.Destination = &([]string{opt.Default})
+			t.Value = []string{opt.Default}
+		case *cli.BoolFlag:
+			t.DefaultText = opt.Default
+			t.Destination = &opt.Hide
```
brandond, line 122: Why do bools store into the Hide flag? I thought Hide was for hiding it from help.

pkg/cli/cmds/k3sopts.go
```diff
@@ -94,3 +83,61 @@ func commandFromK3S(cmd *cli.Command, flagOpts map[string]*K3SFlagOption) (*cli.
...
+	}
+}
+
+// flagSetHide receives a flag and a K3S flag option, parses
+// both and sets the necessary fields based on the underlying
+// flag type.
+func flagSetHide(flag cli.Flag, opt *K3SFlagOption) {
+	v := reflect.ValueOf(flag).Elem()
+	if v.CanSet() {
+		switch t := flag.(type) {
```
brandond, line 134: As with Usage - if all types support Hidden can we skip the type check?

pkg/rke2/server.go
```diff
@@ -43,9 +44,14 @@ func Agent(ctx *cli.Context, cfg Config) error {
...
 
 func setup(ctx *cli.Context, cfg Config) error {
-	dataDir := cmds.ServerConfig.DataDir
-	if dataDir == "" {
-		dataDir = cmds.AgentConfig.DataDir
+	var dataDir string
+	for _, f := range ctx.Command.Flags {
+		switch t := f.(type) {
+		case *cli.StringFlag:
+			if strings.Contains(t.Name, "data-dir") {
```
brandond, line 51: Why is this contains vs equals? Might this cause problems down the road if we add another data-dir related option in the future?

pkg/cli/cmds/k3sopts.go
briandowns: Easier and cleaner means of underlying type access for assignment purposes without direct reflection invocation.

pkg/rke2/server.go
briandowns: It contains a ",d" so ctains cuts to the chase.

pkg/cli/cmds/k3sopts.go
briandowns: See comment in usage.

pkg/cli/cmds/k3sopts.go
briandowns: I'm not really sure what you're asking. Can you clarify?

brandond: As I read this, `t.Destination = &opt.Hide` means that the value of the flag will be stored in the Hide field of the option, which is used to determine if the option is hidden from the autogenerated `help` output. That seems wrong or at least counterintuitive.

davidnuzik:
@brandond Sorry to interrupt you. The remarks you have in this PR - could you follow up? Are these significant? I'm trying to get this PR through soon, if possible and sane. Rancher Fed folks are trying to build a demo cluster for a prospect.
Perhaps if these are small issues we might be able to agree to fix / tweak later?

I really hate to rush things - just want to see this move along if at all possible.

pkg/rke2/server.go
brandond: I haven't dug through the cli library enough to know off the top of my head how the Name for a flag gets built so that feels a little odd but OK.

briandowns: Agreed.

pkg/cli/cmds/k3sopts.go
briandowns: The address of the Hide field in the option struct is being stored in the Destination field on that type. I don't mind removing if not necessary but wanted to have it covered for thoroughness.

brandond:
@davidnuzik if it works then don't let me hold it up - it just scans a little weird to me so I was trying to understand the intent.

pkg/cli/cmds/k3sopts.go
brandond: Got it. This is some deep voodoo, thanks for walking me through it.

PR #98, opened 2020-07-20
use tee with sudo

briandowns:
Make sure the redirect is done in the scope of sudo

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -707,7 +707,7 @@ ExecStart=${BIN_DIR}/rke2 \\
     ${CMD_RKE2_EXEC}
 
 EOF
-    echo "HOME=/root" > /etc/systemd/system/rke2.service.env
+    echo "HOME=/root" | ${SUDO} tee /etc/systemd/system/rke2.service.env
 }
 
 # create_openrc_service_file writes the openrc


ibuildthecloud:
Does this apply to the k3s install.sh also?  Are they basically the same file with just s/k3s/rke2/g?

briandowns:
> Does this apply to the k3s install.sh also? Are they basically the same file with just s/k3s/rke2/g?

Just RKE2. This doesn't exist in K3S.

ibuildthecloud:
Probably some context I missed at some point, but why are the k3s and rke2 install scripts different (with the exception of the name).  Is it because of CIS?  Sorry I haven't read or diffed the scripts.

install.sh
```diff
@@ -707,7 +707,7 @@ ExecStart=${BIN_DIR}/rke2 \\
     ${CMD_RKE2_EXEC}
 
 EOF
-    echo "HOME=/root" > /etc/systemd/system/rke2.service.env
+    echo "HOME=/root" | ${SUDO} tee /etc/systemd/system/rke2.service.env
```
dweomer, line 710: Do we use sudo anywhere else in this script? I think we make it incumbent upon the user/operator to have the correct permissions to execute this script.

dweomer: Ah, searched, we have a SUDO macro of sorts.

briandowns: We use it a quite a few spots throughout the script here and in K3S. I'm with ya in regards to your point though.

ibuildthecloud: In k3s we allow a non-root user to run install.sh.

dweomer: LGTM :+1:

ibuildthecloud:
I'm not sure what exactly is being done for CIS. The goal with rke2 is to maintain the same operational experience/UX as k3s.  So as much as possible keep things the same between the two. So CIS is a rke2 only feature so there needs to be a deviation, so it's probably fine, I just want to restate that goal so we make sure the deviations are still within some understood UX that applies to k3s.

briandowns:
Sounds good. Thanks for the reiteration. As of now, the only operational differences are for CIS.

install.sh
brandond: Just from an audit log perspective, piping stuff into a sudo'd command seems a little sketchy since all you'll see is the tee. Maybe capture the full command by doing:
```bash
${SUDO} /bin/sh -c 'echo HOME=/root > /etc/systemd/system/rke2.service.env'
```

briandowns: I'd be happy to do this in another PR and update the rest of the references at the same time and keep the testing surface area consistent.

davidnuzik: My two cents - separate PR for that later.

PR #104, opened 2020-07-21
Add dispatch to drone to rke2-upgrade

galal-hussein:


.drone.yml
```diff
@@ -118,3 +118,42 @@ steps:
 
 depends_on:
 - amd64
+
```
brandond, line 121: Didn't we just have to take Dapper out of the dispatch step on one of the k3s release branches to get it working right?
https://github.com/rancher/k3s/commit/eb5cdfe7b8f519de48d26865c6dcc4b3d49ebea0

Issue #66, opened 2020-07-08
Need for node-external-ip flag for rke2


ShylajaDevadiga:
As of rke2 version v0.0.1-alpha.6 we do not have node-external-ip flag with rke2. 
Ingress creation fails while importing rke2 into rancher.

davidnuzik:
Cloud controller manager is actually what is needed for RKE2. Once we have this - we will be able to create an ingress with the provider.

PR #113, opened 2020-07-22
Don't drop the node-external-ip flag

brandond:
For rancher/rke2#66

Signed-off-by: Brad Davidson <brad.davidson@rancher.com>

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -30,7 +30,7 @@ var (
 		"pause-image":                drop,
 		"private-registry":           copy,
 		"node-ip":                    copy,
-		"node-external-ip":           drop,
+		"node-external-ip":           copy,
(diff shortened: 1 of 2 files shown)


galal-hussein:
I dont understand why this PR is needed since the node external IP is only works with k3s cloud provider which is not really enabled in rke2 ? @brandond

davidnuzik:
@galal-hussein this originates from https://github.com/rancher/rke2/issues/66 seems like Rancher ingress won't work? Maybe we (QA and myself) are doing something wrong and just need guidance? For example on our demo cluster we can't get wordpress to work. The ingress the wordpress chart creates uses a LAN ip.

brandond:
@davidnuzik I'll leave this here for now - if we end up wanting to also enable the stub CCM in order to resolve #66 let me know and I will update the PR.

PR #121, opened 2020-07-23
config file docs

joshrwolf:
initial wag at basic documentation around config file support with a basic example/sample

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ node-label:
   - "something=amazing"
 ```
 
-In general, cli arguments map to their respective yaml key, with set cli args being represented as yaml lists.
+In general, cli arguments map to their respective yaml key, with repeatable cli args being represented as yaml lists.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,3 +62,37 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
...
+# /etc/rancher/rke2/flags.conf
+write-kubeconfig-mode: "0644"
+tls-san:
+  - "foo.local"
+node-label:
+  - "foo=bar"
+  - "something=amazing"
+```
+
+In general, cli arguments map to their respective yaml key, with set cli args being represented as yaml lists.
```
brandond, line 84: 'set cli args' is a bit ambiguous - could either be read to mean args that you have set, or the set datatype. Maybe say 'repeatable' or 'additive'?

joshrwolf: That's a good point, will change it to say "repeatable"

Issue #118, opened 2020-07-23
README docs


davidnuzik:
For 7/24 Beta release (date subject to change) at a minimum we want to lightly cover the following in README.md:

- Config file (and simple exmaple)
- RPM (and how to hook up yum repo)
- SELinux (this could slip and if so would need to come in a second beta release)

davidnuzik:
Config file: https://github.com/rancher/rke2/pull/121

davidnuzik:
RPM: https://github.com/rancher/rke2/pull/124

davidnuzik:
After discussing with Jacob -- we both agreed we don't really have anything worth noting in the README.md. We could perhaps do something like containerd did in their release notes and basically point out that SELinux support has been improved. But this is more of a release note item anyway. I think that since there is no clear steps the user needs to take for SELinux (other than it should be enforcing) there is no need for an addition to the README here for this.

If anyone disagrees - let me know in a comment and @ me :)

PR #124, opened 2020-07-23
Update README.md for rpm instructions

Oats87:
Contributes to: https://github.com/rancher/rke2/issues/118 and adds simple RPM repo usage instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,42 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
 
 We provide a simple automated way to install RKE2 on AWS via terraform scripts, this method requires terraform to be installed and access to AWS cloud, to get started please checkout the [rke2-build](https://github.com/rancher/rke2-build) repo.
 
+## CentOS/RHEL 7 RPM Installation
+
+Signed RPMs are published for RKE2 within the `rpm.rancher.io` RPM repository. In order to use the RPM repository, on a CentOS 7 or RHEL 7 system, run the following bash snippet:
+
+```bash
+cat << EOF > /etc/yum.repos.d/rancher-rke2-1-18-latest.repo
+[rancher-rke2-common-latest]
+name=Rancher RKE2 Common Latest
+baseurl=https://rpm.rancher.io/rke2/latest/common/centos/7/noarch
+enabled=1
+gpgcheck=1
+gpgkey=https://rpm.rancher.io/public.key
+
+[rancher-rke2-1-18-latest]
+name=Rancher RKE2 1.18 Latest
+baseurl=https://rpm.rancher.io/rke2/latest/1.18/centos/7/x86_64
+enabled=1
+gpgcheck=1
+gpgkey=https://rpm.rancher.io/public.key
+EOF
+```
+
+After this, you can either run:
+
+```bash
+yum -y install rke2-server
+```
+
+or 
+
+```bash
+`yum -y install rke2-agent
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,42 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
...
+gpgkey=https://rpm.rancher.io/public.key
+EOF
+```
+
+After this, you can either run:
+
+```bash
+yum -y install rke2-server
+```
+
```
ShylajaDevadiga, line 93: Run yum makecache to download yum repository data to cache

Oats87: `yum makecache` is only necessary to pull the latest changes from the repository if there is a new RPM released between the first time you run `yum install` and later, otherwise yum will automatically build the metadata cache the first time you run `yum install`

brandond: FYI, by default the repo metadata is cached for 6 hours. yum makecache (or yum clean all) should only be necessary in development or testing workflows where we're updating packages more frequently then that.

README.md
```diff
@@ -63,6 +63,42 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
 
 We provide a simple automated way to install RKE2 on AWS via terraform scripts, this method requires terraform to be installed and access to AWS cloud, to get started please checkout the [rke2-build](https://github.com/rancher/rke2-build) repo.
 
+## CentOS/RHEL 7 RPM Installation
+
+Signed RPM's are published for RKE2 within the `rpm.rancher.io` RPM repository. In order to use the RPM repository, on a CentOS 7 or RHEL 7 system, run the following bash snippet:
```
brandond, line 68: RPMs. Plural, not possessive.

README.md
```diff
@@ -63,6 +63,42 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
...
+
+```bash
+cat << EOF > /etc/yum.repos.d/rancher-rke2-1-18-latest.repo
```
brandond, line 71: We should consider offering a rke2-release package like EPEL's [epel-release](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F) file that contains both the .repo files and GPG keys. This would make repo installation a one-liner: `rpm -i http://get.k3s.io/rke2-release.rpm`

README.md
```diff
@@ -63,6 +63,42 @@ RKE2_URL=https://myserver:6443 RKE2_TOKEN=XXX ./install.sh
...
+```bash
+yum -y install rke2-server
+```
+
+or 
+
+```bash
+`yum -y install rke2-agent
+```
+
```
ShylajaDevadiga, line 99: We may want to include the location of systemd unit file to be /usr/lib/systemd/system/ as compared to install script location /etc/systemd/system

brandond: There have been requests to change that in the install script as well.

README.md
Oats87: I agree that this is something we should eventually provide, but for now, with the RPM repository being in such a state of flux, I am following steps similar to what the Kubernetes project does as evidenced here: https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-using-native-package-management

README.md
Oats87: Should this be explicitly documented? I am under the impression that most users should never be touching the unit file we place on their system, as we are the ones that maintain/control it with our RPM.

PR #130, opened 2020-07-27
reorder env setup, user create calls, add missing sudo

briandowns:
Reorders environment setup so sudo variable is instantiated prior to it's first calling. Adds missing sudo calls.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -886,7 +886,7 @@ setup_etcd_user() {
 update_kernel_params() {
     for param in vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000 vm.overcommit_memory=1; do
         ${SUDO} sysctl -w ${param}
-        echo ${param} | ${SUDO} tee -a /etc/sysctl.d/local.conf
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -886,12 +886,15 @@ setup_etcd_user() {
 update_kernel_params() {
     for param in vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000 vm.overcommit_memory=1; do
         ${SUDO} sysctl -w ${param}
-        echo ${param} | ${SUDO} tee -a /etc/sysctl.d/local.conf
+        echo ${param} | ${SUDO} tee -a /etc/sysctl.d/local.conf >/dev/null
```
brandond, line 889: Should we maybe grep for this already being set in this file? Unless I'm missing something, it seems like re-running the install script will unconditionally append these params to the file every time, causing unnecessary file growth for folks that re-run the install process.

briandowns: I'll add an issue to do in a separate PR. Thanks for pointing that out.

briandowns: https://github.com/rancher/rke2/issues/131

Issue #131, opened 2020-07-27
Update sysctl Values Only If Necessary


briandowns:
The way the install script currently works is that it'll append creating duplicate entries.

briandowns:
PR #135 resolves this issue.

To test:

Run installation in CIS mode. Check `/etc/sysctl.d/local.conf` to see if it has sysctl values. 

```
cat /etc/sysctl.d/local.conf
```

Then run the installation again in CIS mode. The `/etc/sysctl.d/local.conf` file should have the same contents. That can be verified running:

```
cat /etc/sysctl.d/local.conf
```

For further validation, delete `/etc/sysctl.d/local.conf` and run the installation again in CIS mode.

davidnuzik:
We should put this on hold and not touch it until the refactor of the install.sh script happens.

davidnuzik:
install.sh refactor is done and ready for testing from master.

rancher-max:
The install.sh refactor has changed how this is done. It is now a manual step expected to be performed by the operator. Examples of values to set: `vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000 vm.overcommit_memory=1`

Closing this issue as no longer relevant.

PR #135, opened 2020-07-28
add a check if a kernel param already exists locally

briandowns:
This PR resolves issue #131 and adds a check to see if a given kernel parameter used in CIS hardening already exists locally. If it does, then no action is taken. If it doesn't, it's written to the given file.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -886,6 +886,8 @@ setup_etcd_user() {
 update_kernel_params() {
     for param in vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000 vm.overcommit_memory=1; do
         ${SUDO} sysctl -w ${param}
+        # check if we've already written these values and only write them
+        # if we haven't previously.
         if ! grep -Fxq ${param} /etc/sysctl.d/local.conf >/dev/null; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -886,7 +886,11 @@ setup_etcd_user() {
 update_kernel_params() {
     for param in vm.panic_on_oom=0 kernel.panic=10 kernel.panic_on_oops=1 kernel.keys.root_maxbytes=25000000 vm.overcommit_memory=1; do
         ${SUDO} sysctl -w ${param}
-        echo ${param} | ${SUDO} tee -a /etc/sysctl.d/local.conf >/dev/null
+        # check if we've already written these values and only write them
+        # if we haven't previously.
+        if ! grep -Fxq ${param} /etc/sysctl.d/local.conf >/dev/null; then
```
brandond, line 891: Could this be `grep -Fxqv` instead of `! grep -Fxq`?

briandowns:
> As an incremental change this is fine (fix unbounded growth in `/etc/sysctl.d/local.conf`). We may want to consider search for only the parameter key and not key=value

I can start some experimenting for a future PR. Thanks for bringing that to light.

install.sh
briandowns: Fails due to the file not existing at the outset.

PR #141, opened 2020-07-30
Add cloudprovider support to rke2

galal-hussein:
https://github.com/rancher/rke2/issues/142

### DIFF
--- a/pkg/cli/cmds/root.go
+++ b/pkg/cli/cmds/root.go
@@ -25,6 +25,18 @@ var (
 			EnvVars:     []string{"RKE2_REPO"},
 			Destination: &config.Repo,
 		},
+		&cli.StringFlag{
+			Name:        "cloud-provider-name",
+			Usage:       "(cloud provider) Cloud provider name",
+			EnvVars:     []string{"RKE2_CLOUD_PROVIDER_NAME"},
+			Destination: &config.CloudProviderName,
+		},
+		&cli.StringFlag{
+			Name:        "cloud-provider-config",
+			Usage:       "(cloud provider) Cloud provider configuration file path",
+			EnvVars:     []string{"RKE2_CLOUD_PROVIDER_CONFIG"},
+			Destination: &config.CloudProviderConfig,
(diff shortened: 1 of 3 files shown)


pkg/cli/cmds/root.go
```diff
@@ -25,6 +25,18 @@ var (
 			EnvVars:     []string{"RKE2_REPO"},
 			Destination: &config.Repo,
 		},
+		&cli.StringFlag{
```
briandowns, line 28: Are these dependent on one another? Should we have a check to see if both are present and not just one?

galal-hussein: well it depends, so cloud config doesnt work without cloud provider name, but the opposite is not true, so you can pass only cloud provider name  for some providers without the need to set up the cloud config file

galal-hussein: I will add a check if the config is supported without the name

pkg/rke2/server.go
```diff
@@ -74,11 +77,23 @@ func setup(ctx *cli.Context, cfg Config) error {
 
 	managed.RegisterDriver(&etcd.ETCD{})
 
+	var cpConifg *podexecutor.CloudProviderConfig
```
superseb, line 80: typo

galal-hussein: Thanks for spotting that, will fix

pkg/rke2/server.go
```diff
@@ -74,11 +77,23 @@ func setup(ctx *cli.Context, cfg Config) error {
 
 	managed.RegisterDriver(&etcd.ETCD{})
 
+	var cpConfig *podexecutor.CloudProviderConfig
+	if cfg.CloudProviderConfig != "" && cfg.CloudProviderName == "" {
+		return fmt.Errorf("cloud provider name has to be provided with cloud config path")
```
brandond, line 82: The message here could be confusing, maybe: "--cloud-provider-config requires --cloud-provider-name"

Issue #140, opened 2020-07-28
Add airgap image packaging in the ci


galal-hussein:
Right now all images in are being downloaded at runtime by containerd, that includes:

- ranchertest/kubernetes
- ranchertest/etcd
- rancher/rke2-runtime

And also the helm charts images, which includes: kube-proxy, canal, coredns, and metrics server.

this mode breaks airgap however and needs to be fixed at the static pod executor to opt out the .txt file in case of preloading images from a tarball

davidnuzik:
Sounds like this is the plan but just to clarify basically do what k3s is doing today. images.txt and .tar.gz of images.

ShylajaDevadiga:
Validated airgap tarball containt needed images. Airgap was validated in #248 
No mages are pulled from ranchertest repo
```
docker.io/rancher/calico                                         v3.13.3                cf60846041a5a       150MB
docker.io/rancher/coredns                                        v1.6.9                 ec6ed4749e32c       48.3MB
docker.io/rancher/etcd                                           v3.4.13                51401ddb110e3       60.3MB
docker.io/rancher/flannel                                        v0.13.0-rancher1-rc1   49d10619cf554       104MB
docker.io/rancher/k8s-metrics-server                             v0.3.6                 8c74b060adfc3       43.7MB
docker.io/rancher/klipper-helm                                   v0.3.0                 5c7bd28900147       50MB
docker.io/rancher/kube-proxy                                     v1.18.4                207f66a950ed5       92.4MB
docker.io/rancher/kubernetes                                     v1.18.4-beta16-rke2    b01629478cc29       203MB
k8s.gcr.io/defaultbackend-amd64                                  1.5                    b5af743e59849       2MB
k8s.gcr.io/pause                                                 3.2                    80d28bedfe5de       300kB
quay.io/kubernetes-ingress-controller/nginx-ingress-controller   0.30.0                 89ccad40ce8e4       114MB
```

PR #149, opened 2020-08-04
Add airgap images to rke2 and preload support

galal-hussein:
Add airgap support by enabling image preloading and package scripts in the make file

https://github.com/rancher/rke2/issues/140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ remote-debug: build-debug bin/dlv        ## Run with remote debugging listening 
 remote-debug-exit: bin/dlv               ## Kill dlv started with make remote-debug
 	echo exit | ./bin/dlv connect :2345 --init /dev/stdin
 
-.dev-shell-build:
+.dev-shell-build: package-airgap
 	docker build -t ${PROG}-dev --target shell .
 
 clean-cache:                             ## Clean up docker base caches used for development
@@ -140,8 +140,7 @@ dev-peer: .dev-shell-build              ## Launch a server peer to run test buil
 dev-peer-enter:                         ## Enter the peer shell on another terminal
 	docker exec -it ${PROG}-peer${PEER} bash
 
-artifacts: build download-charts
-	mkdir -p dist/artifacts
+artifacts: build download-charts package-airgap
 	cp bin/${PROG} dist/artifacts/${RELEASE}
 
 .ci: validate-ci artifacts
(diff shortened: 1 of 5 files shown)


scripts/package-airgap
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e -x
+
+cd $(dirname $0)/..
+
+airgap_image_file='scripts/airgap/image-list.txt'
+images=$(cat "${airgap_image_file}")
+xargs -n1 docker pull <<< "${images}"
+docker save ${images} -o dist/artifacts/rke2-airgap-images-amd64.tar
+cp "${airgap_image_file}" dist/artifacts/rke2-images.txt
```
briandowns, line 10:
```suggestion
cp "${airgap_image_file}" dist/artifacts/rke2-images.txt

```

pkg/images/images.go
```diff
@@ -65,3 +73,28 @@ func Pull(dir, name, image string) error {
 
 	return nil
 }
+
+func checkPreloadedImages(dir string) (bool, error) {
+	_, err := os.Stat(dir)
```
briandowns, line 78: We generally want to check if the error is nil first and then act upon it.

pkg/images/images.go
```diff
@@ -65,3 +73,28 @@ func Pull(dir, name, image string) error {
...
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		logrus.Errorf("Unable to find images in %s: %v", dir, err)
```
briandowns, line 82: Error strings shouldn't be capitalize but should follow repo convention. https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#error-strings

pkg/images/images.go
```diff
@@ -1,9 +1,11 @@
 package images
 
 import (
+	"github.com/sirupsen/logrus"
```
briandowns, line 4: Can you move this down to where the other 3rd party imports are? This is usually an ide function of goimports, goreturns, etc.

Makefile
```diff
@@ -195,5 +194,9 @@ dispatch:
             -H "Content-Type: application/json" https://api.github.com/repos/rancher/rke2-upgrade/dispatches \
             --data '{"event_type": "create_tag", "client_payload": {"tag":"'"${DRONE_TAG}"'"}}'
 
+package-airgap:
+	mkdir -p dist/artifacts
```
briandowns, line 198: If this directory is used by any other target, it can be made into it's own target.

```
dist/artifacts:
    mkdir -p $@
```

scripts/airgap/generate-list.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e -x
+
+cd $(dirname $0)
+
+export CONTAINER_RUNTIME_ENDPOINT=/run/k3s/containerd/containerd.scok
```
brandond, line 6: Typo: should be `.sock` not `.scok`?

galal-hussein: fixing

Issue #159, opened 2020-08-06
Pod Security Policy


davidnuzik:
- RKE2 should have PSPs off by default. Policies should be set up for all our components (not global e.g. not touching Users, etc).

- If the user sets --profile=cis-1.5 we will use a different default in order to pass CIS tests / benchmark. (Instead of a psp with NO restrictions, deploy one that is restrictive and will pass CIS). If CIS mode enabled, turn on admission controller.

- The default PSP can be changed post-install and no further action will be taken by RKE2. Such as with a flag could disable and specify their own PSP.

The default should be replaceable and removable. Also, if a user would like to have the default re-deployed that should be doable as well. Reference / compare to: https://docs.aws.amazon.com/eks/latest/userguide/pod-security-policy.html

briandowns:
This work has been performed in PR #199. 

Logic for loading the PSP's:

CIS:
- If the globalRestricted annotation does not exist, create PSP, role, binding.
- If the systemUnrestricted annotation does not exist, create the PSP, role, binding.
- If the globalUnrestricted annotation does not exist, check if PSP exists, and if so,
   delete it, the role, and the bindings.

non-CIS:
- If the globalUnrestricted annotation does not exist, then create PSP, role, binding.
- If the systemUnrestricted annotation does not exist, then create PSP, role, binding.
- If the globalRestricted annotation does not exist, then check if the PSP exists and
   if it doesn't, create it. Check if the associated role and bindings exist and
   if they do, delete them.

Testing this can be done by running `kubectl get psps`, `kubectl get roles`, `kubectl get clusterroles`, and `kubectl get clusterrolebindings` in conjunction with grep to identify what should and should exist depending on the operating mode.

rancher-max:
#### Validated on beta11

- With non-CIS mode, everything looks correct to me. There are no `roles` resource types, but there are psps, clusterroles, and clusterrolebindings. Additionally, the clusterroles and clusterrolebindings are correctly pointing to the unrestricted psp. I believe this is the expected behavior. See below:
```
$ k get psp,roles,rolebindings,clusterroles,clusterrolebindings -o wide | grep psp

podsecuritypolicy.policy/global-restricted-psp     false          RunAsAny   MustRunAsNonRoot   MustRunAs   MustRunAs   false            configMap,emptyDir,projected,secret,downwardAPI,persistentVolumeClaim
[...]

PR #160, opened 2020-08-06
add psp with cis flag

briandowns:
Adds PodSecurityPolicy argument to enable-admission-plugins kube-apiserver flag if the `--profile=cis-1.5` flag is issued. We're laying down the default PodSecurityPolicy in anticipation of CIS mode being enabled. This is necessary to deploy pods.

Test Plan:

This can be tested by running `ps aux | grep kube-apiserver | grep PodSecurityPolicy` and verifying the flag is present in the output when running in CIS mode.


pkg/podexecutor/staticpod.go
```diff
@@ -75,14 +75,14 @@ func (s *StaticPod) KubeProxy(args []string) error {
 }
 
 func (s *StaticPod) APIServer(ctx context.Context, etcdReady <-chan struct{}, args []string) (authenticator.Request, http.Handler, error) {
+	if err := images.Pull(s.PullImages, "kube-apiserver", s.Images.KubeAPIServer); err != nil {
```
briandowns, line 78: I moved this above to group the arg setup lines all together.

cjellick:
other devs, please review but i dont want this merged until darren or I also sign off on it.

manifests/system-podsecuritypolicy.yml
```diff
@@ -0,0 +1,91 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: default-psp-role
+subjects:
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: system:serviceaccounts
+
+---
+# Gives Helm the ability to create pods in
```
brandond, line 61: Same comment as above

manifests/system-podsecuritypolicy.yml
```diff
@@ -0,0 +1,91 @@
...
+  name: default-psp-role
+rules:
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - default-psp
+
+---
+# Gives Helm the ability to create pods in
```
brandond, line 44: Shouldn't we either bind this to just Helm, or note in the comment that it's authorizing the entire default system:serviceaccounts group?

manifests/system-podsecuritypolicy.yml
```diff
@@ -0,0 +1,91 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: default-psp
```
brandond, line 4: is someone going to complain about the 'default-psp' being so permissive? Would it be worth naming it something like 'system-permissive-psp' so that the user can provide their own default without getting confused?

briandowns: This is to get it in place and provide the ability for RKE2 to even deploy itself. The restrictiveness is on the customer to decide based on their needs, controls, etc.

manifests/system-podsecuritypolicy.yml
briandowns: Open to suggestions on best approach.

manifests/system-podsecuritypolicy.yml
brandond: Yeah, mostly just thinking about it from a perception perspective. Explaining to an auditor why the default is so privileged, vs something with a name that makes it clear that it's used by core system components.

briandowns: This is going to be updated to comply with CIS reqs in another PR soon. However, even if it were left open, the onus of responsibility is on the customer per their compliance requirements which will vary by a number of factors.

manifests/system-podsecuritypolicy.yml
brandond: Since the Helm Controller jobs run using a ServiceAccount per Job, and we are restricted to using the built-in system:serviceaccounts groups (can't add ServiceAccounts to custom groups), this is the best we can do without modifying the HelmController itself to manage PSPs

```suggestion
# Gives all ServiceAccounts the ability to create pods in
```

manifests/system-podsecuritypolicy.yml
brandond:
```suggestion
# Gives all ServiceAccounts the ability to create pods in
```

davidnuzik:
Linking to gh issue: https://github.com/rancher/rke2/issues/159

cjellick:
Takeaways from our design discussion:

Goals:
1. We want the PSP admission controller on-by-default, so that to get a more secure cluster, the admin just needs manipulate PSPs and bindings, rather than redeploy their cluster.
2. The cluster admin must be able to fully control the PSPs that are deployed in the cluster. This means we cannot automatically reconcile certain special PSPs that we need. we can put them in place to start with, but shouldn't update them after laying them down.
  - After initial deployment, the admin should be able to edit or delete any of our initial PSPs or the bindings without rke2 later modifying or recreating them (even after a restart)
3. If an admin does want to reset our initial PSPs, roles, and bindings, they should be able to do so


Key changes:
1. Enable the PSP admission controller by default. In fact, it will _always_ be on.

2. We need to create 3 PSPs during initial provisioning
- global-unrestricted policy that is wide open and equivalent to what you have in originally in this PR.
- global-restricted policy. Use the example from K8s docs as the basis: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies. Check it against the controls in the CIS doc and make sure it satisfies all the checks in section 5.2.
- system-unrestricted policy that is wide open (for use in the kube-system namespace for our system services)


3. PSPs and their bindings must happen programatically, they cannot be created using the manifests dir because that will cause them to regenerate on start-up.
To accommodate this, we need to track state of these PSPs, roles, and bindings. We agreed on adding annotations to the kube-system namespace. We will need three. We didn't agree on the exact name or format, but here is what I am thinking:
- `psp.rke2.io/global-unrestricted: resolved` (and the same for global-restricted and system-unrestricted). We shouldn't care about the value of the annotation at this point, just the presence or absence of the annotation itself.

4. The bindings we create, PSPs we keep around, and annotations we record should change based on whether you are in CIS mode or not.
- When you are **not** in CIS mode, create bindings to the global-unrestricted policy. Rather than bind to service accounts, go ahead and bind to all authenticated users (including service accounts). This is the widest net we can cast
```
subjects:
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: system:authenticated
```
In this mode, you should also create the other two PSPs as well. The global-restricted will be unused. The system-unrestricted should still be used for the kube-system namespace and system:nodes group.
- When you **are** in CIS mode, the end result is to have all users (represented by the system:authenticated group) to be bound to the restricted policy by default. This will allow them create restricted pods.
- When in this mode, at the end of startup, the global-unrestricted policy should NOT exist. One bit of guidance  that @ibuildthecloud  gave was that when in CIS mode we should delete the unrestricted policiy and delete the annotation as well. I don't think this will work because if the user later came in and created a policy called global-unrestricted, on the next restart we would want to delete it. I think we need to keep the annotation around and only ensure global-unrestricted is absent if the annotation is NOT present. This would lead to "what happens if I go from CIS-mode to non-CIS mode?" In that scenario, you'd be required to delete the annotation if you wanted rke2 to recreate the PSP.

cjellick:
Update: to facilitate swiching back and forth bt cis and non-cis mode, we need to care about the values of the annotations. This prevents the user from having to delete annotations upon switching modes. I am think we use values like `resolved-cis` and` resolved-non-cis`.

Issue #137, opened 2020-07-28
Move the makefile targets to scripts


galal-hussein:
Since the Makefile is increasingly growing, its targets should be moved to scripts in the scripts directory

davidnuzik:
PR: https://github.com/rancher/rke2/pull/163

QA should be able to make locally and this is working now. Closing.

Issue #136, opened 2020-07-28
Separate the kubernetes versions from the build version


galal-hussein:
So right now the release version is tied to the build version, which causes a lot of images to be pushed with the same content, for example:

- v1.18.4-alpha19-rke2
- v1.18.4-alpha20-rke2

so two are basically built with the same content which is v1.18.4, so the k8s version should be an env variable to be passed to the Makefile and be checked against the drone tag for the version

davidnuzik:
PRs: https://github.com/rancher/rke2/pull/209 and https://github.com/rancher/rke2/pull/100

ShylajaDevadiga:
As discussed with @dweomer verified kubectl server and client versions match.
```
rke2 -v
rke2 version v1.18.4-beta11+rke2 (b459d18894e4f2a625e309f686cd95aa1cccf304)

kubectl version
Client Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.4-beta11+rke2", GitCommit:"c96aede7b5205121079932896c4ad89bb93260af", GitTreeState:"clean", BuildDate:"2020-08-27T05:42:41Z", GoVersion:"go1.14.2b4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.4-beta11+rke2", GitCommit:"c96aede7b5205121079932896c4ad89bb93260af", GitTreeState:"clean", BuildDate:"2020-08-27T05:42:41Z", GoVersion:"go1.14.2b4", Compiler:"gc", Platform:"linux/amd64"}
```
Version in nodes listing reflect the rke2 version correctly.
```
kubectl get nodes -o wide
NAME              STATUS   ROLES         AGE   VERSION               INTERNAL-IP    EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION   CONTAINER-RUNTIME
ip-172-31-34-20   Ready    etcd,master   45m   v1.18.4-beta11+rke2   172.31.34.20   <none>        Ubuntu 20.04.1 LTS   5.4.0-1021-aws   containerd://1.3.6-k3s2
ip-172-31-39-83   Ready    etcd,master   64m   v1.18.4-beta11+rke2   172.31.39.83   <none>        Ubuntu 20.04.1 LTS   5.4.0-1021-aws   containerd://1.3.6-k3s2
[...]

Issue #114, opened 2020-07-22
rke2-runtime image can't be preloaded from image directory


galal-hussein:
rke2 bootstrap code always tries to connect to a remote registry to pull the rke2-runtime image, which causes dev builds for rke2-runtime to be useless unless its pushed to a remote registry.

The code should be able to preload the image from the images/ directory

davidnuzik:
This also has an impact on airgap (from @cjellick )

brandond:
This is related to #233

nikkelma:
Also related: #248

nikkelma:
Can this be closed, or is there some other requirement different from #248 that's being tracked here?

cjellick:
yep, this was effectively duped

PR #163, opened 2020-08-07
restructuring Makefile

galal-hussein:
- use scripts with makefile
- fix loading the bootstrap image from a tarball
- fix dev-shell and dev-peer to work directily with preloaded images
- fix validating the code

Signed-off-by: galal-hussein <hussein.galal.ahmed.11@gmail.com>

https://github.com/rancher/rke2/issues/137
https://github.com/rancher/rke2/issues/136
https://github.com/rancher/rke2/issues/114


scripts/version.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Setting kubernetes version for rke2
+if [ -z "${KUBERNETES_VERSION}" ]; then
+  if [ -n "${GIT_TAG}" ]; then
+    if [[ ! "$GIT_TAG" =~ ^"${KUBERNETES_VERSION}"[+-] ]]; then
+          echo "Tagged version '$GIT_TAG' does not match expected version '${KUBERNETES_VERSION}[+-]*'" >&2
+          exit 1
+    fi
+  fi
+fi
```
dweomer, line 66: missing newline

scripts/validate
```diff
@@ -0,0 +1,31 @@
...
+GO111MODULE=on go mod tidy
+GO111MODULE=on go mod vendor
+
+. ./scripts/version.sh
+
+if [ -n "$DIRTY" ]; then
+    echo Source dir is dirty
+    git status --porcelain --untracked-files=no
+    exit 1
+fi
```
dweomer, line 31: missing newline

scripts/package-airgap
```diff
@@ -3,8 +3,9 @@ set -e -x
 
 cd $(dirname $0)/..
 
+mkdir -p dist/artifacts
 airgap_image_file='scripts/airgap/image-list.txt'
 images=$(cat "${airgap_image_file}")
 xargs -n1 docker pull <<< "${images}"
 docker save ${images} -o dist/artifacts/rke2-airgap-images-amd64.tar
-cp "${airgap_image_file}" dist/artifacts/rke2-images.txt
+cp "${airgap_image_file}" dist/artifacts/rke2-images.txt
```
dweomer, line 11: missing newline

Makefile
```diff
@@ -82,121 +15,21 @@ ci-shell: clean .dapper                  ## Launch a shell in the CI environment
...
-    done
-
 ./.dapper:
 	@echo Downloading dapper
 	@curl -sL https://releases.rancher.com/dapper/v0.5.0/dapper-$$(uname -s)-$$(uname -m) > .dapper.tmp
 	@@chmod +x .dapper.tmp
 	@./.dapper.tmp -v
 	@mv .dapper.tmp .dapper
 
-k8s-image: k8s-image-build k8s-image-scan
```
erikwilson, line 173: Might be good to leave the dependencies in place and have separate build & scan scripts (for example).

Makefile
```diff
@@ -82,121 +15,21 @@ ci-shell: clean .dapper                  ## Launch a shell in the CI environment
...
-dispatch:
-	curl -XPOST -u "${PAT_USERNAME}:${PAT_TOKEN}" \
-            -H "Accept: application/vnd.github.everest-preview+json"  \
-            -H "Content-Type: application/json" https://api.github.com/repos/rancher/rke2-upgrade/dispatches \
-            --data '{"event_type": "create_tag", "client_payload": {"tag":"'"${DRONE_TAG}"'"}}'
-
-package-airgap:
-	mkdir -p dist/artifacts
-	scripts/package-airgap
+$(TARGETS):
```
erikwilson, line 31: Might want to declare as phony

scripts/version.sh
```diff
@@ -0,0 +1,66 @@
...
+
+if [[ -n "$GIT_TAG" ]]; then
+    VERSION=$GIT_TAG
+else
+    VERSION="${KUBERNETES_VERSION}-dev+${COMMIT:0:8}$DIRTY"
+fi
+VERSION="$(sed -e 's/+/-/g' <<< "$VERSION")"
+
+# Setting kubernetes version for rke2
+if [ -z "${KUBERNETES_VERSION}" ]; then
```
briandowns, line 59: Can we use same size tab width consistent throughout the file?

scripts/version.sh
```diff
@@ -0,0 +1,66 @@
...
+        GIT_TAG=$(git tag -l --contains HEAD | head -n 1)
+    fi
+    if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+        DIRTY="-dirty"
+        TREE_STATE=dirty
+    fi
+
+    COMMIT=$(git log -n3 --pretty=format:"%H %ae" | grep -v ' drone@localhost$' | cut -f1 -d\  | head -1)
+    if [ -z "${COMMIT}" ]; then
+    COMMIT=$(git rev-parse HEAD || true)
```
briandowns, line 47:
```suggestion
        COMMIT=$(git rev-parse HEAD || true)
```

scripts/version.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ -z "$GOOS" ]; then
+  if [ "${OS}" == "Windows_NT" ]; then
+    GOOS="windows"
+  else
+    UNAME_S=$(shell uname -s)
+		if [ "${UNAME_S}" == "Linux" ]; then
+				GOOS="linux"
+		elif [ "${UNAME_S}" == "Darwin" ]; then
+				GOOS="darwin"
+		elif [ "${UNAME_S}" == "FreeBSD" ]; then
```
briandowns, line 21: Is this necessary?

Makefile
```diff
@@ -1,71 +1,4 @@
...
-GO_BUILDTAGS ?= no_embedded_executor
-GO_BUILDTAGS += ${DEBUG_TAGS}
-GO_TAGS=$(if $(GO_BUILDTAGS),-tags "$(GO_BUILDTAGS)",)
-GO_LDFLAGS=-ldflags '-extldflags "-static"                        \
-					 -X $(RKE2_PKG)/pkg/images.KubernetesVersion=$(KUBERNETES_VERSION)    \
-					 -X $(K3S_PKG)/pkg/version.Program=$(PROG)    \
-					 -X $(K3S_PKG)/pkg/version.Version=$(VERSION) \
-					 -X $(K3S_PKG)/pkg/version.Revision=$(REVISION) $(EXTRA_LDFLAGS)'
-
+TARGETS := $(shell ls scripts | grep -v \\.sh)
```
ibuildthecloud, line 1: I prefer have explicit targets with dependencies and the help comment to `make help` will work.

pkg/bootstrap/stage.go
```diff
@@ -11,6 +11,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+
```
briandowns, line 15: Kill newline

scripts/build
```diff
@@ -0,0 +1,40 @@
...
+GO_BUILDTAGS=${GO_BUILDTAGS}${BUILDTAGS}${DEBUG_TAGS}
+
+VERSION_FLAGS="
+        -X ${RKE2_PKG}/pkg/images.KubernetesVersion=${KUBERNETES_VERSION}
+        -X ${K3S_PKG}/pkg/version.Program=${PROG}
+        -X ${K3S_PKG}/pkg/version.Version=${VERSION}
+        -X ${K3S_PKG}/pkg/version.Revision=${REVISION}
+"
+STATIC_FLAGS='
+    -extldflags "-static"
```
briandowns, line 30: Does this need to be multi line?

Makefile
ibuildthecloud: Basically, I don't want dependencies in scripts, but I do want proper dependencies.  So I should just be able to run `make package` and it does all it needs to.  `make dev-shell`.  Right now `dev-shell` only works if I run the images stuff first myself.

Makefile
ibuildthecloud: PHONY is required, for example `build` will conflict with `build/`

pkg/bootstrap/stage.go
```diff
@@ -165,3 +174,27 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
+		if err != nil {
+			return err
+		}
+		files[path] = info
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	for fileName := range files {
+		if strings.HasSuffix(fileName, "rke2-runtime-image-amd64.tar") {
```
ibuildthecloud, line 191: You shouldn't look for a special file.  Just scan all tarballs and call ImageFromPath(..., Tag) with the tag you are looking for.  We expect for airgap the runtime image will be in the same tar as everything else.  You'll get an error if you don't find the tag, just ignore the error.

scripts/version.sh
ibuildthecloud: And 4 being the correct width. :)

scripts/dev-shell-build.sh
```diff
@@ -0,0 +1,26 @@
...
+cd $(dirname $0)/..
+
+IMAGE_REPO=ranchertest
+
+. ./scripts/version.sh
+
+mkdir -p build
+# getting preloaded images ready
+# creating a special tarball to be used within rke2 dev shell and peer
+docker save ${IMAGE_REPO}/kubernetes:${VERSION} \
```
ibuildthecloud, line 14: I'd have the dev-shell-build target depend on `build/images/airgap.tar` and then a target for `build/images/airgap.tar` that builds that tar.  So the build will happen once.  Then I'd bind mount build/images to /var/lib/rancher/rke2/agent/images.  I wouldn't copy around.

scripts/validate
```diff
@@ -0,0 +1,31 @@
...
+
+echo Running: go fmt
+go fmt ./
+echo Running: go vet
+go vet ./
+
+echo Running: golangci-lint
+golangci-lint run
+
+GO111MODULE=on go mod tidy
```
ibuildthecloud, line 22: GO111MODULE=on should be the default already, I'd remove this line.

scripts/validate
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -e
+
+cd $(dirname $0)/..
+
+GO=${GO-go}
+
+if ! command -v golangci-lint; then
+    echo Skipping validation: no golangci-lint available
+    exit
```
ibuildthecloud, line 10: This should check after go fmt/vet is run.

scripts/validate
```diff
@@ -0,0 +1,31 @@
...
+echo Running: go fmt
+go fmt ./
+echo Running: go vet
+go vet ./
+
+echo Running: golangci-lint
+golangci-lint run
+
+GO111MODULE=on go mod tidy
+GO111MODULE=on go mod vendor
```
ibuildthecloud, line 23: Why vendor?  We don't want to vendor in rke2.  Do you want verify?

scripts/version.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ -d .git ]; then
+    if [ -z "$GIT_TAG" ]; then
+        GIT_TAG=$(git tag -l --contains HEAD | head -n 1)
+    fi
+    if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+        DIRTY="-dirty"
+        TREE_STATE=dirty
+    fi
+
+    COMMIT=$(git log -n3 --pretty=format:"%H %ae" | grep -v ' drone@localhost$' | cut -f1 -d\  | head -1)
```
ibuildthecloud, line 45: Why do we do this? Does drone not just set a env var with the commit in it?

Makefile
galal-hussein: got it, will fix

pkg/bootstrap/stage.go
galal-hussein: the problem is that image from path requires one image in a tarball

Makefile
```diff
@@ -82,86 +19,76 @@ ci-shell: clean .dapper                  ## Launch a shell in the CI environment
...
-    	  sed -e "s|%{CHART_CONTENT}%|$$chart_content|g" $(MANIFEST_DIR)/$$chart_name.yml > $(CHARTS_DIR)/$$chart_name-chart.yml;\
-    	  rm ${CHARTS_DIR}/$$chart_name-$$chart_version.tgz;\
-    done
+download-charts: 						## Download packaged helm charts
+	./scripts/download-charts
+
+package: download-charts image k8s-image package-airgap ## Package the rke2 binary
+	./scripts/package
+
+package-airgap:	build/images/airgap.tar		## Package docker images for airgap environment
```
briandowns, line 90:
```suggestion
package-airgap: build/images/airgap.tar		## Package docker images for airgap environment
```

pkg/bootstrap/stage.go
```diff
@@ -4,20 +4,21 @@ import (
...
-
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/sirupsen/logrus"
+	"io"
```
briandowns: I think your goimports, goreturns, (whatever you're using) might not be working. The stdlibs should be fmt'd above the 3rd party.

scripts/dev-peer
```diff
@@ -0,0 +1,21 @@
...
+    ${PROG}-dev-shell:${PROG}-server \
+    --name ${PROG}-peer${PEER} \
+    --hostname ${PROG}-peer${PEER} \
+    -p 127.0.0.1:234${PEER}:2345 \
+    -ti -e WORKSPACE=$(pwd) \
+    -v ${HOME}:${HOME} \
+    -v ${PROG} -w $(pwd) \
+    -v ${PWD}/build/images:/var/lib/rancher/rke2/agent/images \
+    --privileged -v ${PROG}-pkg:/go/pkg \
+    -v ${PROG}-cache:/root/.cache/go-build ${PROG}-dev bash
```
briandowns, line 21:
```suggestion
    -v ${PROG}-cache:/root/.cache/go-build ${PROG}-dev bash

```

Makefile
```diff
@@ -1,71 +1,8 @@
...
-GO_BUILDTAGS ?= no_embedded_executor
-GO_BUILDTAGS += ${DEBUG_TAGS}
-GO_TAGS=$(if $(GO_BUILDTAGS),-tags "$(GO_BUILDTAGS)",)
-GO_LDFLAGS=-ldflags '-extldflags "-static"                        \
-					 -X $(RKE2_PKG)/pkg/images.KubernetesVersion=$(KUBERNETES_VERSION)    \
-					 -X $(K3S_PKG)/pkg/version.Program=$(PROG)    \
-					 -X $(K3S_PKG)/pkg/version.Version=$(VERSION) \
-					 -X $(K3S_PKG)/pkg/version.Revision=$(REVISION) $(EXTRA_LDFLAGS)'
-
+.PHONY: clean clean-cache ci ci-shell dapper-ci \
```
briandowns, line 1: Can you throw `.PHONY` on the non file targets?

scripts/airgap-images.sh
```diff
@@ -0,0 +1,16 @@
...
+cd $(dirname $0)/..
+
+. ./scripts/version.sh
+
+mkdir -p ./build/images/
+airgap_image_file='scripts/airgap/image-list.txt'
+images=$(cat "${airgap_image_file}")
+xargs -n1 docker pull <<< "${images}"
+if [ -z "${GIT_TAG}" ]; then
+  images=$(echo $images | sed "s/kubernetes:.* /kubernetes:${VERSION} /g")
```
briandowns, line 13:
```suggestion
    images=$(echo $images | sed "s/kubernetes:.* /kubernetes:${VERSION} /g")
```

scripts/build
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -e -x
+
+cd $(dirname $0)/..
+
+. ./scripts/version.sh
+
+if [ -z "${GODEBUG}" ]; then
+    EXTRA_LDFLAGS="${EXTRA_LDFLAGS} -s -w"
+	DEBUG_GO_GCFLAGS=""
```
briandowns, line 10:
```suggestion
    DEBUG_GO_GCFLAGS=""
```

scripts/build
```diff
@@ -0,0 +1,38 @@
...
+	DEBUG_TAGS=""
```
briandowns, line 11:
```suggestion
    DEBUG_TAGS=""
```

pkg/bootstrap/stage.go
```diff
@@ -165,3 +174,30 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	}
 	return os.Rename(tempDir, dir)
 }
+
+func preloadBootstrapImage(dataDir, runtimeImage string) (v1.Image, error) {
+	imagesDir := filepath.Join(dataDir, "agent", "images")
+	files := map[string]os.FileInfo{}
+	if err := filepath.Walk(imagesDir, func(path string, info os.FileInfo, err error) error {
```
brandond, line 181: This is a little hard to scan - is it worth assigning the anonymous function to a local variable and then passing that into Walk to improve readability?

scripts/version.sh
galal-hussein: yeah but thats in case the GIT_TAG doesnt exist which means make is being executed locally from a dev machine

pkg/bootstrap/stage.go
galal-hussein: I think thats the common pattern for calling filepath.Walk() and I found several instances in the code with the same pattern, but I can see your point of making the code more readable

Issue #156, opened 2020-08-06
SELinux needs to be enabled in containerd config


ShylajaDevadiga:
Version
rke2 version v1.18.4-beta1-rke2
Node OS:
Centos 7.8
RHEL 7.8

Issue:
Install rke2 using yum install method 
Selinux is set to Enforcing mode.
Containerd config has SELinux set to false by default
Correct labels are not set.
```
getenforce
Enforcing

cat /var/lib/rancher/rke2/agent/etc/containerd/config.toml  |grep selinux
  enable_selinux = false

ps -eZ|grep controller
system_u:system_r:spc_t:s0       9354 ?        00:00:23 kube-controller
```

No workarounds possible
Tried to manually set selinux to true in the config before starting rke2, but the file gets overridded and reverted back to false.
Updating config to set selinux to true after starting rke2 and deleting the pods, recreates the pods but does not restart the process. 

```
ps -eZ|grep controller
system_u:system_r:spc_t:s0       9354 ?        00:00:30 kube-controller

kubectl delete pod -n kube-system kube-controller-manager-ip-172-31-33-179.us-east-2.compute.internal
pod "kube-controller-manager-ip-172-31-33-179.us-east-2.compute.internal" deleted

kubectl get pods -A |grep controller
kube-system   kube-controller-manager-ip-172-31-33-179.us-east-2.compute.internal   1/1     Running        0          22s

ps -eZ|grep controller
system_u:system_r:spc_t:s0       9354 ?        00:00:31 kube-controller
```

davidnuzik:
(Needs a release to be cut before we can pass this on to QA to test)

ShylajaDevadiga:
With beta8, Issue is partially resolved. 
1. Containerd config has SELinux enabled
2. Correct labels are not set

```
rke2 -v
rke2 version v1.18.4-beta8+rke2 (b62aa4edd0f80252673754b20ff203379cb59986)

cat /var/lib/rancher/rke2/agent/etc/containerd/config.toml |grep enable
  enable_selinux = true

ps -eZ|grep controller
system_u:system_r:spc_t:s0      10648 ?        00:00:08 kube-controller

 ps -eZ|grep nginx
system_u:system_r:spc_t:s0      13948 ?        00:00:00 nginx-ingress-c
system_u:system_r:spc_t:s0      13991 ?        00:00:00 nginx

```

ShylajaDevadiga:
Validated ussing beta17
```
rke2 -v
[...]

PR #173, opened 2020-08-11
cli: add --selinux flag to agent/server sub-cmds

dweomer:
Depends on rancher/k3s#2111
Addresses rancher/rke2#156


brandond:
LGTM, I'll hit it with review once it comes out of draft with the final module locations.

Issue #186, opened 2020-08-13
kubectl port-forward not working on rke2 clusters


MonzElmasry:
**RKE2 Version**: v1.18.4-beta1-rke2
**Kind**: enhancement
using kubectl port-forward to expose a pod to a port, needed for rke2 monitoring in rancher dashboard
```
kubectl port-forward prometheus-rancher-monitoring-prometheus-0 9090 -n monitoring-system
Forwarding from 127.0.0.1:9090 -> 9090
Forwarding from [::1]:9090 -> 9090
Handling connection for 9090
Handling connection for 9090
E0813 21:07:59.561324   41971 portforward.go:400] an error occurred forwarding 9090 -> 9090: error forwarding port 9090 to pod 347320031b4d5ef5aafb9555bd6d8de60c1a05d520a0d3e793764397153c2432, uid : failed to find socat: exec: "socat": executable file not found in $PATH
```
socat should be included in rke2-packaging as containerd/cri 1.3 uses socat for port forwarding

ShylajaDevadiga:
Validated port-forward works with master build
rke2 -v
rke2 version v1.18.4-beta6-rke2 (80d010823c34c6126ff24d1cc1226a5e8302ac4c)

```
 kubectl port-forward redis-master-6b54579d85-szzlv 7000:6379
Forwarding from 127.0.0.1:7000 -> 6379
Forwarding from [::1]:7000 -> 6379
Handling connection for 7000

 redis-cli -p 7000
127.0.0.1:7000> ping
PONG
127.0.0.1:7000> 
```

PR #189, opened 2020-08-14
runtime: snag socat from the k3s image

dweomer:
Addresses #186


dweomer:
Wanted to mention, that we can eliminate this runtime dependency when we move to containerd 1.4.x

PR #198, opened 2020-08-18
building: add BUILDING.md

dweomer:
Describe how to build and run on arbitrary systems.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,51 @@
+# Building RKE2
+
+## Prerequisites
+
+By default, RKE2 is built with Dapper which uses Docker. To build RKE2 you will need to install these packages:
+- bash
+- docker
+- gcc (CGO, don't ya know, if using `scripts/build` directly)
+- go (check the `go.mod` for which series of go, e.g. 1.14.x, 1.15.x, etc)
+- make
+
+### Required for Running
+When running RKE2 you will also need to install these packages:
+- libseccomp (libseccomp2 on Debian/Ubuntu)
+- ca-certificates
+
+## Building
+
+```shell script
+# for non air-gap testing
+make build image
+# for air-gap testing
+make build-airgap
+```
+
+## Running
+
+### rke2 (dev-shell)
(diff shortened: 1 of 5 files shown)


BUILDING.md
```diff
@@ -0,0 +1,51 @@
...
+## Prerequisites
+
+By default, RKE2 is built with Dapper which uses Docker. To build RKE2 you will need to install these packages:
+- bash
+- docker
+- gcc (CGO, don't ya know, if using `scripts/build` directly)
+- go (check the `go.mod` for which series of go, e.g. 1.14.x, 1.15.x, etc)
+- make
+
+When running RKE2 you will also need to install these packages:
```
cjellick, line 12: on my first read-through, i didnt catch that these two were runtime deps. caused me to think i needed them at build time and raise an eyebrow. might be worth either moving these runtime deps or giving them a more explicit sub-heading

BUILDING.md
```diff
@@ -0,0 +1,55 @@
...
+- copy `./bin/rke2` to the path on your host
+- copy `./build/images/rke2-runtime-image-$(go env GOARCH).tar` to `/var/lib/rancher/rke2/agent/images/` on your host
+- if testing airgap, also copy `./build/images/airgap.tar` to `/var/lib/rancher/rke2/agent/images/` on your host
+- run rke2 server: `rke2 server --token=test`
+
+### kubectl
+
+It isn't obvious but `kubectl` will be installed and ready to use after starting up `rke2`. To use it you will need to:
+- `export KUBECONFIG=/etc/rancher/rke2/rke2.yaml`
+- `export $(strings < /proc/$(ps -eaf | grep containerd | head -1 | awk '{print $2}')/environ | grep 'PATH=')`
```
brandond, line 51: I'm a big fan of: 
```bash
export PATH="$(ls -atd /var/lib/rancher/rke2/data/*/bin | head -n 1):$PATH"
```

dweomer: That will work on fresh systems but may cause mysterious weirdness when installing over the top of a previous version.

PR #199, opened 2020-08-18
CIS and non CIS psp implementation

briandowns:
Intent for everything can be gleaned from the SetPSPs doc comment which also goes through logic flow.

### DIFF
--- a/pkg/rke2/psp/psp.go
+++ b/pkg/rke2/psp/psp.go
@@ -1,0 +1,443 @@
+package psp
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/rancher/spur/cli"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/policy/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/transport"
+)
+
+const (
+	defaultRetries          = 5
+	defaultWaitSeconds      = 5
+	defaultTimeout          = 30
+	k8sWrapTransportTimeout = 30
+)
+
+const (
+	kubeSystemNamespace = "kube-system"
+	cisAnnotationValue  = "resolved"
+
+	namespaceAnnotationBase               = "psp.rke2.io/"
+	namespaceAnnotationGlobalRestricted   = namespaceAnnotationBase + "global-restricted"
+	namespaceAnnotationGlobalUnrestricted = namespaceAnnotationBase + "global-unrestricted"
+	namespaceAnnotationSystemUnrestricted = namespaceAnnotationBase + "system-unrestricted"
+)
+
+// setGlobalUnrestricted sets the global unrestricted podSecurityPolicy with
+// the associated role and rolebinding.
+func setGlobalUnrestricted(cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	if _, ok := ns.Annotations[namespaceAnnotationGlobalUnrestricted]; !ok {
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), globalUnrestrictedPSPName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(globalUnrestrictedPSPTemplate, globalUnrestrictedPSPName)
+				if err := deployPodSecurityPolicyFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), globalUnrestrictedRoleName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(roleTemplate, globalUnrestrictedRoleName, globalUnrestrictedPSPName)
+				if err := deployClusterRoleFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+
+		if _, err := cs.RbacV1().ClusterRoleBindings().Get(context.TODO(), globalUnrestrictedRoleBindingName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(globalUnrestrictedRoleBindingTemplate, globalUnrestrictedRoleBindingName, globalUnrestrictedRoleName)
+				if err := deployClusterRoleBindingFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+
+		ns.SetAnnotations(map[string]string{namespaceAnnotationGlobalUnrestricted: cisAnnotationValue})
+	}
+	return nil
+}
+
+// setSystemUnrestricted sets the system unrestricted podSecurityPolicy as
+// the associated role and rolebinding.
+func setSystemUnrestricted(cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	if _, ok := ns.Annotations[namespaceAnnotationSystemUnrestricted]; !ok {
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), systemUnrestrictedPSPName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(systemUnrestrictedPSPTemplate, systemUnrestrictedPSPName)
+				if err := deployPodSecurityPolicyFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), systemUnrestrictedRoleName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(roleTemplate, systemUnrestrictedRoleName, systemUnrestrictedPSPName)
+				if err := deployClusterRoleFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+		if _, err := cs.RbacV1().ClusterRoleBindings().Get(context.TODO(), systemUnrestrictedRoleBindingName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(systemUnrestrictedNodesRoleBindingTemplate, systemUnrestrictedRoleBindingName, systemUnrestrictedRoleName)
+				if err := deployClusterRoleBindingFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+		if _, err := cs.RbacV1().ClusterRoleBindings().Get(context.TODO(), systemUnrestrictedSvcAcctRoleBindingName, metav1.GetOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				tmpl := fmt.Sprintf(systemUnrestrictedServiceAcctRoleBindingTemplate, systemUnrestrictedSvcAcctRoleBindingName, systemUnrestrictedRoleName)
+				if err := deployRoleBindingFromYaml(cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+		ns.SetAnnotations(map[string]string{namespaceAnnotationSystemUnrestricted: cisAnnotationValue})
+	}
+	return nil
+}
+
+// kubeConfigExists checks if the kubeconfig file
+// has been written.
+func kubeConfigExists(kc string) bool {
+	info, err := os.Stat(kc)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+// SetPSPs sets the default PSP's based on the mode that RKE2 is running in. There is either CIS or non
+// CIS mode. For CIS mode, the globalRestricted and systemUnrestructed polices and associated roles and
+// bindings are created. For non CIS mode, globalRestricted, system-unrestructed, and global-unrestricted
+// policies and associated roles and bindings are loaded, if applicable.
+//
+// Each PSP has an annotation associated with it that we check to see exists before performing any
+// write operations. The annotation is created upon successful setting of PSPs.
+//
+// Load logic
+//
+// CIS:
+// - Check if the globalRestricted annotation does not exist, create PSP, role, binding.
+// - Check if the systemUnrestricted annotation does not exist, create the PSP, role, binding.
+// - Check if the globalUnrestricted annotation does not exist, check if PSP exists, and if so,
+//   delete it, the role, and the bindings.
+//
+// non-CIS:
+// - Check if the globalUnrestricted annotation does not exist, create PSP, role, binding.
+// - Check if the systemUnrestructed annotation does not exist, create PSP, role, binding.
+// - Check if the globalRestricted annotation does not exist, check if the PSP exists and
+//   if it doesn't, create it. Check to see if the associated role and bindings exist and
+//   if they do, delete them.
+func SetPSPs(ctx *cli.Context, k8sWrapTransport transport.WrapperFunc) error {
+	const (
+		kubeConfigPath = "/etc/rancher/rke2/rke2.yaml"
+		waitDelay      = time.Second * 1
+	)
+
+	// wait until the kubeconfig file is written
+	for !kubeConfigExists(kubeConfigPath) {
+		time.Sleep(waitDelay)
+	}
+
+	cs, err := newClient(kubeConfigPath, k8sWrapTransport)
+	if err != nil {
+		return err
+	}
+
+	for {
+		// wait until kube-apiserver is running
+		if _, err := cs.Discovery().ServerVersion(); err != nil {
+			time.Sleep(waitDelay)
+			continue
+		}
+
+		// wait until the kube-system namespace is created
+		ns, err := cs.CoreV1().Namespaces().Get(context.TODO(), kubeSystemNamespace, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				time.Sleep(waitDelay)
+				continue
+			}
+			return err
+		}
+
+		if ctx.String("profile") == "" { // non-CIS mode
+			if err := setGlobalUnrestricted(cs, ns); err != nil {
+				return err
+			}
+
+			if err := setSystemUnrestricted(cs, ns); err != nil {
+				return err
+			}
+
+			if _, ok := ns.Annotations[namespaceAnnotationGlobalRestricted]; !ok {
+				// check if the policy exists
+				if _, err := cs.PolicyV1beta1().PodSecurityPolicies().Get(context.TODO(), globalRestrictedPSPName, metav1.GetOptions{}); err != nil {
+					if apierrors.IsNotFound(err) {
+						tmpl := fmt.Sprintf(globalRestrictedPSPTemplate, globalRestrictedPSPName)
+						if err := deployPodSecurityPolicyFromYaml(cs, tmpl); err != nil {
+							return err
+						}
+					}
+				}
+
+				// check if role exists and delete it
+				role, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), globalRestrictedRoleName, metav1.GetOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						logrus.Info(err)
+					} else {
+						if err := cs.RbacV1().ClusterRoles().Delete(context.TODO(), role.Name, metav1.DeleteOptions{}); err != nil {
+							return err
+						}
+					}
+				}
+
+				// check if role binding exists and delete it
+				roleBinding, err := cs.RbacV1().ClusterRoleBindings().Get(context.TODO(), globalRestrictedRoleBindingName, metav1.GetOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						logrus.Info(err)
+					}
+				} else {
+					if err := cs.RbacV1().ClusterRoleBindings().Delete(context.TODO(), roleBinding.Name, metav1.DeleteOptions{}); err != nil {
+						return err
+					}
+				}
+				ns.SetAnnotations(map[string]string{namespaceAnnotationGlobalRestricted: cisAnnotationValue})
+			}
+		} else { // CIS mode
+			if _, ok := ns.Annotations[namespaceAnnotationGlobalRestricted]; !ok {
+				if _, err := cs.PolicyV1beta1().PodSecurityPolicies().Get(context.TODO(), globalRestrictedPSPName, metav1.GetOptions{}); err != nil {
+					tmpl := fmt.Sprintf(globalRestrictedPSPTemplate, globalRestrictedPSPName)
+					if err := deployPodSecurityPolicyFromYaml(cs, tmpl); err != nil {
+						return err
+					}
+				}
+				if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), globalRestrictedRoleName, metav1.GetOptions{}); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/rke2/rke2.go
```diff
@@ -36,6 +38,12 @@ func Server(ctx *cli.Context, cfg Config) error {
 	if err := ctx.Set("secrets-encryption", "true"); err != nil {
 		return err
 	}
+	go func() {
```
cjellick, line 41: Im not sure about this being in a goroutine. I feel like maybe it should block, but I'll seek a second opinion

briandowns: I remember we discussed that but that function call is in k3s that is responsible for a few setup routines that the new code is dependent on. The new code has checks and waits for it to be put in place. eg `/etc/rancher/rke2/rke2.yaml`.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+func kubeConfigExists(kc string) bool {
+	info, err := os.Stat(kc)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+// SetPSPs sets the default PSP's based on the mode that RKE2 is running in. There is either CIS or non
+// CIS mode. For CIS mode, the globalRestricted and systemUnrestructed polices and associated roles and
```
cjellick, line 128: systemUnrestr**U**cted is misspelled several times in this comment

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+	namespaceAnnotationGlobalRestricted   = namespaceAnnotationBase + "global-restricted"
+	namespaceAnnotationGlobalUnrestricted = namespaceAnnotationBase + "global-unrestricted"
+	namespaceAnnotationSystemUnrestricted = namespaceAnnotationBase + "system-unrestricted"
+)
+
+// setGlobalUnrestricted sets the global unrestricted podSecurityPolicy with
+// the associated role and rolebinding.
+func setGlobalUnrestricted(cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	if _, ok := ns.Annotations[namespaceAnnotationGlobalUnrestricted]; !ok {
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), globalUnrestrictedPSPName, metav1.GetOptions{}); err != nil {
```
cjellick, line 44: `context.TODO()`
are we not setting up a proper signal handling context and passing it around in rke2?
Id expect something like this to exist: https://github.com/rancher/k3s/blob/master/pkg/cli/server/server.go#L219

Looks like this PR is the first to need to create a context, so i think it unfortunately falls on you to pull in  a proper  signalhandling context

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+// bindings are created. For non CIS mode, globalRestricted, system-unrestricted, and global-unrestricted
+// policies and associated roles and bindings are loaded, if applicable.
+//
+// Each PSP has an annotation associated with it that we check to see exists before performing any
+// write operations. The annotation is created upon successful setting of PSPs.
+//
+// Load logic
+//
+// CIS:
+// - Check if the globalRestricted annotation does not exist, create PSP, role, binding.
```
cjellick, line 138: Very pedantic nitpick: This reads oddly; the "check" is throwing me off.
It would be a little easier to read for me if it followed if/then, like: "If the globalRestricted annotation does not exist, then create PSP, role, & binding."

This is just my opinion; if you like it as-is, don't feel obliged to change it.

pkg/rke2/rke2.go
dweomer: Seems like the right approach to me, re: the goroutine.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+
+const (
+	defaultRetries          = 5
+	defaultWaitSeconds      = 5
+	defaultTimeout          = 30
+	k8sWrapTransportTimeout = 30
+)
+
+const (
+	kubeSystemNamespace = "kube-system"
```
dweomer, line 31: This const isnt necessary, use `metav1.NamespaceSystem`

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+//
+// non-CIS:
+// - If the globalUnrestricted annotation does not exist, then create PSP, role, binding.
+// - If the systemUnrestricted annotation does not exist, then create PSP, role, binding.
+// - If the globalRestricted annotation does not exist, then check if the PSP exists and
+//   if it doesn't, create it. Check if the associated role and bindings exist and
+//   if they do, delete them.
+func SetPSPs(sCtx context.Context, ctx *cli.Context, k8sWrapTransport transport.WrapperFunc) error {
+	const (
+		kubeConfigPath = "/etc/rancher/rke2/rke2.yaml"
```
cjellick, line 151: DISCLAIMER: I don't have a solution on-hand. Just noting my concerns at this point. Hopefully we can brainstorm a solution today.

I know we dont have a better technique on hand, but relying on a hard-coded location of the config doesn't really feel right. k3s's server.run() method (and the structs available therein) have much better approaches to similar. For example, `config.ControlConfig.Runtime.KubeConfigAdmin` gives you the config.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+
+	cs, err := newClient(kubeConfigPath, k8sWrapTransport)
+	if err != nil {
+		return err
+	}
+
+	for {
+		// wait until kube-apiserver is running
+		if _, err := cs.Discovery().ServerVersion(); err != nil {
```
cjellick, line 167: same DiSCLAIMER, but similarly, this has a better solution in k3s: `serverConfig.ControlConfig.Runtime.APIServerReady`

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+							return err
+						}
+					}
+				}
+
+				// check if role exists and delete it
+				role, err := cs.RbacV1().ClusterRoles().Get(sCtx, globalRestrictedRoleName, metav1.GetOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						logrus.Info(err)
```
cjellick, line 206: An info level log of an error with no additional context explaining it is really unhelpful. Ive see log lines like this pop up and they just leave you scratching your head.

pkg/rke2/psp/psp.go
```diff
@@ -135,18 +135,18 @@ func kubeConfigExists(kc string) bool {
...
-// - Check if the systemUnrestricted annotation does not exist, create PSP, role, binding.
-// - Check if the globalRestricted annotation does not exist, check if the PSP exists and
-//   if it doesn't, create it. Check to see if the associated role and bindings exist and
+// - If the globalUnrestricted annotation does not exist, then create PSP, role, binding.
+// - If the systemUnrestricted annotation does not exist, then create PSP, role, binding.
+// - If the globalRestricted annotation does not exist, then check if the PSP exists and
+//   if it doesn't, create it. Check if the associated role and bindings exist and
 //   if they do, delete them.
-func SetPSPs(ctx *cli.Context, k8sWrapTransport transport.WrapperFunc) error {
+func SetPSPs(sCtx context.Context, ctx *cli.Context, k8sWrapTransport transport.WrapperFunc) error {
```
cjellick, line 149: why sCtx and not just ctx? There's a pretty strong convention for just calling the passed in context "ctx" and as far as this function is concerned, it knows nothing of the "s" part, which only the caller knows is related to signal handling.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+	config.Timeout = time.Second * defaultTimeout
+	return kubernetes.NewForConfig(config)
+}
+
+func decodeYamlResource(data interface{}, yaml string) error {
+	decoder := yamlutil.NewYAMLToJSONDecoder(bytes.NewReader([]byte(yaml)))
+	return decoder.Decode(data)
+}
+
+func retryTo(runFunc deployFn, ctx context.Context, cs *kubernetes.Clientset, resource interface{}, retries, wait int) error {
```
cjellick, line 331: Can we use k8s.io/client-go/util/retry instead?

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+	namespaceAnnotationGlobalRestricted   = namespaceAnnotationBase + "global-restricted"
+	namespaceAnnotationGlobalUnrestricted = namespaceAnnotationBase + "global-unrestricted"
+	namespaceAnnotationSystemUnrestricted = namespaceAnnotationBase + "system-unrestricted"
+)
+
+// setGlobalUnrestricted sets the global unrestricted podSecurityPolicy with
+// the associated role and rolebinding.
+func setGlobalUnrestricted(ctx context.Context, cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	if _, ok := ns.Annotations[namespaceAnnotationGlobalUnrestricted]; !ok {
+		if _, err := cs.RbacV1().ClusterRoles().Get(ctx, globalUnrestrictedPSPName, metav1.GetOptions{}); err != nil {
```
cjellick, line 44: Currently, this just swallows the error if it is anything other than IsNotFound. Those should probably get bubbled up instead. (This applies to multiple places)

pkg/rke2/psp/psp_templates.go
```diff
@@ -0,0 +1,199 @@
...
+  name: %s
+rules:
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - %s
+`
+
+const globalRestrictedPSPTemplate = `apiVersion: policy/v1beta1
```
cjellick, line 30: After seeing how this looks, I think we could consider switching from string constants to just declaring the actual instances structs. It would save on the decoding and provide a bit of type safety.

In my mind, the main justification for string constants would be readability, but I think the structs do not lose enough in terms of readability. Is there another good reason for string constants that would tip the scales?

Sample struct:
```
var falseVal = false
var psp1 = v1beta1.PodSecurityPolicy{
	Spec: v1beta1.PodSecurityPolicySpec{
		Privileged: false,
		AllowPrivilegeEscalation: &falseVal,
		RunAsUser:                       v1beta1.RunAsUserStrategyOptions{
			Rule: v1beta1.RunAsUserStrategyMustRunAsNonRoot,
		},
           ...
```
feels relatively readable (assuming you alrady no go)

cjellick: I don't know. I'm wavering on this. the string constant is probably fine

pkg/rke2/psp/psp.go
briandowns: Thanks!

pkg/rke2/psp/psp_templates.go
briandowns: I'm happy to change it. I was just mirroring RKE.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+							return err
+						}
+					}
+				}
+
+				// check if role binding exists and delete it
+				roleBinding, err := cs.RbacV1().ClusterRoleBindings().Get(sCtx, globalRestrictedRoleBindingName, metav1.GetOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						logrus.Info(err)
```
cjellick, line 217: I wouldn't even log this error. This is "normal" state after the initial boostrapping, so isNotFound can just be swallowed

briandowns: Refactored block:

```go
role, err := cs.RbacV1().ClusterRoles().Get(sCtx, globalRestrictedRoleName, metav1.GetOptions{})
				if err != nil && !apierrors.IsNotFound(err) {
					return err
				}
				if err := cs.RbacV1().ClusterRoles().Delete(sCtx, role.Name, metav1.DeleteOptions{}); err != nil {
					return err
				}
```

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+					}
+				} else {
+					if err := cs.RbacV1().ClusterRoleBindings().Delete(sCtx, roleBinding.Name, metav1.DeleteOptions{}); err != nil {
+						return err
```
cjellick, line 221: In an HA multi-server setup, you will get into scenarios where more than one server is attempting to run this logic, potentially creating race conditions. For that reason, I think you should check the error here. If it is the DoesNotExist error (or whatever error is returned when you do a delete on a thing that doesnt exist), then dont return the error

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+				}
+				if _, err := cs.RbacV1().ClusterRoleBindings().Get(sCtx, globalRestrictedRoleBindingName, metav1.GetOptions{}); err != nil {
+					if apierrors.IsNotFound(err) {
+						tmpl := fmt.Sprintf(globalRestrictedRoleBindingTemplate, globalRestrictedRoleBindingName, globalRestrictedRoleName)
+						if err := deployClusterRoleBindingFromYaml(sCtx, cs, tmpl); err != nil {
+							return err
+						}
+					}
+				}
+				ns.SetAnnotations(map[string]string{namespaceAnnotationGlobalRestricted: cisAnnotationValue})
```
cjellick, line 250: You are stomping on all existing annotations on the namespace. can't do that

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+			} else {
+				tmpl := fmt.Sprintf(nodeClusterRoleBindingTemplate, globalUnrestrictedPSPName)
+				if err := deployClusterRoleBindingFromYaml(sCtx, cs, tmpl); err != nil {
+					return err
+				}
+			}
+		}
+		return nil
+	}
+}
```
cjellick, line 308: I don't seen anywhere in this function where you are actually saving the updates to the namespace

pkg/rke2/psp/psp.go
briandowns: I agree. I couldn't find any better way at that point.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,443 @@
...
+	for !kubeConfigExists(kubeConfigPath) {
+		time.Sleep(waitDelay)
+	}
+
+	cs, err := newClient(kubeConfigPath, k8sWrapTransport)
+	if err != nil {
+		return err
+	}
+
+	for {
```
cjellick, line 164: Wrapping the entirety of this function in an infinite loop is concerning. What logic are you trying to retry? Is it just the waiting for the client or is it to do retries around specific client calls as well? 

In a non-error scenario, as best I can tell, you never exit this loop. Is that your intention?

pkg/rke2/psp/psp.go
briandowns: Sure. This was just lifted from RKE.

pkg/rke2/psp/psp.go
cjellick: > In a non-error scenario, as best I can tell, you never exit this loop. Is that your intention?

That's not true...you do have a return nil in the last if block

pkg/rke2/psp/psp.go
cjellick: looks good, though i make another comment further down that to avoid race conditions you should check if the error post delete is due to the resource not existing and if thats the case, ignore the error

pkg/rke2/psp/psp.go
briandowns: I've pushed error handling changes.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,442 @@
...
+						}
+					}
+				}
+
+				ns.SetAnnotations(map[string]string{namespaceAnnotationGlobalUnrestricted: cisAnnotationValue})
+			}
+		}
+
+		// apply node cluster role binding regardless of whether we're in CIS mode or not
+		if _, err := cs.RbacV1().ClusterRoleBindings().Get(sCtx, globalUnrestrictedRoleName, metav1.GetOptions{}); err != nil {
```
cjellick, line 288: Create a binding for global unrestricted regardless of whether we are in CIS mode?
You don't want the unrestricted policy applied when you're in CIS mode, do you?

cjellick:
Ok, I've said a lot in this PR. Going to break it down between must haves and nice-to haves. I'll go through my comments and mark them accordingly.

pkg/rke2/psp/psp.go
briandowns: This was an error in the name. Resolved.

pkg/rke2/psp/psp.go
cjellick: must-have. hopefully a small change and want to get error handling right

pkg/rke2/psp/psp.go
cjellick: nice-to-have, but hopefully would be a simple replace

pkg/rke2/psp/psp.go
cjellick: must-have BUT I dont have a better solution. I just want to go over this one with @ibuildthecloud, but he said not until tomorrow

pkg/rke2/psp/psp.go
cjellick: must-have BUT I dont have a better solution. I just want to go over this one with @ibuildthecloud, but he said not until tomorrow. after reviewing with him, maybe we leave as-is

pkg/rke2/psp/psp.go
cjellick: must-have - its a bug thatll burn us later

pkg/rke2/psp/psp.go
cjellick: must-have, unless im missing something

pkg/rke2/psp/psp.go
cjellick: nice-to-have. can be left for another day if time is an issue

pkg/rke2/psp/psp_templates.go
cjellick: nice-to-have. in fact, I think a case could be made either way, so you can leave as-is

pkg/rke2/rke2.go
cjellick: must-have - but needs a conversation with Darren.

This is really a bigger conversation about how RKE2 can tie into the server bootstrapping process in a clean way. This sets the precedence for how we do that. Right now, the k3s code exposes no hooks for doing this cleanly.

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,434 @@
...
+		ns.SetAnnotations(map[string]string{namespaceAnnotationGlobalUnrestricted: cisAnnotationValue})
+	}
+	return nil
+}
+
+// setSystemUnrestricted sets the system unrestricted podSecurityPolicy as
+// the associated role and rolebinding.
+func setSystemUnrestricted(ctx context.Context, cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	if _, ok := ns.Annotations[namespaceAnnotationSystemUnrestricted]; !ok {
+		if _, err := cs.RbacV1().ClusterRoles().Get(context.TODO(), systemUnrestrictedPSPName, metav1.GetOptions{}); err != nil {
```
cjellick, line 79: mistakenly passing wrong context?

pkg/rke2/psp/psp.go
```diff
@@ -0,0 +1,434 @@
...
+func deployPodSecurityPolicy(ctx context.Context, cs *kubernetes.Clientset, p interface{}) error {
+	psp, ok := p.(v1beta1.PodSecurityPolicy)
+	if !ok {
+		return fmt.Errorf("invalid type provided: %T, expected: PodSecurityPolicy", p)
+	}
+	if _, err := cs.PolicyV1beta1().PodSecurityPolicies().Create(ctx, &psp, metav1.CreateOptions{}); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		if _, err := cs.PolicyV1beta1().PodSecurityPolicies().Update(context.TODO(), &psp, metav1.UpdateOptions{}); err != nil {
```
cjellick, line 357: mistakenly passing wrong context?

pkg/rke2/psp/psp.go
briandowns: Was not to trample the use of `ctx` throughout the repo as the ctx for the CLI args. I'm happy to switch to the more accepted convention for ctx and update all of the cli ctx refs.

pkg/rke2/psp/psp.go
briandowns: We have the `--write-kubeconfig` flag that I could possibly reference but it looks like rke2 explicitly writes that config to that location and this flag will create a symlink to where the user may want to reference it.

pkg/rke2/psp/psp.go
briandowns: This appears to be a no go. k3s uses it and it's implementation panics if it's called twice so calling it before we get to k3s isn't going to work.

pkg/rke2/psp/psp.go
cjellick: i see. fine as-is

pkg/rke2/psp/psp.go
briandowns: Issue created.

pkg/rke2/psp/psp_templates.go
briandowns: Issue created.

pkg/rke2/rke2.go
briandowns: Updated.

scripts/validate
```diff
@@ -31,5 +31,5 @@ go mod tidy
 if [ -n "$DIRTY" ]; then
     echo Source dir is dirty
     git status --porcelain --untracked-files=no
-    exit 1
+    #exit 1
```
cjellick, line 34: Is this meant to be here or some debugging accidentally left in or what?
Should we just delete the line all together?

pkg/rke2/psp.go
```diff
@@ -0,0 +1,449 @@
...
+		logrus.Info("Applying PSP's...")
+		go func() {
+			<-cfg.Runtime.APIServerReady
+
+			cs, err := newClient(cfg.Runtime.KubeConfigAdmin, nil)
+			if err != nil {
+				logrus.Fatalf("psp: new k8s client: %s", err.Error())
+			}
+
+			ns, err := cs.CoreV1().Namespaces().Get(ctx, metav1.NamespaceSystem, metav1.GetOptions{})
```
cjellick, line 167: I recall you thinking that this might need to be retried in a loop. not necessary?

pkg/rke2/psp.go
```diff
@@ -0,0 +1,449 @@
...
+			if err != nil {
+				logrus.Fatalf("psp: get kube-system namespace: %s", err.Error())
+			}
+
+			annotations := make(map[string]string)
+
+			logrus.Info("Applying PSP's...")
```
cjellick, line 174: you have the same log message twice (here and line 158). Any way we can disambiguate (or delete one)?

pkg/rke2/psp.go
```diff
@@ -0,0 +1,449 @@
...
+			if err != nil {
+				logrus.Fatalf("psp: new k8s client: %s", err.Error())
+			}
+
+			ns, err := cs.CoreV1().Namespaces().Get(ctx, metav1.NamespaceSystem, metav1.GetOptions{})
+			if err != nil {
+				logrus.Fatalf("psp: get kube-system namespace: %s", err.Error())
+			}
+
+			annotations := make(map[string]string)
```
cjellick, line 172: I think you create a new map here, then you add annotations to it in the logic, then at the end you call ns.SetAnnotations()

I think this will cause you to fully overwrite all annotations on the namespace. This is a problem bc in the situation where the user is trying to "reset" their PSP setup, the namespace might have many other annotations that we dont want to lose.

So, instead of creating a new map, you have to initialize this to annotations on the current map (or new map if that is nil i guess)

pkg/rke2/psp.go
briandowns: Now, not necessary.

cjellick:
It would be nice to have an info level log statement any time we are about to create or delete  a psp/binding/role. Just something simple like "Creating globalUnrestricted PSP"

This is actually a general critique i have of k3s and rke2 in general (really a lot of rancher) and every time I'm on a support call, I wish we have a bit more logging.

pkg/rke2/psp.go
briandowns: Initialized map and passing that in. Confirmed it doesn't overwrite existing annotations.

Issue #175, opened 2020-08-11
Change release artifacts and refactor install.sh script


davidnuzik:
Based on our design discussions and talks with Darren, we'd like to change how RKE2 release works (artifacts from release process) and how the install.sh script works.

There is a desire to **simplify the install.sh script**. We instead should just have a tarball artifact for RKE2 -- basically including the structure of RKE2 (/etc/rancher, /var/lib/rancher, etc) and deploy via the install script. Tarbal should contain full dir structure: systemctl files, uninstall, killall, binary.
The install script should just extract tarball or do yum repo and rpm and package installs.

The install script should do **CIS mitigation but it will not create the etcd user** any longer.

The unit file should also be separated out. The install script can deploy the **unit file** but it should be a separate release artifact.

The install script should be cleaned up/simplified so it's not making so many changes to the system. It should essentially be a simple means to install rke2 onto the machine with minimal modifications to the system - the exception being the light CIS mitigation (sans create etcd user). **kill-all and uninstall script** should be separate release artifacts and not a part of the install.sh script.

Install.sh should be it's own artifact.

ShylajaDevadiga:
Need to pass --tls-san through flags or env variable to access cluster outside node. INSTALL_RKE2_EXEC variable is missing. Issue tracked in #258

ShylajaDevadiga:
@dweomer install script does not have the INSTALL_RKE2_SKIP_DOWNLOAD variable. When we do installation using private registry in airgapped setup having INSTALL_RKE2_SKIP_DOWNLOAD=true would enable running install script.

dweomer:
> @dweomer install script does not have the INSTALL_RKE2_SKIP_DOWNLOAD variable. When we do installation using private registry in airgapped setup having INSTALL_RKE2_SKIP_DOWNLOAD=true would enable running install script.

In the spirit of making the top-level install script less complicated, this functionality was indeed dropped. If you are installing for airgap scenario then I would assume you have the installer downloaded already (which is all the top-level install.sh is doing for you anyway) therefor you should be running the installer directly.

ShylajaDevadiga:
Validated install.sh using beta16/beta17.
Simple install using refactored install script
```
sudo INSTALL_RKE2_VERSION=v1.18.4-beta17+rke2 INSTALL_RKE2_TYPE='server' INSTALL_RKE2_METHOD='installer' ./install.sh
```
Passing CIS profile
```
[...]

Issue #138, opened 2020-07-28
Remove the install.sh script


davidnuzik:
The focus for rke2 has shifted slightly. We will be focusing on our RPM packages going forward and this shall be the primary method of installing RKE2. We also would like to create a DEB package as well, however this will come later and we do not have a specific date in mind for this.

We are going to remove the install.sh script. Any CIS mitigation will be in the RPM now and the RPM (for now) will be the recommended way to install RKE2.

davidnuzik:
We are probably just going to refactor the install script as per: https://github.com/rancher/rke2/issues/175

Still, I'm keeping this issue open for now as we work out the details. Small chance we change our minds still and indeed want to delete the install.sh script (instead supply RPM/DEB)

PR #220, opened 2020-08-21
installer: refactor install script

dweomer:
Make the install script simpler by leveraging a self-extracting tarball
built with [`makeself`](https://github.com/megastep/makeself). This adds a new release artifact,
rke2-installer.${GOOS}-${GOARCH}.run, that is a self-extracting version
of the tarball artifact. The install, killall, and uninstall logic has
been moved into this artifact which ends up on hosts, by default, in the
/usr/local/share/rke2/scripts directory.

The install.sh at the root of the repository has been simplified. It
will attempt to detect if yum is available and install via that method.
Lacking yum, it will download and invoke the self-extracting tarball.

Addresses #175
Moots #138 as `install.sh` has been made simpler such that it attempts to install via rpm if yum is on the system, generic installer otherwise.


dweomer:
`install.sh` works for me in:
- `centos:7` container (installing via yum by default, but installer override works too)
- `alpine:edge` container (installer only)

Issue #228, opened 2020-08-24
Build from master fails at make image


ShylajaDevadiga:
Version:
Build after beta1
Here are the steps followed:
```
pt-get update 
apt install docker.io
systemctl start docker
apt install gcc
snap install go --classic
apt install make
git clone https://github.com/rancher/rke2.git
cd rke2/
make clean build 
make image
make: *** No rule to make target 'image'.  Stop.
```

dweomer:
Fixed via #214

davidnuzik:
This should be fixed - needs re-test from QA.

davidnuzik:
zubebot appears to be acting up - is undoing things I change.

ShylajaDevadiga:
Validated. Thanks

PR #229, opened 2020-08-24
docs: fix BUILDING.md

dweomer:
Be specific about which image tarballs to copy when testing air-gapped
vs non-air-gapped.


dweomer:
Adds some clarity that should help fully resolve #228

PR #236, opened 2020-08-26
Add a cachebust to ensure we are always pulling the latest charts

Oats87:
Add cachebust to ensure that we are always pulling the latest charts
Modify the rm command to explicity delete `.sh` and `.md` files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,12 +64,12 @@ COPY --from=build-k8s \
 FROM build AS charts
 ARG CHARTS_REPO="https://rke2-charts.rancher.io"
 COPY charts/ /charts/
-RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN CHART_VERSION="v1.18.4"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -64,12 +64,12 @@ COPY --from=build-k8s \
 FROM build AS charts
 ARG CHARTS_REPO="https://rke2-charts.rancher.io"
 COPY charts/ /charts/
-RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN CHART_VERSION="v1.18.4"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN rm -vf /charts/*.{sh,md}
+RUN echo CACHEBUST>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
```
brandond, line 67: Shouldn't this be a build arg variable that gets set to something random or time-based when building (`docker build --build-arg CACHEBUST=$(date)`? Otherwise you're just echoing the word CACHEBUST into /dev/null every time it runs, which I don't believe will invalidate the layer cache...

brandond: For the record, this works for k/k because they preprocess the Dockerfile with sed and replace CACHEBUST with something random every build:
https://github.com/vwfs/fork-kubernetes/blob/ee8caaf3e1a7ab648949a7bf036a288ef0516fe4/cluster/images/hyperkube/Makefile#L83

dweomer: Yes, it should be a build arg but we needed a short term fix. I've got a PR i am working on that can work in a fix.

Issue #238, opened 2020-08-26
Invalidate chart layers of Dockerfile selectively without having to rebuild the rest of the image


Oats87:
https://github.com/rancher/rke2/pull/236 was committed/merged in haste in order to invalidate the cache for a rebuild of chart content. We should invalidate the cache in a manner that is repeatable for every build run.

cjellick:
@dweomer said in the associated pr that he has a longer term fix in the works. im going to assign him

davidnuzik:
@dweomer is there an associated PR or WIP PR yet? This is in working state and note Next Up state?

brandond:
We should revisit this; doing so could make builds a good bit faster due to layer reuse.

caroline-suse-rancher:
@briandowns Could this potentially be work for the ecm-tools team?

PR #241, opened 2020-08-26
Add a proper cachebust by echoing a build arg set to the current date…

Oats87:
Add a proper cachebust by echoing a build arg set to the current date to /dev/null

https://github.com/rancher/rke2/issues/238

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,12 +63,14 @@ COPY --from=build-k8s \
 
 FROM build AS charts
 ARG CHARTS_REPO="https://rke2-charts.rancher.io"
+ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v1.18.4"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN echo ${CACHEBUST}>/dev/null
+RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
+RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -63,12 +63,13 @@ COPY --from=build-k8s \
...
 FROM build AS charts
 ARG CHARTS_REPO="https://rke2-charts.rancher.io"
+ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v1.18.4"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN echo $CACHEBUST>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
```
briandowns, line 68: I think the convention is to use the curl brace around the variable name for these.

Dockerfile
```diff
@@ -63,12 +63,13 @@ COPY --from=build-k8s \
...
 FROM build AS charts
 ARG CHARTS_REPO="https://rke2-charts.rancher.io"
+ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="v1.18.4"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN echo CACHEBUST>/dev/null; CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN echo ${CACHEBUST}>/dev/null; CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
```
dweomer, line 68: please refactor to have the echo on a line by itself before any charts are built. this is enough to invalidate all subsequent layers

scripts/build-image-runtime
```diff
@@ -8,6 +8,7 @@ source ./scripts/version.sh
 DOCKER_BUILDKIT=${DOCKER_BUILDKIT:-1} docker image build \
     --build-arg TAG=${VERSION} \
     --build-arg KUBERNETES_VERSION=${KUBERNETES_VERSION} \
+    --build-arg CACHEBUST=$(date +%s%N) \
```
dweomer, line 11: CACHEBUST should default to the value of $TAG unless specified in the env. this can be done in the version.sh script

dweomer: this will allow for caching by default but guarantee a cache miss when building tagged releases (except for re-tags, which we shouldnt be doing anyway). this way you have the option of overridding CACHEBUST in the drone CI (or locally) to force it

Oats87: As per offline discussion, we will use the `date` approach for now, and eventually move to a more standardized method that allows for selective cache invalidation.

Issue #242, opened 2020-08-26
Create public channel-server and use it in install script


davidnuzik:
Similar to K3s, we'd like to have the ability to have different channels / streams for releases. For example, the ability to specify a stable release is beneficial.

cjellick:
@davidnuzik please link this in the EIO ticket so it all gets hooked up

davidnuzik:
Linking issues and PRs:

https://github.com/rancher/rancher/pull/28596 - PR to get channels.yaml added to Rancher project
https://github.com/rancher/rke2/pull/246 - PR to get channels.yaml added to RKE2 project

Internal EIO ticket linked. `#226`

davidnuzik:
EIO ticket completed.

cjellick:
OK, the channel server service is up and running. install script should move over to using that.

Check out the channels because right now, there is no release in the "latest" channel, only "testing" (which makes sense).

ShylajaDevadiga:
Validated using rke2 version v1.18.9-beta22+rke2
Channel can be passed in install script
CHANNE=testing accepts the latest beta release with tarball install as well as rpm install
Validated default CHANNEL is set to testing if INSTALL_RKE2_CHANNEL is not passed.

tarball install
```
sudo INSTALL_RKE2_CHANNEL=testing sh ./install.sh
[INFO]  finding release for channel testing
[INFO]  using v1.18.9-beta22+rke2 as release
[INFO]  downloading checksums at https://github.com/rancher/rke2/releases/download/v1.18.9-beta22+rke2/sha256sum-amd64.txt
[INFO]  downloading tarball at https://github.com/rancher/rke2/releases/download/v1.18.9-beta22+rke2/rke2.linux-amd64.tar.gz
[INFO]  verifying installer
[INFO]  unpacking tarball file
```
rpm install
```
sudo INSTALL_RKE2_CHANNEL=testing sh ./install.sh
Loaded plugins: amazon-id, product-id, search-disabled-repos, subscription-manager
rancher-rke2-1-18-testing                                                                                                                                         | 2.9 kB  00:00:00     
rancher-rke2-common-testing                                                                                                                                       | 2.9 kB  \                                                                                                                                                | 3.5 kB  00:00:00     
...
Resolving Dependencies
--> Running transaction check
---> Package rke2-server.x86_64 0:1.18.9~beta22-1.el7 will be installed
```
Default without passing CHANNEL
```
sudo sh ./install.sh 
[...]

PR #246, opened 2020-08-27
Add channels.yaml file

cjellick:
In preparation for having a rke2 channel server, add channels.yaml

First half of: https://github.com/rancher/rke2/issues/242

Once this is merged, EIO will have to actually stand up the service, point it at this file, and put dns at it.

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,0 +1,10 @@
+channels:
+- name: latest
+  latestRegexp: .*
+  excludeRegexp: ^[^+]+-
+- name: testing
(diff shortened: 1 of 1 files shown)


channels.yaml
```diff
@@ -0,0 +1,11 @@
+# Example channels config
```
briandowns, line 1: Is this an example or an actual config?

cjellick: i simultaneously noticed that as well. removed it. copied from the k3s config (which should also drop this comment)

Issue #244, opened 2020-08-26
Install fails when manifests dir is present before installation


ShylajaDevadiga:
Version:
rke2 version v1.18.4-dev+e3330898

Describe the issue:
Install fails due to the existence of manifests dir after extracting tar file

Logs:
```
journalctl -u rke2
-- Logs begin at Wed 2020-08-26 01:15:52 UTC, end at Wed 2020-08-26 22:33:39 UTC. --
Aug 26 01:54:14 ip-172-31-6-12 systemd[1]: Starting Rancher Kubernetes Engine v2...
Aug 26 01:54:14 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:14Z" level=warning msg="not running in CIS 1.5 mode"
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/kubectl..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/kubelet..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/containerd..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/containerd-shim..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/containerd-shim-runc-v1..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/containerd-shim-runc-v2..."
Aug 26 01:54:19 ip-172-31-6-12 rke2[3063]: time="2020-08-26T01:54:19Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-dev-e3330898 bin/ctr..."
[...]

PR #252, opened 2020-08-27
fix bootstrap rename manifests error

MonzElmasry:
issue : https://github.com/rancher/rke2/issues/244

### DIFF
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -172,7 +172,7 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 			return err
 		}
 		for _, file := range fileInfo {
-			if err := os.Rename(tempDir+"/"+file.Name(), dir+"/"+file.Name()); err != nil {
+			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err != nil {
 				return err
 			}
 		}
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -12,9 +12,8 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/rancher/wrangler/pkg/merr"
-
 	errors2 "github.com/pkg/errors"
+	"github.com/rancher/wrangler/pkg/merr"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -166,17 +166,22 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	if err := os.Rename(tempDir, dir); err != nil {
+	if err := os.Rename(tempDir, dir); err == os.ErrExist {
+		logrus.Print(err)
 		files, err := ioutil.ReadDir(tempDir)
 		if err != nil {
 			return err
 		}
 		for _, file := range files {
-			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err != nil {
+			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err == os.ErrExist {
 				os.Remove(filepath.Join(dir, file.Name()))
 				os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name()))
+			} else if err != nil {
+				return err
 			}
 		}
+	} else if err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -166,7 +166,18 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil {
+		fileInfo, err := ioutil.ReadDir(tempDir)
+		if err != nil {
+			return err
+		}
+		for _, file := range fileInfo {
+			if err := os.Rename(tempDir+"/"+file.Name(), dir+"/"+file.Name()); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func preloadBootstrapImage(dataDir, runtimeImage string) (v1.Image, error) {
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -167,7 +167,6 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 		return err
 	}
 	if err := os.Rename(tempDir, dir); err == os.ErrExist {
-		logrus.Print(err)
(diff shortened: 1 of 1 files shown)


pkg/bootstrap/stage.go
```diff
@@ -166,7 +166,18 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 		return err
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil {
+		fileInfo, err := ioutil.ReadDir(tempDir)
+		if err != nil {
+			return err
+		}
+		for _, file := range fileInfo {
+			if err := os.Rename(tempDir+"/"+file.Name(), dir+"/"+file.Name()); err != nil {
```
briandowns, line 175: Take a look at: https://golang.org/pkg/path/filepath/#Join

briandowns, line 175: What do we do if a file already exists?

pkg/bootstrap/stage.go
```diff
@@ -166,7 +166,18 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil {
+		fileInfo, err := ioutil.ReadDir(tempDir)
```
briandowns, line 170:
```suggestion
		files, err := ioutil.ReadDir(tempDir)
```

`fileInfo` is a little misleading as `ioutil.ReadDir` returns a slice of FileInfo's.

pkg/bootstrap/stage.go
```diff
@@ -166,7 +166,19 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil {
```
brandond, line 169: Should check that err is due to target directory already existing. Could be due to permissions, etc.

pkg/bootstrap/stage.go
```diff
@@ -166,7 +166,19 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
+			return err
+		}
+		for _, file := range files {
+			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err != nil {
+				os.Remove(filepath.Join(dir, file.Name()))
+				os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name()))
+			}
+		}
+	}
+	return nil
```
brandond, line 181: Need to remove the temp directory, since we didn't rename it.

MonzElmasry: https://github.com/rancher/rke2/blob/43f82980a947857fddf4f8fec2cf1ee4aec288b9/pkg/bootstrap/stage.go#L161

pkg/bootstrap/stage.go
```diff
@@ -166,7 +166,19 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil {
+		files, err := ioutil.ReadDir(tempDir)
+		if err != nil {
+			return err
+		}
+		for _, file := range files {
+			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err != nil {
+				os.Remove(filepath.Join(dir, file.Name()))
```
briandowns, line 176: We should at least log the errors returned here if any. Possible escalate and stop execution.

brandond: Could accumulate errors like we do here: https://github.com/rancher/k3s/blob/master/pkg/deploy/controller.go#L121

pkg/bootstrap/stage.go
```diff
@@ -166,17 +166,22 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	if err := os.Rename(tempDir, dir); err != nil {
+	if err := os.Rename(tempDir, dir); err == os.ErrExist {
+		logrus.Print(err)
```
brandond, line 170: I don't think we should print a message if the directory already exists. It's going to exist most of the time.

pkg/bootstrap/stage.go
```diff
@@ -166,17 +166,22 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 		if err != nil {
 			return err
 		}
 		for _, file := range files {
-			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err != nil {
+			if err := os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name())); err == os.ErrExist {
 				os.Remove(filepath.Join(dir, file.Name()))
 				os.Rename(filepath.Join(tempDir, file.Name()), filepath.Join(dir, file.Name()))
+			} else if err != nil {
+				return err
```
brandond, line 180: Accumulate errors and return them together using github.com/rancher/wrangler/pkg/merr, like the deploy controller does.

pkg/bootstrap/stage.go
dweomer: bail on first unrecoverable error condition, imo

brandond: I kinda like accumulating and then reporting them together so that the user has a complete list of problems, rather than having to fix one thing and then retry so it can fail on the next one.

cjellick: just realized i never hit "send" on my comment! But it was the same thing: these errors need handled.

I'm with @dweomer: just bail on first one. At least in this scenario, I think that gets the job done simply enough without adding the complexity of gathering and properly displaying multiple errors.

pkg/bootstrap/stage.go
cjellick: ah well, if theres a ready made tool for it in our own codebase, i take back my comment. I don't feel to strongly either way for return on first or accumulate

pkg/bootstrap/stage.go
```diff
@@ -12,6 +12,10 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/rancher/wrangler/pkg/merr"
+
```
briandowns, line 16: Remove new line

pkg/bootstrap/stage.go
```diff
@@ -166,7 +170,36 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	return os.Rename(tempDir, dir)
+	if err := os.Rename(tempDir, dir); err != nil && err != os.ErrExist {
+		return err
+	} else if err == nil {
```
briandowns, line 175: I don't think this else is necessary.

MonzElmasry: the rest of the code here should work only if there's an os.ErrExist, shouldnt continue when there's no returned errors

brandond:
```suggestion
	} else if err != nil {
```

brandond: Right, if we renamed it successfully then we're done (can return) instead of continuing on to read/rename the files.

cjellick:
Maaaaybe also addresses https://github.com/rancher/rke2/issues/259

Issue #208, opened 2020-08-19
Need network policies in default namespaces when in cis mode


cjellick:
cis has a control that requires each namespace to have a network policy (control 5.3.2). We currently don't lay any policies down.

We should add logic to setup network policies for each of the built-in namespaces (default, kube-system, kube-node-lease) that locks the networking down to just that namespace.

briandowns:
Instructions for verification can be found in the PR #254

rancher-max:
#### Validated in `beta14`

- Default is to have a network policy whether CIS mode or not
- Network policies are present in the three mentioned namespaces:
```
$ kubetl get networkpolicy -A
NAMESPACE     NAME                     POD-SELECTOR   AGE
default       default-network-policy   <none>         56s
kube-public   default-network-policy   <none>         56s
kube-system   default-network-policy   <none>         56s
```
- The three mentioned namespaces have the correct annotation (note this is from a non-CIS setup as you can tell by the PSP's that are applied on the kube-system namespace, but other than that the data is the same for a CIS setup):
```
$ k describe namespace
Name:         default
Labels:       <none>
Annotations:  np.rke2.io: resolved
Status:       Active

No resource quota.

No LimitRange resource.


Name:         kube-node-lease
Labels:       <none>
Annotations:  <none>
Status:       Active

No resource quota.

No LimitRange resource.


Name:         kube-public
Labels:       <none>
Annotations:  np.rke2.io: resolved
Status:       Active

No resource quota.

No LimitRange resource.


Name:         kube-system
Labels:       <none>
Annotations:  np.rke2.io: resolved
              psp.rke2.io/global-restricted: resolved
              psp.rke2.io/global-unrestricted: resolved
              psp.rke2.io/system-unrestricted: resolved
Status:       Active

No resource quota.

No LimitRange resource.
```
- Intra-namespace traffic works as expected, and inter-namespace traffic is blocked as expected. Validated by deploying two simple nginx services and deployments in default namespace. Ensured they could reach each other by executing `curl <service clusterip>` while exec'ed into one of the pods. Ensured that exec'ing into a pod in the kube-system namespace could not access either of these endpoints. (This scenario is possible when these network policies are not present, so this means that the network policies are working as we would expect.)

PR #254, opened 2020-08-27
Default Network Policy for CIS

briandowns:
Adds a default network policy to the kube-system, default, and kube-public namespaces to adhere to CIS 1.5 compliance requirements. Addresses #208 

The default policy is set to allow only inter-namespace, pod to pod traffic.

Verification:

Check that the network policy has been created:

```sh
kubectl get networkpolicy
```

Check that annottions have been applied 

```sh
for i in "kubec-system kube-public default"; do 
    kubectl describe namespace ${i} | grep 'np.rke2.io: resolved'
done
```

Make sure that a deployed service isn't accessible outside of the namespace it's deployed to.

### DIFF
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -59,9 +59,9 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
 			if !apierrors.IsNotFound(err) {
 				return err
 			}
-			if err := cs.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, defaultNetworkPolicyName, metav1.DeleteOptions{}); err != nil {
-				return err
-			}
+		}
+		if err := cs.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, defaultNetworkPolicyName, metav1.DeleteOptions{}); err != nil {
+			return err
 		}
 		if _, err := cs.NetworkingV1().NetworkPolicies(namespace).Create(ctx, &networkPolicy, metav1.CreateOptions{}); err != nil {
 			return err
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -67,13 +67,19 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
 			return err
 		}
 		ns.Annotations[namespaceAnnotationNetworkPolicy] = cisAnnotationValue
+
 		if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 			if _, err := cs.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
-				return fmt.Errorf("networkPolicy: update namespace annotation: %w", err)
+				if apierrors.IsConflict(err) {
+					if err := updateNamespace(ctx, cs, ns); err != nil {
+						return err
+					}
+				}
+				return err
 			}
 			return nil
 		}); err != nil {
-			return err
+			logrus.Fatalf("networkPolicy: update namespace: %s - %s", ns.Name, err.Error())
 		}
 	}
 	return nil
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -48,6 +48,9 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
 	if err != nil {
 		return fmt.Errorf("networkPolicy: get %s - %w", namespace, err)
 	}
+	if ns.Annotations == nil {
+		ns.Annotations = make(map[string]string)
+	}
 	if _, ok := ns.Annotations[namespaceAnnotationNetworkPolicy]; !ok {
 		_, err := cs.NetworkingV1().NetworkPolicies(namespace).Get(ctx, defaultNetworkPolicyName, metav1.GetOptions{})
 		if err != nil {
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -1,0 +1,98 @@
+package rke2
+
+import (
+	"context"
+	"fmt"
+
+	daemonsConfig "github.com/rancher/k3s/pkg/daemons/config"
+	"github.com/rancher/spur/cli"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	namespaceAnnotationNetworkPolicy = "np.rke2.io"
+
+	defaultNetworkPolicyName = "default-network-policy"
+)
+
+// networkPolicy specifies a base level network policy applied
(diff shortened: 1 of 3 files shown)


pkg/rke2/np.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return fmt.Errorf("networkPolicy: get %s - %w", namespace, err)
+	}
+	if ns.Annotations == nil {
+		ns.Annotations = make(map[string]string)
+	}
+	if _, ok := ns.Annotations[namespaceAnnotationNetworkPolicy]; !ok {
+		_, err := cs.NetworkingV1().NetworkPolicies(namespace).Get(ctx, defaultNetworkPolicyName, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsAlreadyExists(err) {
```
dweomer, line 58: Not sure this branch will ever be followed for a GET.

pkg/rke2/np.go
```diff
@@ -0,0 +1,113 @@
...
+				if err := cs.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, defaultNetworkPolicyName, metav1.DeleteOptions{}); err != nil {
+					return err
+				}
+			} else if !apierrors.IsNotFound(err) {
+				return err
+			}
+		}
+		if _, err := cs.NetworkingV1().NetworkPolicies(namespace).Create(ctx, &networkPolicy, metav1.CreateOptions{}); err != nil {
```
dweomer, line 66: So the semantics here are intentional replacement instead of "patching" via UPDATE?

dweomer: To be clear, this is fine if that is what is required. A comment to that effect would be :+1:

pkg/rke2/psp.go
```diff
@@ -324,16 +325,46 @@ func setPSPs(clx *cli.Context) func(context.Context, daemonsConfig.Control) erro
...
+					if apierrors.IsConflict(err) {
+						if err := updateNamespace(ctx, cs, ns); err != nil {
+							return err
+						}
+					}
+					return err
+				}
+				return nil
+			}); err != nil {
+				logrus.Fatalf("psp: update namespace: %s - %s", ns.Name, err.Error())
```
dweomer, line 339: psp -> nsp ?

pkg/rke2/psp.go
```diff
@@ -324,16 +325,46 @@ func setPSPs(clx *cli.Context) func(context.Context, daemonsConfig.Control) erro
...
 			logrus.Info("Applying PSP's complete")
 		}()
 		return nil
 	}
 }
 
+// updateNamespace receives a value of type v1.Namespace pointer
+// and updates that value to point to a newly retrieve value in
+// the event a conflict error is returned.
+func updateNamespace(ctx context.Context, cs *kubernetes.Clientset, ns *v1.Namespace) error {
```
dweomer, line 350: with this func name i was expecting to see an `cs.CoreV1().Namespaces().Update()` invocation. can we rename to `getNamespace` or `cacheNamespace`, etc?

pkg/rke2/np.go
```diff
@@ -0,0 +1,113 @@
...
+
+const (
+	namespaceAnnotationNetworkPolicy = "np.rke2.io"
+
+	defaultNetworkPolicyName = "default-network-policy"
+)
+
+// networkPolicy specifies a base level network policy applied
+// to the 3 primary namespace: kube-system, default, and kube-public.
+// This policy only allows for all inter-namespace traffic.
```
cjellick, line 25: You mean it only allows traffic within the namespace, right? If so, I think you want "intra-namespace", not "inter-namespace".

pkg/rke2/np.go
cjellick: yeah, I think the delete needs to be outside of this if `err != nil` block all together and this `err != nil` should just return the error if it is any error other than `IsNotFound`.

pkg/rke2/np.go
cjellick: agree on the comment. answering intention here: yes, intention is that if the annotation is not present, this is either a fresh cluster or the user is explicitly asking for a "reset" of net policy. So, just nuke and pave.

pkg/rke2/np.go
briandowns: That's a lot cleaner.

pkg/rke2/psp.go
```diff
@@ -324,16 +325,46 @@ func setPSPs(clx *cli.Context) func(context.Context, daemonsConfig.Control) erro
...
 		return nil
 	}
 }
 
+// updateNamespaceRef receives a value of type v1.Namespace pointer
+// and updates that value to point to a newly retrieve value in
+// the event a conflict error is returned.
+func updateNamespaceRef(ctx context.Context, cs *kubernetes.Clientset, ns *v1.Namespace) error {
+	logrus.Info("updating namespace: " + ns.Name)
+	newNS, err := cs.CoreV1().Namespaces().Get(ctx, metav1.NamespaceSystem, metav1.GetOptions{})
```
brandond, line 352: Unless the intent is to update kube-system regardless of what namespace is passed in...
```suggestion
	newNS, err := cs.CoreV1().Namespaces().Get(ctx, ns.Name, metav1.GetOptions{})
```

pkg/rke2/np.go
```diff
@@ -0,0 +1,113 @@
...
+// policy exists. If it does, we delete it, and create the new default policy.
+func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clientset) error {
+	ns, err := cs.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("networkPolicy: get %s - %w", namespace, err)
+	}
+	if ns.Annotations == nil {
+		ns.Annotations = make(map[string]string)
+	}
+	if _, ok := ns.Annotations[namespaceAnnotationNetworkPolicy]; !ok {
```
brandond, line 57: Just to be clear, we don't care about annotation value, just that one exists with this key.

pkg/rke2/psp.go
briandowns: This only operates in the kube-system namespace.

pkg/rke2/np.go
briandowns: Correct.

pkg/rke2/psp.go
brandond: OK. Seems a little weird to pass in one NS and operate on another though. Maybe document this behavior in the doc comment?

pkg/rke2/np.go
```diff
@@ -0,0 +1,113 @@
...
+		if _, err := cs.NetworkingV1().NetworkPolicies(namespace).Get(ctx, defaultNetworkPolicyName, metav1.GetOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return err
+			}
+			if err := cs.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, defaultNetworkPolicyName, metav1.DeleteOptions{}); err != nil {
```
cjellick, line 62: Doesn't this need to go outside the `err != nil` block? At this point, you HAVE gotten an IsNotFound error, so there is nothing to delete. You only want to delete if it was found successfully, ie err == nil

PR #265, opened 2020-08-29
Explicitly Add k3s Flags to RKE2

briandowns:
#### Proposed Changes ####

Adds k3s flags to rke2

#### Verification ####

```sh
rke2 server -h | grep etcd
```

Also, no panic.


MonzElmasry (approved):
LGTM

Issue #248, opened 2020-08-27
Airgap install fails due to attempting to fetch image from docker.io


ShylajaDevadiga:
Version:
rke2 -v
rke2 version v1.18.4-beta11+rke2 (b459d18894e4f2a625e309f686cd95aa1cccf304)

Describe the issue:
rke2 is trying to fetch images from docker.io instead of from tarball.

Steps to reproduce:
1. Placed the rke2 binary under /usr/local/bin dir
2. Unzipped  rke2-images.linux-amd64.tar.gz and placed the tar file under /var/lib/rancher/rke2/agent/images dir
3. rke2 server --debug

Console output:
```
rke2 server  --debug
WARN[0000] not running in CIS 1.5 mode                  
FATA[0154] Get "https://index.docker.io/v2/": dial tcp 54.236.224.26:443: i/o timeout 
```

nikkelma:
Looks like the airgap image bundle is created without arch suffixes [1] (e.g. `docker.io/rancher/rke2-runtime:v1.18.4-beta13-rke2`) but `preloadBootstrapImage` [2] looks for a tag with the arch suffix (e.g. `docker.io/rancher/rke2-runtime:v1.18.4-beta13-rke2-amd64`). Was able to get past this specific error by removing the suffix in `preloadBootstrapImage` but unsure if that's a long-term solution for multi-arch.

[1] https://github.com/rancher/rke2/blob/ca5f9bacfac555317852ce0f2b908791cf9c5978/scripts/build-images#L11-L23
[2] https://github.com/rancher/rke2/blob/ca5f9bacfac555317852ce0f2b908791cf9c5978/pkg/bootstrap/stage.go#L222-L234

joshrwolf:
Seeing the same error with the latest beta14 release on our fully airgapped mock C2S environment:

```
[root@ip-10-0-101-252 centos]# rke2 server
WARN[0000] not running in CIS 1.5 mode
FATA[0000] Get "https://index.docker.io/v2/": EOF
```

Steps to reproduce:

0. Be airgapped
1. rke2 beta14 binary in `/usr/bin/rke2` (from `yum install` method)
2. Released airgap images placed in `/var/lib/rancher/rke2/agent/images` directory
3. `rke2 server --debug` (or install script)

joshrwolf:
False alarm, `gunzip`ing the images and placing the `.tar` directly in `/var/lib/rancher/rke2/agent/images/` gets past the error.

[...]

PR #272, opened 2020-08-31
Remove GOARCH from preloaded bootstrap image tag

erikwilson:
#### Proposed Changes ####

Remove arch from bootstrap image airgap tag

#### Types of Changes ####

Bugfix

#### Verification ####

Needs airgap test

#### Linked Issues ####

#248

#### Further Comments ####

Credit for fix thanks to @nikkelma

### DIFF
--- a/pkg/bootstrap/stage.go
+++ b/pkg/bootstrap/stage.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 
 	errors2 "github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


brandond:
cc @cjellick - Just to be clear, this means that anyone wanting to support multi-arch with a private registry will need to jump through a number of extra hoops to push all the architectures into a manifest list. If you just push the airgap tarballs sequentially, they will overwrite each other. I don't think the tools to create manifests lists currently exist in the core docker tool - users will need the docker buildx plugin and/or a tool like skopeo to create or copy the manifest lists.

PR #273, opened 2020-08-31
selinux: specify the build-tag

dweomer:
- Compile with selinux
- Remove `Type=notify` from the systemd unit file
- Require root to execute install script

This plus https://github.com/rancher/rke2-selinux/pull/5 should address #156

### DIFF
--- a/bundle/share/rke2/rke2.service
+++ b/bundle/share/rke2/rke2.service
@@ -23,4 +23,3 @@ RestartSec=5s
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
 ExecStart=rke2 server
-Type=notify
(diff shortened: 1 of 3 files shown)


briandowns:
QA pushed to have the installer ran as a non root user and the installer invoke sudo where necessary. Are we changing that expectation and are they aware?

davidnuzik:
cc: @cjellick  ^^

I like that install.sh would automatically execute necessary commands as sudo, but at the same time this complicated the script - probably something we should avoid. In my opinion, we should no longer automatically sudo necessary commands, thus the user should be expected to run the install as sudo or the root user.

dweomer:
> cc: @cjellick ^^
> 
> I like that install.sh would automatically execute necessary commands as sudo, but at the same time this complicated the script - probably something we should avoid. **In my opinion, we should no longer automatically sudo necessary commands, thus the user should be expected to run the install as sudo or the root user.**

@davidnuzik @cjellick I agree with the bolded here.

Issue #269, opened 2020-08-31
Kubelet in NotReady state with beta13


ShylajaDevadiga:
Version:
```
rke2 -v
rke2 version v1.18.4-beta13+rke2
```
NodeOS and Arch:
Ubuntu 20.04, amd 64

Describe the issue:
Installation fails with node being in NotReady state.
Logs attached
Node info:
```
 Ready            False   Mon, 31 Aug 2020 16:05:42 +0000   Mon, 31 Aug 2020 16:05:40 +0000   KubeletNotReady              [runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized, CSINode is not yet initialized]
```
System log:
```
Aug 31 16:05:44 ip-172-31-32-217 systemd[1]: Started Rancher Kubernetes Engine v2.
Aug 31 16:05:44 ip-172-31-32-217 rke2[2680]: time="2020-08-31T16:05:44Z" level=info msg="Setting PSP: global-unrestricted-psp"
Aug 31 16:05:44 ip-172-31-32-217 rke2[2680]: time="2020-08-31T16:05:44Z" level=fatal msg="networkpolicies.networking.k8s.io \"default-network->
Aug 31 16:05:44 ip-172-31-32-217 systemd[1]: rke2.service: Main process exited, code=exited, status=1/FAILURE
Aug 31 16:05:44 ip-172-31-32-217 systemd[1]: rke2.service: Failed with result 'exit-code'.

```
Command used:
```
INSTALL_RKE2_VERSION=v1.18.4-beta13+rke2 INSTALL_RKE2_TYPE='server' INSTALL_RKE2_METHOD='installer' ./install.sh
```
[rke2.log](https://github.com/rancher/rke2/files/5151150/rke2.log)

cjellick:
@briandowns these are the three issues that @ShylajaDevadiga has associated with directory creation:
#244
#259
#269

Can you work on these as a group and ensure theyre all working?

briandowns:
On it. Thanks for aggregating the issues here.

briandowns:
Work performed in PR #274

briandowns:
Should be resolved with beta14 which I'll be cutting momentarily.

rancher-max:
Validated this is resolved in beta14

gellban:
**Is there a solution for rke2 version v1.21.5+rke2r2?**
Info:
Installed rke2 under proxy
$ rke2 -v
[...]

Issue #259, opened 2020-08-28
With no changes made on a running node, restarting rke2 service returns error


ShylajaDevadiga:
Version:
rke2 version v1.18.4-beta11+rke2

OS:
Ubuntu 20.04
Describe the issue:
Restarting rke2 service returns not to be in NotReady state.
```
sudo INSTALL_RKE2_VERSION=v1.18.4-beta11+rke2 INSTALL_RKE2_TYPE='server' INSTALL_RKE2_METHOD='installer' ./install.sh
kubectl get nodes
NAME              STATUS   ROLES         AGE   VERSION
ip-172-31-46-95   Ready    etcd,master   97s   v1.18.4-beta11+rke2

sudo cat /usr/local/lib/systemd/system/rke2.service
[Unit]
Description=Rancher Kubernetes Engine v2
Documentation=https://github.com/rancher/rke2#readme
Wants=network-online.target
After=network-online.target

[Install]
WantedBy=multi-user.target

[Service]
EnvironmentFile=-/etc/systemd/system/rke2.env
KillMode=process
Delegate=yes
# Having non-zero Limit*s causes performance problems due to accounting overhead
# in the kernel. We recommend using cgroups to do container-local accounting.
LimitNOFILE=1048576
LimitNPROC=infinity
LimitCORE=infinity
TasksMax=infinity
TimeoutStartSec=0
Restart=always
RestartSec=5s
ExecStartPre=-/sbin/modprobe br_netfilter
ExecStartPre=-/sbin/modprobe overlay
ExecStart=rke2 server
Type=notify

sudo systemctl restart rke2

Job for rke2.service failed because the control process exited with error code.
See "systemctl status rke2.service" and "journalctl -xe" for details.
```

Logs:
```
Aug 28 20:10:23 ip-172-31-46-95 systemd[1]: rke2.service: Found left-over process 6600 (nginx) in control group while starting unit. Ignoring.
Aug 28 20:10:23 ip-172-31-46-95 systemd[1]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
Aug 28 20:10:24 ip-172-31-46-95 rke2[7195]: time="2020-08-28T20:10:24Z" level=warning msg="not running in CIS 1.5 mode"
Aug 28 20:10:24 ip-172-31-46-95 rke2[7195]: time="2020-08-28T20:10:24Z" level=info msg="Extracting rancher/rke2-runtime:v1.18.4-beta11-rke2 bin/kubectl..."
[...]

PR #274, opened 2020-08-31
Update Directory Manifest Directory Exists Logic and Network Policy Existence Checks

briandowns:
#### Proposed Changes ####

Change processing logic for directory existence and processling logic for network policy processing.

#### Linked Issues ####

Addresses issues in #244, #259, and #269

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -169,7 +169,7 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 		return err
 	}
 
-	// we're ignoring and returned errors since the likelihood here is that
+	// we're ignoring any returned errors since the likelihood is that
 	// the error is that the new path already exists. That's indicative of a
 	// previously bootstrapped system. If it's a different error, it's indicative
 	// of an operating system or filesystem issue.
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -170,9 +170,7 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 		return err
 	}
 
-	if err := os.Rename(tempDir, dir); err != nil {
-		logrus.Warn(err.Error())
-	} else {
+	if err := os.Rename(tempDir, dir); err == nil {
 		return nil
 	}
 
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -169,6 +169,10 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
 		return err
 	}
 
+	// we're ignoring and returned errors since the likelihood here is that
+	// the error is that the new path already exists. That's indicative of a
+	// previously bootstrapped system. If it's a different error, it's indicative
+	// of an operating system or filesystem issue.
 	if err := os.Rename(tempDir, dir); err == nil {
 		return nil
 	}
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -12,16 +12,15 @@ import (
 	"runtime"
 	"strings"
 
-	errors2 "github.com/pkg/errors"
-	"github.com/rancher/wrangler/pkg/merr"
-
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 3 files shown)


scripts/validate
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -ex
 
-if [ -n "${SKIP_VALIDATE}" ]; then
+if [ -z "${SKIP_VALIDATE}" ]; then
```
dweomer, line 4: This will cause CI builds to no longer validate.

pkg/bootstrap/bootstrap.go
```diff
@@ -162,23 +161,27 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 	// extracting manifests
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	if err := os.Rename(tempDir, dir); err != nil && err != os.ErrExist {
-		return err
-	} else if err == nil {
+
+	if err := os.Rename(tempDir, dir); err != nil {
+		logrus.Warn(err.Error())
```
dweomer, line 174: this will give us an unnecessary warning log line every restart

scripts/validate
briandowns: -n doesn't work locally. Got another suggestion?

dweomer: Ah, I think -n treats "" as being "non-zero"

dweomer: So, remove the quotes around the var expansion and we should be good

pkg/rke2/np.go
```diff
@@ -57,14 +57,15 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
 	if _, ok := ns.Annotations[namespaceAnnotationNetworkPolicy]; !ok {
 		if _, err := cs.NetworkingV1().NetworkPolicies(namespace).Get(ctx, defaultNetworkPolicyName, metav1.GetOptions{}); err != nil {
 			if !apierrors.IsNotFound(err) {
+				if err := cs.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, defaultNetworkPolicyName, metav1.DeleteOptions{}); err != nil {
```
dweomer, line 60: The last two sentences on the description for setNetworkPolicy have me scratching my head:
> If the annoation doesnt' exist, we check to see if the policy exists. If it does, we delete it, and create the new default policy.

The intent is to always replace it when a server starts up? Seems like that will result in unnecessary churn in a multi-server setup.

As for the execution of that logic, this branch to delete/create will never be visited because the `cs.NetworkingV1().NetworkPolicies(namespace).Get()` call will not return an error if the policy exists.

pkg/bootstrap/bootstrap.go
```diff
@@ -161,23 +160,25 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 
 	// extracting manifests
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	if err := os.Rename(tempDir, dir); err != nil && err != os.ErrExist {
-		return err
-	} else if err == nil {
+
+	if err := os.Rename(tempDir, dir); err == nil {
```
cjellick, line 172: The old logic would return an error if it was not nil and wasnt an ErrExist. Just reasoning it out in my head, that makes sense. Can you explain how this change to remove that bit of error handling fixes the problem?

briandowns: The error returned is a LinkError and carries the underlying syscall error in it so it wouldn't actually match on ErrExists. This change simplifies for the expected scneario where the new path exists and continues silently. The other fail scneairos are os/fs related and would be evident from system state/stability.

pkg/rke2/np.go
dweomer: Marking as resolved: worked through this with @briandowns on a zoom.

pkg/bootstrap/bootstrap.go
```diff
@@ -161,23 +160,29 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
 
 	// extracting manifests
 	if err := extract(imgName, tempDir, prefix, r); err != nil {
 		return err
 	}
-	if err := os.Rename(tempDir, dir); err != nil && err != os.ErrExist {
-		return err
-	} else if err == nil {
+
+	// we're ignoring and returned errors since the likelihood here is that
```
cjellick, line 172: s/and/any

PR #287, opened 2020-09-02
selinux: custom context type for static pods

dweomer:
Set custom selinux context type to containers running in a static pod.
Refactor Vagrantfile to leverage installer and airgap images by default.
Depends on https://github.com/rancher/rke2-selinux/pull/6


ibuildthecloud (approved):

dweomer:
Addresses the last blocker for #4

Issue #271, opened 2020-08-31
Need ability to configure default private registry via command-line flag


brandond:
We need to give users a way to configure a default registry for core images and charts. rke2-charts support this via a value: `global.systemDefaultRegistry`. For system images (rke2 apiserver/controller-manager/etc) and bundled controllers (helm-controller) we will need to add the ability to pass a value that is prepended to the image name, same as how the helm charts handle it.

Design goals:
* Allow users to bootstrap a system from a local private registry **without providing any local content via airgap tarball**
* Prefix core images
* Pass through to `global.systemDefaultRegistry` for core charts
* Private registry must use TLS (no plaintext HTTP), with the cert trusted by the host OS CA bundle
* Private registry must not require authentication
* Prefix must be specified as host:port

Non-goals:
* Prefix any images, or configure helm charts, other than those required to support the core system and packaged components
* Configure the default registry as a docker.io registry mirror in containerd
* Advanced registry use cases (plaintext/auth/cert validation disabled)
* Tarball image preload is not supported when --system-default-registry is configured. 
  _k3s code will still load the images, but they won't be used because the images will be tagged as coming from the wrong registry._

Stretch goals:
* Provide a way for users to take advantage of the same behavior for their own manifests

rancher-max:
This is mostly working. The only image that isn't using the private registry is `rke2-ingress-nginx-controller`, which is therefore failing in an airgap setup. See below the comparison of the deployments from metrics-server and this image.
```
$ kubectl get -n kube-system deploy/rke2-metrics-server -o yaml | grep image
                f:image: {}
                f:imagePullPolicy: {}
        image: maxsprivateregistry.com/rancher/hardened-k8s-metrics-server:v0.3.6
        imagePullPolicy: IfNotPresent

$ kubectl get -n kube-system deploy/rke2-ingress-nginx-controller -o yaml | grep image
                f:image: {}
                f:imagePullPolicy: {}
        image: rancher/nginx-ingress-controller:nginx-0.30.0-rancher1
        imagePullPolicy: IfNotPresent
```

brandond:
IMO @rancher-max this should not be reopened, since the issue is with the chart, not rke2 core. I'm going to close this out as working, and you can test the chart fix in #369

PR #289, opened 2020-09-03
Add support for --system-default-registry CLI flag

brandond:
#### Proposed Changes ####

* Add support for --system-default-registry flag
* Remove legacy --repo flag
* Clean up, comment, and add logging to bootstrap code paths

#### Types of Changes ####

- New feature

#### Verification ####
See limitations for default registry in linked issue.
```
rke2 server --system-default-registry myregistry:443
```

#### Linked Issues ####

In service of #271

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -18,25 +19,42 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	errors2 "github.com/pkg/errors"
+	helmv1 "github.com/rancher/helm-controller/pkg/apis/helm.cattle.io/v1"
+	"github.com/rancher/helm-controller/pkg/helm"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/rancher/wrangler/pkg/merr"
+	"github.com/rancher/wrangler/pkg/schemes"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 var releasePattern = regexp.MustCompile("^v[0-9]")
 
-func dataDirFor(dataDir, dataName string) string {
-	return filepath.Join(dataDir, "data", dataName, "bin")
+const bufferSize = 4096
+
+// binDirForDigest returns the path to dataDir/data/refDigest/bin.
+func binDirForDigest(dataDir string, refDigest string) string {
+	return filepath.Join(dataDir, "data", refDigest, "bin")
 }
 
+// manifestsDir returns the path to dataDir/server/manifests.
 func manifestsDir(dataDir string) string {
 	return filepath.Join(dataDir, "server", "manifests")
 }
 
+// imagesDir returns the path to dataDir/agent/images.
+func imagesDir(dataDir string) string {
+	return filepath.Join(dataDir, "agent", "images")
+}
+
+// symlinkBinDir returns the path to dataDir/bin.
+// This will be symlinked to the current runtime bin dir.
 func symlinkBinDir(dataDir string) string {
 	return filepath.Join(dataDir, "bin")
 }
 
+// dirExists returns true if a directory exists at the given path.
 func dirExists(dir string) bool {
 	if s, err := os.Stat(dir); err == nil && s.IsDir() {
 		return true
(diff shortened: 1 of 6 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -135,22 +160,30 @@ func extract(image, targetDir, prefix string, reader io.Reader) error {
...
 		parts := strings.SplitN(str, ":", 2)
 		if len(parts) == 2 {
-			return parts[1]
+			return parts[1], nil
 		}
-		return parts[0]
+		return parts[0], nil
 	}
-	return ""
+	return "", fmt.Errorf("Bootstrap image %q is not a not a reference to a digest or version tag (%q)", ref, releasePattern)
```
brandond, line 179: It looks like the runtime image is always either rke2-runtime:v1.18-xxxx or rke2-runtime@sha256:yyyy, so we are safe considering anything else an error... now that we've fixed the type assertions.

pkg/bootstrap/bootstrap.go
```diff
@@ -25,76 +26,100 @@ import (
...
-		if dir := dataDirFor(dataDir, dataName); dirExists(dir) {
-			return dir, nil
-		}
+	// Try to use preload image from airgap tarball.
+	// Note that it is not considered an error for the image to be missing.
+	img, err := preloadBootstrapImage(dataDir, images.Runtime)
+	if err != nil {
+		return "", err
 	}
-	if dataName == "" {
```
brandond, line 71: No longer falling back to setting the data dir name from the image digest, since we've fixed the releaseRefDigest to properly generate unique names for the images without having to extract or pull them.

pkg/bootstrap/bootstrap.go
```diff
@@ -25,76 +26,100 @@ import (
...
+	// Skip all content extraction if the bin dir for this runtime image already exists
+	refBinDir := binDirForDigest(dataDir, refDigest)
+	if dirExists(refBinDir) {
+		logrus.Infof("Runtime image %q bin dir already exists at %q; skipping extract", ref, refBinDir)
+		return refBinDir, nil
 	}
 
-	dataName := releaseName(ref)
-	if dataName != "" {
-		if dir := dataDirFor(dataDir, dataName); dirExists(dir) {
```
brandond, line 67: This path was never taken. releaseName (now releaseRefDigest) always returned an empty string due to bad type assertions, so we never checked for existence and returned early.

cjellick:
the flag in rancher is `system-default-registry`. What do you think of making it match? I'm waffling on whether that matters or not.

brandond:
@cjellick I started with that, but it felt a little too long for a flag name so I dropped it down. Think we can just get away with documenting it? The SystemDefaultRegistry for Helm charts is inherited from the default-registry of the system (rke2)

brandond:
Should be feature-complete, with the caveat that system-default-registry is incompatible with the airgap images, since the new image names (prefixed with system-default-registry value) do not match the images in the tarball -which are tagged such that they appear to be from docker.io and k8s.gcr.io.

pkg/bootstrap/bootstrap.go
```diff
@@ -223,17 +279,124 @@ func preloadBootstrapImage(dataDir, runtimeImage string) (v1.Image, error) {
...
+
+	f, err := os.Open(fileName)
+	if err != nil {
+		return errors2.Wrapf(err, "Failed to open manifest %q for default-registry modification", fileName)
+	}
+
+	// Attempt to decode file as HelmChart; if it fails we assume the file contains some other
+	// sort of object and return.
+	// NOTE: This probably won't properly handle manifests containing multiple documents.
+	defer f.Close()
```
briandowns, line 356: Nit: move up to immediately under the if err check.

pkg/images/images.go
```diff
@@ -26,43 +31,49 @@ var (
...
+		KubeScheduler:       override(override(dockerRegistry, defaultRegistry)+"/rancher/kubernetes:"+KubernetesVersion, scheduler),
+		ETCD:                override(override(dockerRegistry, defaultRegistry)+"/rancher/etcd:"+EtcdVersion, etcd),
+		Pause:               override(override(k8sgcrRegistry, defaultRegistry)+"/pause:"+PauseVersion, pause),
 	}
 }
 
+// Pull checks for preloaded images in dir. If they are available, nothing is done.
+// If they are not available, it adds the image to name.txt in dir.
+// This is mostly used to track what images are being pulled for static pods.
 func Pull(dir, name, image string) error {
```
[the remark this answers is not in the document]
brandond: I found it, k3s starts a goroutine to read and preload these at the same time it imports the preload tarball.

pkg/bootstrap/bootstrap.go
```diff
@@ -187,54 +247,166 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
+func setSystemDefaultRegistry(manifestsDir string, systemDefaultRegistry string) error {
+	files := map[string]os.FileInfo{}
+	if err := filepath.Walk(manifestsDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		switch {
+		case info.IsDir():
+			return nil
+		case strings.HasSuffix(path, ".json"):
```
brandond, line 327: I copied this from the deploy controller, but the end result here is that we end up rewriting json files to yaml due to the serializer being hardcoded to output yaml. If this is a problem I can just drop support for json since we're only shipping yaml anyway.

pkg/bootstrap/bootstrap.go
```diff
@@ -18,83 +19,131 @@ import (
...
+	"github.com/rancher/wrangler/pkg/schemes"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-var releasePattern = regexp.MustCompile("^v[0-9]")
+var (
+	releasePattern = regexp.MustCompile("^v[0-9]")
+	serializer     = json.NewSerializerWithOptions(json.DefaultMetaFactory, schemes.All, schemes.All, json.SerializerOptions{Yaml: true, Pretty: true, Strict: true})
```
briandowns, line 34: How often is this called? Looks like it's only twice in one function so should be defined there so not to pollute package space. If this is a frequent call in the system and better to be defined here for resource purposes, that's fine.

pkg/podexecutor/staticpod.go
```diff
@@ -180,8 +188,9 @@ func (s *StaticPod) CurrentETCDOptions() (opts executor.InitialOptions, err erro
 	return
 }
 
-func (s *StaticPod) ETCD(args executor.ETCDConfig) error {
-	if err := images.Pull(s.PullImages, "etcd", s.Images.ETCD); err != nil {
+// ETCD starts the etcd static pod
```
briandowns, line 191:
```suggestion
// ETCD starts the etcd static pod.
```

pkg/bootstrap/bootstrap.go
brandond: It gets used to both decode and encode files from the manifests directory, so it is used more than once but only on startup. For GC purposes I could probably just create it in the calling function and then pass it in.

briandowns: Cool. That would be preferred.

scripts/build-images
```diff
@@ -16,9 +16,9 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images.txt
     docker.io/rancher/k8s-metrics-server:v0.3.6
     docker.io/rancher/klipper-helm:v0.3.0
     docker.io/rancher/kube-proxy:${KUBERNETES_VERSION}
-    k8s.gcr.io/defaultbackend-amd64:1.5
-    k8s.gcr.io/pause:3.2
-    quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.30.0
+    docker.io/rancher/pause:3.2
+    docker.io/rancher/nginx-ingress-controller-defaultbackend:1.5-rancher1
+    docker.io/rancher/nginx-ingress-controller:nginx-0.30.0-rancher1
```
dweomer, line 21: It seems like we've tried to setup a pattern for the image name + tag but applied it inconsistently.

brandond: Yeah, this is an existing image that's used by RKE1. See the rant here: https://github.com/rancher/image-mirror/pull/26#discussion_r479576501

briandowns (approved):
LGTM

cjellick:
I like "In service of" in the description. sounds so official :-D

pkg/bootstrap/bootstrap.go
```diff
@@ -187,54 +244,167 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
+		chart.Spec.Set = map[string]intstr.IntOrString{}
+	}
+	if chart.Spec.Set["global.systemDefaultRegistry"].StrVal != systemDefaultRegistry {
+		chart.Spec.Set["global.systemDefaultRegistry"] = intstr.FromString(systemDefaultRegistry)
+		chartChanged = true
+	}
+
+	jobImage := helm.DefaultJobImage
+	if systemDefaultRegistry != "" {
+		jobImage = systemDefaultRegistry + "/" + helm.DefaultJobImage
```
dweomer, line 384: @brandond won't this get us the weird scenario you were describing yesterday i.e. `mycustomregistry.com/docker.io/rancher/klipper-helm`

dweomer: Nevermind, DefaultJobImage doesn't have the repo prefix

pkg/bootstrap/bootstrap.go
```diff
@@ -187,54 +244,167 @@ func extractFromDir(dir, prefix string, img v1.Image, imgName string) error {
...
+	}
+
+	if chart.Spec.JobImage != jobImage {
+		chart.Spec.JobImage = jobImage
```
dweomer, line 388: This assumes the chart-defined `JobImage` is the default and clobbers anything the user chart may have specified.

pkg/bootstrap/bootstrap.go
brandond: Yeah, I left it unqualified. If we ever change that we'll need to handle it.

pkg/bootstrap/bootstrap.go
brandond: Correct. This is necessary to ensure that the current flag value is applied to the files on disk so that they can get picked up by the deploy controller.

Do you think we should actually rewrite the image string to apply the registry setting, instead of just replacing it? Or only change it if it's empty or ends with the default value? 

cc: @cjellick

dweomer (approved):

brandond:
Confirmed that config file works:
```
root@centos01 ~]# cat /etc/rancher/rke2/config.yaml
---
system-default-registry: reg.lan.khaus:443

[root@centos01 ~]# rke2 --debug server
WARN[0000] not running in CIS 1.5 mode
INFO[0000] Runtime image tarball preload disabled by --system-default-registry=reg.lan.khaus:443
INFO[0000] Pulling runtime image "reg.lan.khaus:443/rancher/rke2-runtime:v1.18.4-dev-97f4e876"
2020-09-10 11:36:01.364272 I | Unable to read "/root/.docker/config.json": open /root/.docker/config.json: no such file or directory
2020-09-10 11:36:01.364325 I | No matching credentials were found, falling back on anonymous
FATA[0000] Failed to pull runtime image "reg.lan.khaus:443/rancher/rke2-runtime:v1.18.4-dev-97f4e876": Get "https://reg.lan.khaus:443/v2/": dial tcp: lookup reg.lan.khaus on 10.0.1.1:53: no such host
```

Issue #297, opened 2020-09-09
Update To Only Apply Network Policy if in CIS Mode


briandowns:
Currently, the code applies a network policy regardless of whether the user has specified CIS mode or not. This needs to be changed to only apply the policy if in CIS mode. 

The policy should also allow for DNS to be passed to kube-system namespace.

briandowns:
Work performed in #299

rancher-max:
#### Validated using `v1.18.8-beta18+rke2`

There have been some changes that make the previous noted here: https://github.com/rancher/rke2/issues/208#issuecomment-685151866 a little bit different. See below:
 - There is one additional network policy (dns-policy):
```
$ k get networkpolicy -A
NAMESPACE     NAME                         POD-SELECTOR       AGE
default       default-network-policy       <none>             3h24m
kube-public   default-network-policy       <none>             3h24m
kube-system   default-network-dns-policy   k8s-app=kube-dns   3h24m
kube-system   default-network-policy       <none>             3h24m
```
- `kube-system` namespace has one additional annotation: `np.rke2.io/dns: resolved`
- Due to psps in CIS mode, need to pay attention to the images in manifests being applied. For reference, the error received when applying a simple nginx deployment is: `Error: container has runAsNonRoot and image will run as root` and setting securitycontext on this will not fix it (although there can be custom nginx images that a securitycontext would resolve. This should be expected with the "more-secure-by-default" nature of CIS mode with the PSPs and networkpolicies).

- I applied the following to the default namespace:
```
kind: Service
apiVersion: v1
metadata:
  name: test
spec:
  type: LoadBalancer
  selector:
    app: test
  ports:
    - name: http
      protocol: TCP
      port: 8080
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deploy
  labels:
    app: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: test
        image: nginxinc/nginx-unprivileged:1.15.12
        ports:
        - containerPort: 8080
```
Then I ran `kubectl run curl-default --image=radial/busyboxplus:curl -i --tty --rm` and `curl`ed the cluster-ip of that service. It returned the expected nginx result since those pods were running in the same namespace. I then ran the same command but in the `test` namespace and tried to curl the service running in the default namespace. This failed as expected as there should not be traffic allowed betwen namespaces with these network policies.

[...]

PR #299, opened 2020-09-09
Apply Network Policies Only in CIS Mode and Add Policy For DNS

briandowns:
#### Proposed Changes ####

* Adds a check to see if running in CIS mode and if so, apply the network policies.
* Adds a new network policy for DNS traffic.

#### Linked Issues ####

#297

### DIFF
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -133,8 +133,9 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
 	return nil
 }
 
-// setDNSPolicy
-func setDNSPolicy(ctx context.Context, cs *kubernetes.Clientset) error {
+// setNetworkDNSPolicy sets the default DNS policy allowing the
+// required DNS traffic.
+func setNetworkDNSPolicy(ctx context.Context, cs *kubernetes.Clientset) error {
 	ns, err := cs.CoreV1().Namespaces().Get(ctx, metav1.NamespaceSystem, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/rke2/np.go
```diff
@@ -85,29 +133,78 @@ func setNetworkPolicy(ctx context.Context, namespace string, cs *kubernetes.Clie
...
+		ns.Annotations[namespaceAnnotationNetworkDNSPolicy] = cisAnnotationValue
+
+		if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			if _, err := cs.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
+				if apierrors.IsConflict(err) {
+					if err := updateNamespaceRef(ctx, cs, ns); err != nil {
+						return err
+					}
+				}
+				return err
```
galal-hussein, line 168: in case of conflict this will still return error, should this be return updateNamespaceRef(...

Oats87:
I think we should document that if you have these network policies set, ingress to namespaces other than kube-system will be broken.

brandond:
Does allowing DNS ingress to coredns cover everything, or do other pods also need egress rules allowing this as well?

briandowns:
> Does allowing DNS ingress to coredns cover everything, or do other pods also need egress rules allowing this as well?

I believe this is all that's needed.

brandond (approved):

galal-hussein (approved):
LGTM

Issue #295, opened 2020-09-09
RPM-based install results in a different systemd service name than rke2-installer.run


bgulla:
os: CentOS Linux release 7.8.2003 (Core)
kernel:  3.10.0-1127.el7.x86_64
version: rke2-beta16

It would seem that the systemd service that is installed differs based on the installation method. RPM-based installs registers as `rke2-server` where the self-installer registers a `rke2` service. 
 

`rke2-installer.run` -> `/usr/local/lib/systemd/system/rke2.service`
`rke2-server.rpm` -> `(/usr/lib/systemd/system/rke2-server.service`

dweomer:
This was intentional on my part. My reasoning:
- Naming the systemd service by the role it plays reeks. When on a host I need to know what role it plays in the larger cluster before I can inspect the service that defines the hosts participation in said cluster? No, this is a flaw with the RPM installer(s) in my opinion.

Would like a design discussion with @Oats87 about this.

dweomer:
> we did not want to run into conflicting unit files when you try to install both RPM's, which was a desired capability.

@Oats87 I understand everything up until this statement. Under what scenario would one want two services installed on a host that leverage content on the filesystem that is largely overlapping and cannot be run at the same time?

cjellick:
We should discuss this today and come to a conclusion.

davidnuzik:
RPM to have 2 different unit files for each server and agent. We will no longer capture env vars - user will need to use config file. Will need to add conflicts for each server and agent (if server running then cannot also run agent service, if agent service running then cannot also run server).

cjellick:
@dweomer This is at least partially addressed, as per your references in the above prs. Is it fully addressed and can be moved to test?
cc @davidnuzik - please follow up on this today

ShylajaDevadiga:
Upgrade using SUC fails due to the location rke2 is running.
Upgrade from rke2 -v
rke2 version v1.18.8-beta19+rke2 to beta20
```
system-upgrade   apply-rke2-server-on-ip-172-31-15-79.us-east-2.compute.in-9p4ws      0/1     Error       0          2m44s
```
Logs:
```
kubectl logs -n system-upgrade apply-rke2-server-on-ip-172-31-15-79.us-east-2.compute.in-9p4ws
+ upgrade
+ get_rke2_process_info
+ grep -E '/usr/local/bin/rke2 .*(server|agent)'
+ ps -ef
+ grep -E -v '(init|grep)'
[...]

PR #323, opened 2020-09-15
Update k3s; use new k3s no_stage build tag

brandond:
#### Proposed Changes ####

Add no_stage build tag to disable k3s packaged components.

#### Types of Changes ####

* build

#### Verification ####

* Start RKE2
* Note lack of /var/lib/rancher/rk2/server/static directory and /var/lib/rancher/rke2/manifests/rolebindings.yaml

#### Linked Issues ####

Related to #171 and #232

#### Further Comments ####

Next step is to stop hiding the --disable flag and make it available for rke2 use.

PR #325, opened 2020-09-15
bundle: fix killall script

dweomer:
Fix the killall script to properly stop rke2 systemd service(s).

Partially addresses #295


dweomer:
Had to rebase and tidy to fix `go.sum` which was broken via #323

brandond (approved):

Issue #324, opened 2020-09-15
k3s ClusterRoleBinding missing from RKE2 RBAC


brandond:
rancher/rke2#323 removed the packaged components from k3s, which includes the k3s rolebindings.yaml. Two of the three roles in there are k3s specific (clusterrole.rbac.authorization.k8s.io/system:k3s-controller and clusterrolebinding.rbac.authorization.k8s.io/system:k3s-controller) but the third (clusterrolebinding.rbac.authorization.k8s.io/kube-apiserver-kubelet-admin) is needed to allow the kubelet to function properly. Without it the tunnel endpoint controller does not work and we get errors like:

```
ERRO[0278] Unable to watch for tunnel endpoints: unknown (get endpoints)
```

rancher-max:
@brandond I still occasionally see the mentioned error as a warning:
```
INFO[0059] etcd data store connection OK                
INFO[0059] Waiting for API server to become available   
WARN[0063] Unable to watch for tunnel endpoints: unknown (get endpoints) 
INFO[0064] labels have been set successfully on node: ip-172-31-12-53 
```
It only happens once and always in this same spot, and doesn't cause any errors. Is this expected? Potentially a timing issue with the fix here?

brandond:
You will see `WARN[0063] Unable to watch for tunnel endpoints:` with various reasons after it repeated until after etcd is up, apiserver is up, and rbac is loaded. From the other lines in your log it looks like the apiserver has not yet started.

We probably need to figure out how to get this to wait until later to start so that it's not spamming errors all the time.

rancher-max:
In that case this issue itself is fixed. There are no blocking errors, the kubelet appears to be functioning properly. Down the road we can figure out how to reduce the warning spam. As a note, I do not see the warnings when performing actions on the cluster and checking kubelet logs.
Also, ensured the clusterrolebinding is present: `clusterrolebinding.rbac.authorization.k8s.io/kube-apiserver-kubelet-admin                           ClusterRole/system:kubelet-api-admin`

Validated using `v1.18.9-beta21+rke2`

PR #329, opened 2020-09-16
Add missing ClusterRoleBindings

brandond:
#### Proposed Changes ####

Several critical ClusterRoleBindings were being inherited from k3s's rolebindings.yaml. Now that we have purged all k3s content, we need to provide them locally.

The `system:k3s-controller` identifier should be changed to `system:rke2-controller` when we have pulled through the change from https://github.com/rancher/k3s/pull/2258 so that we're not adding more k3s-branded resources.

#### Types of Changes ####

* RBAC
* Startup hooks

#### Verification ####

* Start RKE2. System will function normally. Without this PR, master builds will give repeated messages as described in #324 and multi-node clusters will not work due lack of agent tunnel.

The name targeted here (

#### Linked Issues ####

#324

#### Further Comments ####

This is heavily inspired by @briandowns work on PSP deployment. I cleaned up a few things in psp.go while I was reading through to figure out how it works.

### DIFF
--- a/pkg/rke2/clusterrole_templates.go
+++ b/pkg/rke2/clusterrole_templates.go
@@ -2,7 +2,7 @@ package rke2
 
 const (
 	kubeletAPIServerRoleBindingName = "kube-apiserver-kubelet-admin"
-	tunnelControllerRoleName        = "system:k3s-controller"
+	tunnelControllerRoleName        = "system:rke2-controller"
 )
 
 const kubeletAPIServerRoleBindingTemplate = `apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/rke2/clusterrole_templates.go
+++ b/pkg/rke2/clusterrole_templates.go
@@ -1,0 +1,70 @@
+package rke2
+
+const (
+	kubeletAPIServerRoleBindingName = "kube-apiserver-kubelet-admin"
+	tunnelControllerRoleName        = "system:k3s-controller"
+)
+
+const kubeletAPIServerRoleBindingTemplate = `apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: %s
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:kubelet-api-admin
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: kube-apiserver
+`
+
+const tunnelControllerRoleTemplate = `apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: %s
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - "networking.k8s.io"
+  resources:
+  - networkpolicies
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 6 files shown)


pkg/rke2/psp.go
```diff
@@ -324,36 +324,34 @@ func setPSPs() func(context.Context, <-chan struct{}, string) error {
...
 					} else {
-						logrus.Fatalf("psp: get clusterrole binding: %s", err.Error())
+						logrus.Fatalf("psp: get clusterrolebinding: %s", err.Error())
 					}
 				}
 			}
 
 			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-				if _, err := cs.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
-					if apierrors.IsConflict(err) {
```
brandond, line 337: The initial update try and check for IsConflict are unnecessary as per [RetryOnConflict docs](https://godoc.org/k8s.io/client-go/util/retry#RetryOnConflict). Docs say you should always retrieve before changing/updating, and the retry handler will figure out on its own if the error is a conflict or not.

pkg/rke2/clusterrole_templates.go
```diff
@@ -0,0 +1,70 @@
+package rke2
```
cjellick, line 1: When @briandowns did the PSP work, as a non-essential review item I asked him to consider using structs instead of const strings. He did that for networkPolicy: https://github.com/rancher/rke2/blob/master/pkg/rke2/np.go#L28

Now having seen now seen both, I'm not sure which is right/better. The string consts are much more readable than the structs, but then instead of dealing with the "real" objects and clients in your code, you're dealing with yaml strings.

design tradeoffs amirght?

At any rate, either is fine for now and I would not want you to spend the time refactoring.

briandowns: Having done both now, the strings definitely yield better readability. Since this code is run only at start up, I'd rather have readability over one-off efficiency or proficiency.

pkg/rke2/clusterrole_templates.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	kubeletAPIServerRoleBindingName = "kube-apiserver-kubelet-admin"
+	tunnelControllerRoleName        = "system:k3s-controller"
```
dweomer, line 5: thought this was going to be changing?

pkg/rke2/clusterrole_templates.go
dweomer: I prefer structs with helper functions but that does make for a lot more scaffolding code. Maybe unstructured types hit the sweet spot? https://godoc.org/github.com/kubernetes/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured

brandond: There are actually a bunch of helpers for creating RBAC entries with a builder pattern, as opposed to using raw strings or structs. We should probably be doing that instead. The end result less immediately visible than a nice chunk of YAML, but I think overall it's cleaner.

https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L105

pkg/rke2/clusterrole_templates.go
brandond: Yes, I noted that in the PR description. If the k3s PR gets accepted and pulled through to rke2 then I'll update it before merging, otherwise it will need to be changed after that's done.

brandond: I've added a k3s update to this PR to address this.

dweomer (approved):

briandowns (approved):

Issue #331, opened 2020-09-16
Update Kubernetes Build Process to Build a Static Binary


briandowns:
It's been found that the kubernetes image build process isn't building a static binary.

PR #332, opened 2020-09-16
adds the flag necessary to create a statically linked binary

briandowns:
#### Proposed Changes ####

Adds the necessary arguments to have Kubernetes build a static binary.

#### Linked Issues ####

rancher/rke2#331

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ ARG TAG
 WORKDIR /
 RUN git clone -b ${KUBERNETES_VERSION} --depth=1 https://github.com/kubernetes/kubernetes.git
 WORKDIR /kubernetes
-RUN make KUBE_GIT_VERSION=${TAG} WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -51,7 +51,7 @@ ARG TAG
 WORKDIR /
 RUN git clone -b ${KUBERNETES_VERSION} --depth=1 https://github.com/kubernetes/kubernetes.git
 WORKDIR /kubernetes
-RUN make KUBE_GIT_VERSION=${TAG} WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
+RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
```
brandond, line 54: We have some extra logic around setting cflags in build-binary - do we need to add that here, or are we OK just adding the static ldflag for now?

brandond (approved):

Dockerfile
briandowns: I think this is fine for now until it becomes necessary to pass more flags for the linker.

dweomer (approved):

Issue #333, opened 2020-09-17
Static build of kubelet crashes with signal SIGSEGV: segmentation violation


brandond:
The change to static builds in rancher/rke2#332 results in kubelet segfaulting. Unfortunately because we don't capture kubelet stderr, all that rke2 reports is `ERRO[1199] Kubelet exited: exit status 2`. Rebuilding rke2 with stderr capture enabled shows a nice segfault:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x63 pc=0x7ff71c04a5ed]

runtime stack:
runtime.throw(0x43bae6b, 0x2a)
	/usr/local/go/src/runtime/panic.go:1116 +0x72
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:679 +0x46a

goroutine 306 [syscall]:
runtime.cgocall(0x329a4c0, 0xc0005b25c0, 0xc000b7cad0)
	/usr/local/go/src/runtime/cgocall.go:133 +0x5b fp=0xc0005b2590 sp=0xc0005b2558 pc=0x40638b
net._C2func_getaddrinfo(0xc000e800e0, 0x0, 0xc0008d35f0, 0xc000b7cad0, 0x0, 0x0, 0x0)
	_cgo_gotypes.go:92 +0x55 fp=0xc0005b25c0 sp=0xc0005b2590 pc=0x566145
net.cgoLookupIPCNAME.func1(0xc000e800e0, 0x13, 0x13, 0xc0008d35f0, 0xc000b7cad0, 0x2, 0xc0005b26a0, 0x569522)
	/usr/local/go/src/net/cgo_unix.go:161 +0xd2 fp=0xc0005b2608 sp=0xc0005b25c0 pc=0x56bba2
net.cgoLookupIPCNAME(0x4315f41, 0x2, 0x7ffc347ed536, 0x12, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/usr/local/go/src/net/cgo_unix.go:161 +0x183 fp=0xc0005b2718 sp=0xc0005b2608 pc=0x5675d3
net.cgoIPLookup(0xc0007fcae0, 0x4315f41, 0x2, 0x7ffc347ed536, 0x12)
	/usr/local/go/src/net/cgo_unix.go:218 +0x67 fp=0xc0005b27b8 sp=0xc0005b2718 pc=0x567d07
runtime.goexit()
[...]

PR #334, opened 2020-09-17
Build with -tags=netgo,osusergo to prevent crashes when statically linked

brandond:
#### Proposed Changes ####

Build Kubernetes components with -tags=netgo,osusergo to prevent segfaulting. We're already doing this for k3s so the same should be safe here.

Adding the osuersgo tag caused the go linker to crash with SIGBUS when building, which led me to discover that we have been building with golang 1.14. We should be building with 1.13, as upstream skipped 1.14 entirely. Moving back to 1.13.15 allows the build to succeed, and the resulting binaries run without crashing. There are some linker warnings about missing symbols that need to be resolved, but I will open another issue for that.

#### Types of Changes ####

* Build flags
* k3s update

#### Verification ####

* Start rke2
* Note that kubelet no longer crash-loops at 5 second intervals and the server comes up as expected.

#### Linked Issues ####

Related to #333

#### Further Comments ####

Also update k3s to master, as the last update apparently didn't get us to the rev we needed to be at to make the agent tunnel controller work with the RBAC changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG KUBERNETES_VERSION=dev
 # Build environment
-FROM rancher/hardened-build-base:v1.14.2-amd64 AS build
+FROM rancher/hardened-build-base:v1.13.15-amd64 AS build
 RUN set -x \
     && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y update \
@@ -51,7 +51,7 @@ ARG TAG
 WORKDIR /
 RUN git clone -b ${KUBERNETES_VERSION} --depth=1 https://github.com/kubernetes/kubernetes.git
 WORKDIR /kubernetes
-RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
+RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' GOFLAGS='-tags=netgo,osusergo' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
 
 FROM registry.access.redhat.com/ubi7/ubi-minimal:latest AS kubernetes
 RUN microdnf update -y           && \
--- a/go.mod
+++ b/go.mod
@@ -59,7 +59,7 @@ require (
 	github.com/google/go-containerregistry v0.0.0-20190617215043-876b8855d23c
 	github.com/pkg/errors v0.9.1
 	github.com/rancher/helm-controller v0.7.3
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -51,7 +51,7 @@ ARG TAG
 WORKDIR /
 RUN git clone -b ${KUBERNETES_VERSION} --depth=1 https://github.com/kubernetes/kubernetes.git
 WORKDIR /kubernetes
-RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
+RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' GOFLAGS='-tags=netgo' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
```
briandowns, line 54: should have a ref to home dir too

brandond: added osusergo

dweomer (approved):

Dockerfile
```diff
@@ -51,7 +51,7 @@ ARG TAG
 WORKDIR /
 RUN git clone -b ${KUBERNETES_VERSION} --depth=1 https://github.com/kubernetes/kubernetes.git
 WORKDIR /kubernetes
-RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
+RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' GOFLAGS='-tags="netgo osusergo"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
```
dweomer, line 54:
```suggestion
RUN make KUBE_GIT_VERSION=${TAG} GOLDFLAGS='-extldflags "-static"' GOFLAGS='-tags "netgo osusergo"' WHAT='cmd/kube-apiserver cmd/kube-controller-manager cmd/kube-proxy cmd/kube-scheduler cmd/kubeadm cmd/kubectl cmd/kubelet vendor/k8s.io/apiextensions-apiserver'
```

dweomer (changes requested):
build is failing. left suggestion that may fix

brandond:
Any attempt at quoting or whitespace fails with something like `go: parsing $GOFLAGS: non-flag "\"netgo` regardless of what I use. Seems like something in the makefile or build script is breaking it. I did find a workaround that lets us fail a little later...

brandond:
This is blocked on https://github.com/rancher/image-build-base/pull/2 - we should be building Kubernetes 1.18 with golang 1.13, not 1.14... so we need to tag a 1.13 release in image-build-base before we can use it here for the image builds.

briandowns (approved):

dweomer (approved):

Issue #320, opened 2020-09-15
Restarting after quorum loss does not successfully reset the cluster


rancher-max:
**Description:**
After setting up a multi-server node cluster, and then stopping the majority of the server nodes so that it loses quorum, it then fails to come back up successfully to a usable state. NOTE this only affects rke2, not k3s. The same steps in k3s result in the expected behavior. Also NOTE this applies whether using the binary or install scripts.

**Steps to reproduce:**
1. Create a 3 server-node cluster. I used the binary directly to do this:
```
wget https://github.com/rancher/rke2/releases/download/v1.18.8-beta19+rke2/rke2.linux-amd64 && \
sudo mv rke2.linux-amd64 /usr/local/bin/rke2 && \
sudo chmod u+x /usr/local/bin/rke2 
# Node 1:
sudo rke2 server --write-kubeconfig-mode 644 --cluster-init --etcd-snapshot-retention 7  --etcd-snapshot-schedule-cron="*/5 * * * *" 2>&1 &
# Other nodes:
sudo rke2 server --token="<token value>" --server="https://<node 1 public ip>:9345" 2>&1 &
```
2. Deploy a few workloads in the cluster. Wait until a snapshot is taken. Deploy a few more workloads.
3. Shutdown nodes 2 and 3. 
    - The cluster should now be unreachable as quorum has been lost.
4. Restart node 1 using: `sudo rke2 server --write-kubeconfig-mode 644 --cluster-init --etcd-snapshot-retention 7  --etcd-snapshot-schedule-cron="*/5 * * * *" --cluster-reset --cluster-reset-restore-path="/var/lib/rancher/rke2/server/db/snapshots/<val with the first few workloads>" 2>&1 &`
    - I saw errors for a while: `ERRO[0020] This server is a not a member of the etcd cluster. Found [ip-172-31-23-63-dde66403=https://172.31.23.63:2380 ip-172-31-18-246-0602b4e3=https://172.31.18.246:2380 ip-172-31-28-148-24429a45=https://172.31.28.148:2380], expect: ip-172-31-18-246-0250c7cc=172.31.18.246 `
    - It eventually resolved itself, ending with: `INFO[0032] etcd is running, restart without --cluster-reset flag now. Backup and delete ${datadir}/server/db on each peer etcd server and rejoin the nodes`. However, I believe this part is already incorrect as instead it should tell me something about "restart without restore-path" 
[...]

PR #338, opened 2020-09-19
Add etcd name file to hash files

galal-hussein:
#### Proposed Changes ####

use etcd name file in list of hash files that passed as env variable to api, scheduler, and controller manager

#### Types of Changes ####

bug fix

#### Verification ####

same verification herer **https://github.com/rancher/rke2/issues/320**

#### Linked Issues ####

https://github.com/rancher/rke2/issues/320

depends on https://github.com/rancher/k3s/pull/2284

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -39,6 +39,7 @@ type StaticPodConfig struct {
 	Images        images.Images
 	CloudProvider *CloudProviderConfig
 	CISMode       bool
+	DataDir       string
 }
 
 type CloudProviderConfig struct {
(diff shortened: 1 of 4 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -101,14 +102,17 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
...
-
 	after(etcdReady, func() error {
+		etcdNameFile := filepath.Join(s.DataDir, "server", "db", "etcd", "name")
 		return staticpod.Run(s.ManifestsDir, staticpod.Args{
 			Command:   "kube-apiserver",
 			Args:      args,
 			Image:     s.Images.KubeAPIServer,
 			Dirs:      ssldirs,
 			CPUMillis: 250,
+			Files: []string{
```
brandond, line 113: This is a single line in other places, should it be the same here?

pkg/podexecutor/staticpod.go
```diff
@@ -101,14 +102,17 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
 			args = append(args[:i], args[i+1:]...)
 		}
 	}
-
 	after(etcdReady, func() error {
+		etcdNameFile := filepath.Join(s.DataDir, "server", "db", "etcd", "name")
```
brandond, line 106: Maybe make this a local func since we do the same thing with the same magic list of strings so many places.

cjellick (approved):

Issue #177, opened 2020-08-12
Add conformance tests to CI


davidnuzik:
We need to add conformance tests to RKE2 ci - similar to what k3s is doing today.

davidnuzik:
I've edited this issue. This issue is needed for GA - it now just covers adding conformance tests. I've separated out the CIS Scan aspect into this issue here which can come post-GA: https://github.com/rancher/rke2/issues/243

cjellick:
if all the conformance tests dont currently pass, i would settle now for just ensuring that a cluster comes up successfully and the k8s api becomes available

PR #340, opened 2020-09-21
Add initial testing

erikwilson:
#### Proposed Changes ####
Add conformance testing for rke2

#### Types of Changes ####
ci testing scripts

#### Verification ####
n/a

#### Linked Issues ####
#177


Dockerfile
```diff
@@ -103,3 +114,44 @@ COPY --from=charts \
...
+    bash \
+    conntrack \
+    ebtables \
+    ethtool \
+    iptables \
+    jq \
+    less \
+    libseccomp2 \
+    socat \
+    vim
```
dweomer, line 155: Since #352 has been merged the base is now derived from Alpine therefore this will need to be updated to use apk.

scripts/package-images
```diff
@@ -8,4 +8,4 @@ source ./scripts/version.sh
 mkdir -p dist/artifacts
 
 cp -f build/images.txt dist/artifacts/${PROG}-images.${PLATFORM}.txt
-gzip < build/images/${PROG}-airgap.tar > dist/artifacts/${PROG}-images.${PLATFORM}.tar.gz
+gzip < build/${PROG}-airgap.tar > dist/artifacts/${PROG}-images.${PLATFORM}.tar.gz
```
dweomer, line 11: `build/images` is mounted into dev-shell at `/var/lib/rancher/rke2/agent/images`. This change will have the side-effect of forcing rke2 to reach out to internet registries for all images other than `rancher/rke2-runtime` and `rancher/hardened-kubernetes`. Is that intentional?

dweomer:
@erikwilson this looks great!

scripts/package-images
erikwilson: The problem was that I couldn't get the air-gap tar file into the test image, presumably because `build/images` is in `.dockerignore`. Am thinking either we should remove that entry from `.dockerignore` or change dev-shell to use `build`.

Dockerfile
```diff
@@ -13,11 +13,27 @@ RUN set -x \
 # Dapper/Drone/CI environment
 FROM build AS dapper
 ENV DAPPER_ENV GODEBUG REPO TAG DRONE_TAG PAT_USERNAME PAT_TOKEN KUBERNETES_VERSION DOCKER_BUILDKIT
+ARG DAPPER_HOST_ARCH
+ENV ARCH $DAPPER_HOST_ARCH
+ENV DRONE_BUILD_EVENT IMAGE_NAME GCLOUD_AUTH
```
dweomer, line 18: @erikwilson is is possible that the gcloud bit is not working because of the GCLOUD_AUTH not being in the DAPPER_ENV line above?

dweomer (approved):
LGTM (and left a sugestion for GCLOUD_AUTH)

Dockerfile
erikwilson: Thanks! Looks like that fixed it! :)

galal-hussein (approved):
LGTM

Issue #335, opened 2020-09-17
[process][feature] assert control plane components are statically built artifacts


dweomer:
Review the control-plane components and apply build-time assertions:
- `ldd` gets you "not a dynamic executable"
- `go tool nm` gets you `*_Cfunc__goboringcrypto_*` symbols where appropriate, prior to stripping. see https://kupczynski.info/2019/12/15/fips-golang.html#build-process

If the component is built from a separate repository the assertion(s) should be made there at build-time:
- [x] etcd :arrow_right: https://github.com/rancher/image-build-etcd/pull/2
  - etcdctl
- [x] metrics-server :arrow_right: https://github.com/rancher/image-build-k8s-metrics-server/pull/1
- [x] coredns :arrow_right: https://github.com/rancher/image-build-coredns/pull/2
- [x] kube-proxy :arrow_right: https://github.com/rancher/image-build-kube-proxy/pull/4
- [x] calico :arrow_right: https://github.com/rancher/image-build-calico/pull/3
- [x] flannel :arrow_right: https://github.com/rancher/image-build-flannel/pull/4
- [x] crictl :arrow_right: https://github.com/rancher/image-build-crictl/pull/1
- [x] containerd :arrow_right: https://github.com/rancher/image-build-containerd/pull/4
  - ctr
  - containerd
  - containerd-shim
  - containerd-shim-runc-v1
  - containerd-shim-runc-v2
- [x] runc :arrow_right: https://github.com/rancher/image-build-runc/pull/1
- [x] kubernetes :arrow_right: https://github.com/rancher/rke2/pull/352
  - kubelet
  - kubectl
  - kube-apiserver
  - kube-scheduler
  - kube-controller-manager
- [x] rke2 :arrow_right: https://github.com/rancher/rke2/pull/352

davidnuzik:
QA Smoke test, ensure we didn't break anything obvious. Can do some automation checks as you see fit.

rancher-max:
Validated using `v1.18.9-beta21+rke2` not seeing any regressions through smoke tests and automation.

PR #352, opened 2020-09-23
statically linked kubernetes

dweomer:
Leverage new alpine-based rancher/hardened-build-base (goboring built on
alpine) while matching the upstream version of go to compile with. Also
assert everything is statically linked and assert presence of goboring
symbols prior to install-with-strip.

Addresses rancher/rke2#335

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,38 +1,14 @@
 ARG KUBERNETES_VERSION=dev
 # Build environment
-FROM rancher/hardened-build-base:v1.13.15-amd64 AS build
+FROM rancher/hardened-build-base:v1.13.15b4 AS build
 RUN set -x \
-    && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get -y update \
-    && apt-get -y install \
+ && apk --no-cache add \
+    bash \
+    curl \
+    file \
+    git \
     libseccomp-dev \
     rsync
-# Shell used for debugging
-FROM build AS shell
-RUN set -x \
-    && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get -y update \
-    && apt-get -y install \
-    bash \
-    bash-completion \
-    git \
-    iptables \
-    less \
-    libseccomp2 \
-    psmisc \
-    rsync \
-    seccomp \
-    socat \
-    sudo \
-    vim
-RUN GO111MODULE=off GOBIN=/usr/local/bin go get github.com/go-delve/delve/cmd/dlv
-RUN cp -f /etc/skel/.bashrc /etc/skel/.profile /root/ && \
-    echo 'alias abort="echo -e '\''q\ny\n'\'' | dlv connect :2345"' >> /root/.bashrc
-ENV PATH=/var/lib/rancher/rke2/bin:$PATH
-ENV KUBECONFIG=/etc/rancher/rke2/rke2.yaml
-VOLUME /var/lib/rancher/rke2
-# This makes it so we can run and debug k3s too
-VOLUME /var/lib/rancher/k3s
 
 # Dapper/Drone/CI environment
 FROM build AS dapper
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -51,14 +46,60 @@ ARG TAG
...
+RUN sh -ex ./go-build-static -o bin/kube-apiserver          ./cmd/kube-apiserver
+RUN sh -ex ./go-build-static -o bin/apiextensions-apiserver ./vendor/k8s.io/apiextensions-apiserver
+RUN sh -ex ./go-build-static -o bin/kube-controller-manager ./cmd/kube-controller-manager
+RUN sh -ex ./go-build-static -o bin/kube-proxy              ./cmd/kube-proxy
+RUN sh -ex ./go-build-static -o bin/kube-scheduler          ./cmd/kube-scheduler
+RUN sh -ex ./go-build-static -o bin/kubeadm                 ./cmd/kubeadm
+RUN sh -ex ./go-build-static -o bin/kubectl                 ./cmd/kubectl
+RUN sh -ex ./go-build-static -o bin/kubelet                 ./cmd/kubelet
+# assert statically linked executables
+RUN echo '[ -e $1 ] && (file $1 | grep -E "executable, x86-64, .*, statically linked")' \
```
brandond, line 79: Does this catch everything? I had `file` report statically linked binaries while `nm` still showed some undefined and weak symbols.

scripts/build-binary
```diff
@@ -25,13 +25,21 @@ VERSION_FLAGS="
...
         -X ${K3S_PKG}/pkg/version.GitCommit=${REVISION}
 "
+
+#STATIC_FLAGS='-extldflags "-static -Wl,--fatal-warnings"'
+### ON GLIBC SYSTEMS YOU WILL SEE LINKER WARNINGS #####################################################################
+# github.com/rancher/rke2
+#/usr/bin/ld: /tmp/go-link-743463534/000027.o: in function `nvmlInit_dl':
+#/home/jacob/go/pkg/mod/github.com/mindprince/gonvml@v0.0.0-20190828220739-9ebdce4bb989/bindings.go:152: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
+### THESE CAN BE SAFELY IGNORED UNLESS YOU ARE TRYING TO LEVERAGE NVIDIA METRICS FROM CADVISOR AND YOU HAPPEN TO BE ###
+### RUNNING RKE2 ON A SYSTEM OTHER THAN THE ONE WHICH IS WAS BUILT THAT ALSO HAS AN INCOMPATIBLE GLIBC ################
```
brandond, line 35: Do we want to patch the nvml stuff out of our cadvisor fork?

brandond (approved):

brandond:
I have a few questions but LGTM!

cjellick (approved):

scripts/build-binary
dweomer: We do not see this warning when building in CI, aka using the alpine-based goboring and I imagine some folks will want nvidia metrics at some point and I would prefer to see reports of panics over reports that "I cannot get nvidia metrics out of my installation". If we want to patch the cadvisor stuff I am fine with that but it should be a separate issue.

brandond: It looks like the cadvisor nvml stuff ends up calling `dlopen("libnvidia-ml.so.1", RTLD_LAZY);` at https://github.com/mindprince/gonvml/blob/master/bindings.go#L152.

With musl libc, dlopen is a stub call for static binaries:
* https://github.com/alandekok/jlibtool/issues/3#issue-16398345
* https://git.musl-libc.org/cgit/musl/tree/src/ldso/dlopen.c

Dockerfile
```diff
@@ -1,64 +1,106 @@
...
+RUN echo 'GO_BUILD_FLAGS=" \
+         -mod=vendor \
+         -gcflags=-trimpath=/kubernetes \
+         "' \
+    >> ./go-build-static
+RUN echo 'GO_LDFLAGS=" \
+         -X k8s.io/component-base/version.gitVersion=${TAG} \
+         -X k8s.io/component-base/version.gitCommit=$(git rev-parse HEAD) \
+         -X k8s.io/component-base/version.gitTreeState=clean \
+         -X k8s.io/component-base/version.buildDate=$(date -u +%Y-%m-%dT%H:%M:%SZ) \
```
erikwilson, line 59: nit: would setup a variable for the date since the two build dates might be off by a second or so. could do the same for git rev.

Dockerfile
```diff
@@ -1,64 +1,106 @@
...
+         -X k8s.io/client-go/pkg/version.gitVersion=${TAG} \
+         -X k8s.io/client-go/pkg/version.gitCommit=$(git rev-parse HEAD) \
+         -X k8s.io/client-go/pkg/version.gitTreeState=clean \
+         -X k8s.io/client-go/pkg/version.buildDate=$(date -u +%Y-%m-%dT%H:%M:%SZ) \
+         -linkmode=external -extldflags \"-static -Wl,--fatal-warnings\""' \
+    >> ./go-build-static
+RUN echo 'go build ${GO_BUILD_FLAGS} -ldflags "${GO_LDFLAGS}" "${@}"' \
+    >> ./go-build-static
+# force code generation
+RUN make WHAT=cmd/kube-apiserver
```
erikwilson, line 69: nit: might be nice to have a code-gen target. does this build apiserver twice?

dweomer: It does but the second run is faster. I tried disentangling the code generation and it was too thorny for my tired brain.

Issue #232, opened 2020-08-26
Need ability to disable packaged components (similar to k3s --disable)


brandond:
rke2 does not currently allow disabling any of the packaged components (exclusively HelmChart manifests, at the moment). The k3s `--disable` flag is currently hidden and uses a hardcoded value, since we're overloading it to disable the k3s packaged components.

We should:
1. Put the k3s manifests and static content inside a build flag so that we can disable them when compiling rke2
1. Stop using the `--disable` flag as an internal flag to k3s, and expose it to the CLI as a string slice (same as how it behaves in k3s).  We should not validate or otherwise restrict the values, as we should be able to add (and support disabling) bootstrap image content without modifying a static value list.
1. Check for matching file basenames [when extracting content from the bootstrap image](https://github.com/rancher/rke2/blob/master/pkg/bootstrap/stage.go#L87), and (if matched), skip extracting the file.
1. To make this identical to the k3s `--disable` flag we would also need to actively remove any matching files from the manifests directory, if found.

cjellick:
I'm not going to close this, but I think we dont need this for GA because we have the .skip behavior, ie:
```
mkdir -p /var/lib/rancher/rke2/server/manifests/charts/; touch /var/lib/rancher/rke2/server/manifests/charts/rke2-canal-chart.yml.skip
```

That just needs clearly documented.

cjellick:
@davidnuzik to open a documentation issue

davidnuzik:
Created https://github.com/rancher/rke2/issues/257

cjellick:
Moving into GA since @brandond has a fix in

brandond:
The k3s PR takes care of 1. on the list; the remainder should be fairly easy to do as well since I've been poking about in the bootstrap code recently.

ShylajaDevadiga:
Validated using rke2 version v1.18.9-beta21+rke2

Updated config.yaml wtih disable: rke2-ingress-nginx

rke2-ingress-ngimx chart is not deployed
```
helm ls -n kube-system
NAME               	NAMESPACE  	REVISION	UPDATED                                	STATUS  	CHART                       	APP VERSION
rke2-canal         	kube-system	1       	2020-09-29 22:50:24.194218136 +0000 UTC	deployed	rke2-canal-v3.13.3          	v3.13.3    
rke2-coredns       	kube-system	1       	2020-09-29 22:50:24.090076152 +0000 UTC	deployed	rke2-coredns-1.10.101       	1.6.9      
[...]

PR #361, opened 2020-09-24
Add support for disabling packaged components (charts)

brandond:
#### Proposed Changes ####

Allow disabling packaged components with `--disable` flag, similar to k3s. Update help to list components that can be disabled.

#### Types of Changes ####

* CLI flags

#### Verification ####

* Stark rke2 with `rke2 server --help | grep disable`
* Note list of disableable components is shown
* Start rke2 with `rke2 server --disable rke2-ingress-nginx`
* Note that ingress chart is not deployed

#### Linked Issues ####

Related to #232

#### Further Comments ####

This change removes the `-chart` suffix from the packaged charts as a convenience to the user. I do not see anywhere that we are relying upon this suffix.

I have hardcoded the list of disableable components, as we do in k3s at https://github.com/rancher/k3s/blob/master/pkg/cli/cmds/server.go#L11

Also fixed a previously unnoticed issue with the --data-dir flag (#367) and several flag types that we had not been handling (was preventing --disable from being overridable), and added an error to catch future unhandled types.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,11 +94,11 @@ ARG CHARTS_REPO="https://rke2-charts.rancher.io"
 ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
 RUN echo ${CACHEBUST}>/dev/null
-RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN CHART_VERSION="v1.18.9"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal.yaml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
+RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns.yaml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
+RUN CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx.yaml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN CHART_VERSION="v1.18.9"     CHART_FILE=/charts/rke2-kube-proxy.yaml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
+RUN CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server.yaml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
 RUN rm -vf /charts/*.sh /charts/*.md
 
 # rke-runtime image
--- a/pkg/cli/cmds/k3sopts.go
+++ b/pkg/cli/cmds/k3sopts.go
@@ -67,7 +67,42 @@ func commandFromK3S(cmd cli.Command, flagOpts map[string]*K3SFlagOption) (cli.Co
 				strFlag.Hidden = true
 			}
 			flag = strFlag
+		} else if strFlag, ok := flag.(*cli.StringFlag); ok {
+			if opt.Usage != "" {
+				strFlag.Usage = opt.Usage
+			}
+			if opt.Default != "" {
+				strFlag.Value = opt.Default
+			}
+			if opt.Hide {
+				strFlag.Hidden = true
+			}
+			flag = strFlag
+		} else if strSliceFlag, ok := flag.(cli.StringSliceFlag); ok {
+			if opt.Usage != "" {
+				strSliceFlag.Usage = opt.Usage
+			}
+			if opt.Default != "" {
+				slice := &cli.StringSlice{}
+				parts := strings.Split(opt.Default, ",")
+				for _, val := range parts {
+					slice.Set(val)
+				}
+				strSliceFlag.Value = slice
+			}
+			if opt.Hide {
+				strSliceFlag.Hidden = true
+			}
+			flag = strSliceFlag
 		} else if intFlag, ok := flag.(cli.IntFlag); ok {
+			if opt.Usage != "" {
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -66,11 +66,11 @@ ARG CHARTS_REPO="https://rke2-charts.rancher.io"
 ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
 RUN echo ${CACHEBUST}>/dev/null
-RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal-chart.yml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.10.101"    CHART_FILE=/charts/rke2-coredns-chart.yml           CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="1.36.300"    CHART_FILE=/charts/rke2-ingress-nginx-chart.yml     CHART_BOOTSTRAP=false   /charts/build-chart.sh
-RUN CHART_VERSION="v1.18.8"     CHART_FILE=/charts/rke2-kube-proxy-chart.yml        CHART_BOOTSTRAP=true    /charts/build-chart.sh
-RUN CHART_VERSION="2.11.100"    CHART_FILE=/charts/rke2-metrics-server-chart.yml    CHART_BOOTSTRAP=false   /charts/build-chart.sh
+RUN CHART_VERSION="v3.13.3"     CHART_FILE=/charts/rke2-canal.yaml             CHART_BOOTSTRAP=true    /charts/build-chart.sh
```
dweomer, line 69: We've lopped off the `-chart` infix to make the `--disable` logic work? There was a reason the infix made sense to me but I can't remember it now. Hopefully it isn't load-bearing.

pkg/cli/cmds/server.go
```diff
@@ -69,7 +70,7 @@ var (
 		"with-node-id":                drop,
 		"node-label":                  copy,
 		"node-taint":                  copy,
-		"docker":                      copy,
+		"docker":                      drop,
```
dweomer, line 73: This seems to be addressing an unrelated issue. If this does need to be "fixed" I'd prefer that to be in a separate pull-request. Or, at the very least, link to the issue this addresses in this pull-request and call it out in the commit comment.

pkg/cli/cmds/k3sopts.go
```diff
@@ -60,13 +61,40 @@ func commandFromK3S(cmd cli.Command, flagOpts map[string]*K3SFlagOption) (cli.Co
...
+			if opt.Usage != "" {
+				strFlag.Usage = opt.Usage
+			}
+			if opt.Default != "" || opt.OverrideDefault {
+				strFlag.Value = opt.Default
+			}
+			if opt.Hide {
+				strFlag.Hidden = true
+			}
+			flag = strFlag
```
dweomer, line 81: Why the duplicative cases for `*cli.StringFlag`, `*cli.IntFlag`, and `*cli.BoolFlag`?

dweomer (commented):
Please see comments

pkg/cli/cmds/k3sopts.go
brandond: Because we have flags of these types, and type assertions for the non-pointer types don't work on pointers. They were being silently skipped.

pkg/cli/cmds/server.go
brandond: I don't believe rke2 supports docker as a backend, does it? I'll create an issue if there isn't one.

Dockerfile
brandond: I couldn't find anything that depends on it. Finding the list of things that can be disabled is already difficult enough, given how we package and distribute the charts. I was hoping to just point users at the rke2-charts site, as described in the PR notes, but the chart names there don't match because we're appending -chart to the names here.

brandond: Another option would be to template the names in at build time, but even that is a bit of a hack as we don't actually have a fixed list of items - it's just whatever is in the charts subdir of the runtime image.

pkg/cli/cmds/k3sopts.go
dweomer: We shouldn't be mixing types. Is this left over from the spur/cli reversion to urfave/cli?

brandond: I don't know why they vary, I just figured it was something that got broken when the cli was reverted. I can dig into it, but if it is a problem it'll need fixing on the k3s side.

Dockerfile
dweomer: I think your first instinct for a "document fix" is probably good enough. If it turns out not to be we can generate a chart inventory at build-time similar to how we do for images.

pkg/cli/cmds/server.go
brandond: Opened #365, will remove this from here.

dweomer (approved):

pkg/cli/cmds/k3sopts.go
brandond: Opened an issue to follow up in k3s: https://github.com/rancher/k3s/issues/2308

pkg/rke2/rke2.go
```diff
@@ -57,23 +54,22 @@ func Agent(clx *cli.Context, cfg Config) error {
...
 			}
 		}
 	}
 
 	for _, f := range clx.App.Flags {
 		switch t := f.(type) {
 		case cli.StringFlag:
-			if t.Name == "profile" && t.Destination != nil && *t.Destination != "" {
-				cisMode = true
+			if strings.Contains(t.Name, "profile") {
```
brandond, line 70: Changed to be consistent with clx.Command.Flags processing above.

pkg/cli/cmds/k3sopts.go
```diff
@@ -19,10 +19,11 @@ var (
...
 type K3SFlagOption struct {
-	Hide    bool
-	Drop    bool
-	Usage   string
-	Default string
+	Hide            bool
+	Drop            bool
+	Usage           string
+	Default         string
+	OverrideDefault bool
```
brandond, line 26: Needed to allow setting the default to an empty string. Could also just remove the empty string check, but I assumed it was there for a reason?

pkg/rke2/rke2.go
```diff
@@ -57,23 +54,22 @@ func Agent(clx *cli.Context, cfg Config) error {
 
 func setup(clx *cli.Context, cfg Config) error {
 	var dataDir string
+
 	for _, f := range clx.Command.Flags {
 		switch t := f.(type) {
 		case cli.StringFlag:
 			if strings.Contains(t.Name, "data-dir") {
-				dataDir = t.Value
+				dataDir = *t.Destination
```
brandond, line 62: Value holds the default value, the actual user-provided input is stored in Destination.

dweomer (approved):

pkg/rke2/rke2.go
erikwilson: Would it be better to make sure it is equal?

brandond: The way @briandowns set it up, the value is validated over at https://github.com/rancher/rke2/blob/master/pkg/cli/cmds/root.go#L148 so by the point we get here, if it's not an empty string then we want to turn on cisMode.

erikwilson (approved):

Issue #374, opened 2020-09-26
Setting profile=cis-1.5 does not trigger expected etcd changes


cjellick:
If running in etcd mode, the following are expected to happen:
- the etcd process is ran as the etcd user & group. The process running as etcd:etcd should be triggered by a securityContext section in the etcd static pod yaml (/var/lib/rancher/rke2/agent/pod-manifests/etcd.yaml) where uid and gid are specified.
- the etcd data dir is owned by etcd:etcd

This isnt happening.

version: beta20
on an ubuntu 20.04 box

Steps to repro:
1. Create an rke2 cluster (single node)
2. create the etcd user (see readme on how to do that)
2. Once its up, turn on cis profile:
```
cat /usr/local/lib/systemd/system/rke2-server.env
RKE2_CIS_PROFILE="cis-1.5"
```

cjellick:
found during this time: https://github.com/rancher/rke2/issues/373

cjellick:
This was blocking some of my CIS assertions and documentation, so I had to investigae. I believe I have a fix.

cjellick:
Fix is merged, but im holding onto the issue until i can do a beta and test it out.

cjellick:
Turning on CIS via the profile does a number of things, like turning on PSPs, network policy, and affecting how etcd is ran. I was testing etcd related stuff when I ran into this and didn't have time to fully investigate if PSP and net policy were affected, but for etcd, here are the things you should observe:
The etcd data dir is owned by the etcd user and group:
```
stat -c %U:%G /var/lib/rancher/rke2/server/db/etcd
etcd:etcd
```

The etcd process is ran as the etcd user:
```
etcd     2494105 2494029  3 Sep27 ?        00:35:15 etcd --config-file=/var/lib/rancher/rke2/server/db/etcd/config
```
and relatedly, the etcd pod manifest has a security context specified which points at the etcd user:
```
 grep -i securitycontext -A3  /var/lib/rancher/rke2/agent/pod-manifests/etcd.yaml
  securityContext:
    runAsGroup: 996
    runAsUser: 996
```
that 996 corresponds to the etcd user I had made earlier:
```
grep etcd /etc/passwd
etcd:x:996:996::/var/lib/rancher/rke2/server/db:/bin/false
```

rancher-max:
#### Validated on beta23

- All of the above is working correctly and is the same behavior I see
- Validated etcd pod is running as etcd user (in my case 997):
```
[...]

PR #375, opened 2020-09-27
Ensure profile flag is respected

cjellick:
#### Proposed Changes ####
Problem: the --profile was not being respecting, causing things like the
etcd pod not being ran as the etcd user. This is because we were
shadowing a variable declation.

Solution: Removing the shadowed declaration of the cisMode variable. The
global variable of the same name that is properly assigned a value
earlier on in the logic will now be used.

Fixing this also exposed that a CIS related flag,
`--protect-kernel-defaults` was improperly set.

Solution: Removing the shadowed declaration of the cisMode variable. The
global variable of the same name that is properly assigned a value
earlier on in the logic will now be used.

To fix protect-kernel-defaults, remove superfluous `--`.

#### Types of Changes ####
Bug fix

#### Verification ####
Set the profile flag to cis-1.5. Right now you have to do that via the
system unit env file because the flag is not respected in the config
file (separate bug for that). Here's how to set it in the env file:
```
vim /usr/local/lib/systemd/system/rke2-server.env
RKE2_CIS_PROFILE="cis-1.5"
```

#### Linked Issues ####
#374

#### Further Comments ####
I needed to fix this to move forward with CIS documentation

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -91,7 +91,6 @@ func setup(clx *cli.Context, cfg Config) error {
 
 	agentManifestsDir := filepath.Join(dataDir, "agent", config.DefaultPodManifestPath)
 	agentImagesDir := filepath.Join(dataDir, "agent", "images")
-	cisMode := clx.String("profile") != ""
 
 	managed.RegisterDriver(&etcd.ETCD{})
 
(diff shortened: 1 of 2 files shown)


pkg/rke2/rke2.go
```diff
@@ -91,7 +91,6 @@ func setup(clx *cli.Context, cfg Config) error {
 
 	agentManifestsDir := filepath.Join(dataDir, "agent", config.DefaultPodManifestPath)
 	agentImagesDir := filepath.Join(dataDir, "agent", "images")
-	cisMode := clx.String("profile") != ""
```
briandowns, line 94: Ugh. I removed this line in a previous PR since it's being set above. Not sure how that got back.

briandowns (approved):

cjellick:
merging with just one review 😬

PR #378, opened 2020-09-28
Add support for rke2 public channelserver in install script

MonzElmasry:
Problem:
Install.sh uses github releases, instead it should use rke2 public channelserver https://update.rke2.io

Solution:
Adding support by using rke2 public channelserver, Used by adding
INSTALL_RKE2_CHANNEL= `stable` || `latest` || `testing` , default is "stable"

Signed-off-by: MonzElmasry <menna.elmasry@rancher.com>
issue : https://github.com/rancher/rke2/issues/242

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -132,8 +132,9 @@ get_release_version() {
         version=${INSTALL_RKE2_VERSION}
     else
         info "finding release for channel ${INSTALL_RKE2_CHANNEL}"
-        # version_url="${INSTALL_RKE2_CHANNEL_URL}/${INSTALL_RKE2_CHANNEL}"
-        version_url="${INSTALL_RKE2_GITHUB_URL}/releases/latest"
+        INSTALL_RKE2_CHANNEL_URL=${INSTALL_RKE2_CHANNEL_URL:-'https://update.rke2.io/v1-release/channels'}
+        INSTALL_RKE2_CHANNEL=${INSTALL_RKE2_CHANNEL:-'stable'}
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -132,8 +132,9 @@ get_release_version() {
         version=${INSTALL_RKE2_VERSION}
     else
         info "finding release for channel ${INSTALL_RKE2_CHANNEL}"
-        # version_url="${INSTALL_RKE2_CHANNEL_URL}/${INSTALL_RKE2_CHANNEL}"
-        version_url="${INSTALL_RKE2_GITHUB_URL}/releases/latest"
+        INSTALL_RKE2_CHANNEL_URL=${INSTALL_RKE2_CHANNEL_URL:-'https://update.rke2.io/v1-release/channels'}
+        INSTALL_RKE2_CHANNEL=${INSTALL_RKE2_CHANNEL:-'stable'}
+        version_url="${INSTALL_RKE2_CHANNEL_URL}/${INSTALL_RKE2_CHANNEL}"
```
dweomer, line 137: This looks fine but does it leave `INSTALL_RKE2_GITHUB_URL` cruft? If so we should remove that.

dweomer: Nevermind, I see that `$INSTALL_RKE2_GITHUB_URL` is used for the downloads.

dweomer (approved):

briandowns (approved):
LGTM

brandond (approved):

install.sh
MonzElmasry: it's used for actually downloading tarball and checksum after detecting the desired version from channelserver

Issue #376, opened 2020-09-27
fix coredns config map warning


cjellick:
make this not happen:
```
Sep 27 22:01:11 craig-rke2 rke2[2493703]: time="2020-09-27T22:01:11Z" level=warning msg="Unable to fetch coredns config map: configmap \"coredns\" not found"
```

brandond:
@cjellick I think this is a regression caused by #361 that should be fixed before GA. I failed to consider that --disable still turns off controllers associated with a few of the manifests, even if those manifests are disabled at build time. That change means that the coredns and servicelb controllers are now active on rke2, when they should not be.

ShylajaDevadiga:
Validated in v1.18.9-beta22+rke2, coredns config map warning is not seen in logs.

PR #381, opened 2020-09-28
Restore disabling built-in k3s components

brandond:
#### Proposed Changes ####

This is a partial revert of #631 - we still need to --disable the k3s built-in components to disable their controllers.

#### Types of Changes ####

* CLI
* Packaged components

#### Verification ####

* Start RKE2
* Note lack of warnings described in linked issue

#### Linked Issues ####

Related to #376

#### Further Comments ####

This isn't a straight revert of the change since the previous behavior was only kind of accidentally working - the StringSlice ended up looking like the following:

```go
[]string{"rke2-coredns", "coredns, servicelb, traefik, local-storage, metrics-server"}
```

This is fine because k3s accesses the flag as a string, and urfave hides the odd intermediate form by just joining all the args with a comma... but if k3s ever tries to use this as an actual string slice it will not work correctly.

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -32,8 +32,18 @@ func Server(clx *cli.Context, cfg Config) error {
 	if err := setup(clx, cfg); err != nil {
 		return err
 	}
+
 	if err := clx.Set("secrets-encryption", "true"); err != nil {
 		return err
+	}
+
+	// Disable all disableable k3s packaged components. In addition to manifests,
+	// this also disables several integrated controllers.
+	disableItems := strings.Split(cmds.DisableItems, ",")
+	for _, item := range disableItems {
+		if err := clx.Set("disable", strings.TrimSpace(item)); err != nil {
+			return err
+		}
 	}
 
 	cmds.ServerConfig.StartupHooks = append(cmds.ServerConfig.StartupHooks,


pkg/rke2/rke2.go
```diff
@@ -32,10 +32,20 @@ func Server(clx *cli.Context, cfg Config) error {
 	if err := setup(clx, cfg); err != nil {
 		return err
 	}
+
 	if err := clx.Set("secrets-encryption", "true"); err != nil {
 		return err
 	}
 
+	// Disable all disableable k3s packaged components. In addition to manifests,
+	// this also disables several integrated controllers.
```
dweomer, line 41: Which controllers?

brandond: ServiceLB controller is disabled here:
https://github.com/rancher/k3s/blob/master/pkg/cli/server/server.go#L194
Node controller is disabled here:
https://github.com/rancher/k3s/blob/master/pkg/server/server.go#L165

The Cloud Controller frustratingly does the opposite - it has a dedicated `--disable-cloud-controller` flag which adds a hidden entry to the --disable list to skip deploying the RBAC manifest.

dweomer (approved):

galal-hussein (approved):
LGTM

Issue #351, opened 2020-09-23
Need a way to enable CIS mode when installing rke2 via install script


rancher-max:
**Description:**
Currently, CIS mode is enabled from a "Global Option" level in the rke2 binary via the `--profile` flag. This is not valid for the config.yaml, so there is no easy way to enable CIS mode when installing rke2 via the install script. The only known workaround would be to either run the binary instead, or directly edit `/usr/local/lib/systemd/system/rke2-server.env`

We should either allow global flags, such as `debug` and `profile` to be used in the config.yaml, or make rke2 aware of the `RKE2_CIS_PROFILE` environment variable. Note if we do not do this with the environment variable, then we should remove mention of this environment variable in the help text as well.

cjellick:
@brandond can you please figure out why the profile flag (and any other non-functional global rke2 flags) are not working?

davidnuzik:
_Some more information for context:_
The `--profile` flag doesn't do any mitigation, it only checks that the kernel params are set. No "flag" exists to enable "CIS mode". Previously we would perform CIS mitigation via the `INSTALL_RKE2_CIS_MODE=true` install environment variable. This no longer works after the refactor (it was taken out to simplify things).

So, in order for us to do CIS mitigation we must have a new flag or some script to perform this. **Or**, if this is no longer our goal then we need clarification and confirmation that the user will be expected to mitigate as per our documentation.

FYI here was the original issue where we tested cis mitigation and `--profile` flag: https://github.com/rancher/rke2/issues/61

davidnuzik:
@cjellick QA approached me to indicate concern on this ^^ Can we get confirmation that it was intentional to remove the capability to enable "CIS mode". This "mode" was how we programmatically did some CIS mitigation for the user. This is no longer possible.

Would like confirmation that the user must follow docs and do their own mitigation thus QA too would be expected to do this before passing the `--profile=cis-1.5` flag and value.

Also confused on purpose of https://github.com/rancher/rke2/blob/master/bundle/share/rke2/rke2-cis-sysctl.conf

cc: @dweomer

cjellick:
Just to be clear: this bug is for the fact that you cannot set the profile flag in the config.yaml. It has nothing to do with the functionality of profile once it is turned on.

The profile flag NEVER did mitigation and it still doesn't. The user must follow docs for mitigation, but the docs are currently incomplete. I am working on fixing the CIS docs up now.

brandond:
[...]

PR #387, opened 2020-09-29
Move profile flag to common flags

brandond:
#### Proposed Changes ####

Move profile flag to common flags for server/agent so that it can be loaded from config YAML. Also simplified flag use a bit to get rid of unnecessary iteration over flag lists.

#### Types of Changes ####

* CLI

#### Verification ####

* Start rke2 with `profile: cis-1.5`in config.yaml
* Note that user and sysctls are validated; no "Not running in CIS mode" warning shown.

#### Linked Issues ####

Related to #351

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -54,8 +54,16 @@ func NewAgentCommand() cli.Command {
 }
 
 func AgentRun(clx *cli.Context) error {
-	if profile == "" {
+	switch profile {
+	case "cis-1.5":
+		if err := validateCISreqs(); err != nil {
+			logrus.Fatal(err)
+		}
+	case "":
 		logrus.Warn("not running in CIS 1.5 mode")
+	default:
+		logrus.Fatal("invalid value provided for --profile flag")
 	}
+
 	return rke2.Agent(clx, config)
 }
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -60,18 +60,10 @@ func setup(clx *cli.Context, cfg Config) error {
 		case cli.StringFlag:
 			if strings.Contains(t.Name, "data-dir") {
 				dataDir = *t.Destination
-			}
-		}
-	}
-
-	for _, f := range clx.App.Flags {
-		switch t := f.(type) {
-		case cli.StringFlag:
-			if t.Name == "profile" && t.Destination != nil && *t.Destination != "" {
+			} else if t.Name == "profile" && t.Destination != nil && *t.Destination != "" {
 				cisMode = true
 			}
-		default:
-			// nothing to do. Keep moving.
+
 		}
 	}
 
--- a/pkg/cli/cmds/root.go
+++ b/pkg/cli/cmds/root.go
@@ -37,6 +37,12 @@ var (
 			EnvVar:      "RKE2_CLOUD_PROVIDER_CONFIG",
 			Destination: &config.CloudProviderConfig,
 		},
+		&cli.StringFlag{
+			Name:        "profile",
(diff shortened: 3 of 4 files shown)


pkg/rke2/rke2.go
```diff
@@ -53,27 +52,8 @@ func Agent(clx *cli.Context, cfg Config) error {
...
-		case cli.StringFlag:
-			if t.Name == "profile" && t.Destination != nil && *t.Destination != "" {
-				cisMode = true
-			}
-		default:
-			// nothing to do. Keep moving.
-		}
-	}
+	cisMode = clx.String("profile") != ""
+	dataDir := clx.String("data-dir")
```
cjellick, line 56: I have to ask why? 
Why did we iterate through the flags before and why do we not need to now.

pkg/cli/cmds/agent.go
```diff
@@ -54,8 +53,6 @@ func NewAgentCommand() cli.Command {
 }
 
 func AgentRun(clx *cli.Context) error {
-	if profile == "" {
-		logrus.Warn("not running in CIS 1.5 mode")
-	}
+	validateCISreqs()
```
cjellick, line 56: Teasing out the logic of the validateCISreqs, I now wonder if we will refuse to start a worker node because the etcd user doesn't exist even though that user need not exist on a worker node.

cjellick: I think that bug may already exist, but this change here made me think about it.

cjellick: The answer is: yes it already exists.

pkg/rke2/rke2.go
brandond: I suspect this was either necessary when using a different CLI library, or sometime didn't know the right way to do it.

pkg/cli/cmds/agent.go
```diff
@@ -54,8 +54,16 @@ func NewAgentCommand() cli.Command {
 }
 
 func AgentRun(clx *cli.Context) error {
-	if profile == "" {
+	switch profile {
+	case "cis-1.5":
+		if err := validateCISreqs(); err != nil {
```
cjellick, line 59: why did you need to move the validateCISreqs() logic?

Part of the reason I ask is that I think you are oging to have to change it to fix the etcd user bug i mentioned

cjellick (approved):

pkg/cli/cmds/agent.go
cjellick: answered offline and the TL;DR is that this needed moved once the arg was moved out to the server and agent commands

pkg/cli/cmds/agent.go
```diff
@@ -54,8 +54,16 @@ func NewAgentCommand() cli.Command {
 }
 
 func AgentRun(clx *cli.Context) error {
-	if profile == "" {
+	switch profile {
```
briandowns, line 57: This would probably make sense to have in a function in root.go that takes the profile string as an arg and does this logic so it's not needed to be maintained in 2 different places. More easily extensible too if/when we move to 1.6, etc. Not blocking however it'd be cool to have an issue that captures this.

briandowns (approved):

pkg/cli/cmds/agent.go
brandond: I agree 100%. I had originally moved this into the validateCISreqs function, but Craig asked me to reduce the amount of change so I went for a more minimal approach at the cost of some repetition.

cjellick: +1. He basically had that, but I asked him to minimize the pr to the smallest change possible for the purpose of reviewing and merging as quickly as possible.

briandowns: Cool. Do we have an issue for this for after GA?

Issue #327, opened 2020-09-15
CIS Documentation


briandowns:
* Hardening Guide
* Self Assessment

davidnuzik:
Let's use the original issue https://github.com/rancher/rke2/issues/84
Craig and my bad on that - it wasn't assigned to you earlier.

Issue #389, opened 2020-09-29
When running in CIS mode, we need to change what we do around protect-kernel-defaults


cjellick:
When running with the CIS profile turned on, we have logic that ensures the following kernel parameters are set as defined:
```
vm.panic_on_oom=0
vm.overcommit_memory=1
kernel.keys.root_maxbytes=25000000
kernel.panic=10
kernel.panic_on_oops=1
```
rke2 will not modify these, it will just error out if their values aren't set as expected.

This was done to conform with CIS control 4.2.6, but that is not what the control is looking for. The control just requires that that the kubelet flag `protect-kernel-defaults` is set to true. Setting this flag to true will cause the kubelet to error out if the parameters are not set to the above values (see this pr where the feature was introduced: https://github.com/kubernetes/kubernetes/pull/27874/files)

Also, the rke2 code is checking 5 kernel parameters, but only four are required by the kubelet (see: https://github.com/kubernetes/kubernetes/blob/587b1f845a445e83a2288545f59cb22bf211cf52/pkg/kubelet/cm/container_manager_linux.go#L247-L253). `kernel.keys.root_maxbytes=25000000` is not needed.

So here's the behavior we need:
1. When running with `profile=cis-1.5`, let's keep our check for the kernel params as a convenience for the user so that they don't have to dig through kubelet logs. But DROP the check for `kernel.keys.root_maxbytes=25000000`.
2. When running with `profile=cis-1.5`, set the kubelet flag `protect-kernel-defaults` to true. NOTE: we expose this flag direclty in `rke2 agent` for the user to set. If the user has explicitly set it to false and supplied `profile=cis-1.5`, we should error out. If they have not explicitly set it, then we should just set it to true for them. ...This is just my initial thought on the behavior. open to alternatives.

rancher-max:
#### Validated on beta23

- When running with profile=cis-1.5, let's keep our check for the kernel params as a convenience for the user so that they don't have to dig through kubelet logs. But DROP the check for kernel.keys.root_maxbytes=25000000 -- This is working. If I don't have this value, rke2 still installs successfully, but if I don't have one of the other values, it fails to continue install giving errors like: `level=fatal msg="invalid kernel parameter value kernel.panic=-1 - expected 10\n"`
- When running with profile=cis-1.5, set the kubelet flag protect-kernel-defaults to true. -- This is working. When CIS mode is not set, this parameter is not set on the kubelet at all.
[...]

PR #391, opened 2020-09-29
Add Hardening Guide into Docs

briandowns:
#### Proposed Changes ####

Adds the hardening guide from the previous issue-327 PR (since closed for cleanliness reasons).

#### Linked Issues ####

#327


docs/hardening_guide.md
```diff
@@ -0,0 +1,403 @@
+# Hardening Guide
+
+This document provides prescriptive guidance for hardening a production installation of RKE2. It outlines the configurations and controls required to address Kubernetes benchmark controls from the Center for Information Security (CIS).
+
+If RKE2 is run in CIS mode, RKE2 applies a restrictive NetworkPolicy and PodSecurityPolicy. The restrictive NetworkPolicy allows for only namespace traffic with the exception of DNS and applies to `kube-system`, `kube-public`, and `default` namespaces. The restrictive PodSecurityPolicy addresses CIS controls defined in section 5.2. More details can be found below.
```
cjellick, line 5: I'd like "in CIS mode" replaced with "with the `profile` flag set to `cis-1.5`.
I'd also like us to enumerate everything that happens when you turn on this flag.

docs/hardening_guide.md
```diff
@@ -0,0 +1,403 @@
...
+#!/bin/bash -e
+
+for namespace in $(kubectl get namespaces -A -o json | jq -r '.items[].metadata.name'); do
+  kubectl apply -f default-allow-all.yaml -n ${namespace}
+done
+```
+
+Execute this script to apply the default-allow-all.yaml permissive NetworkPolicy to all namespaces.
+
+## Reference Hardened RKE2 cluster.yml configuration
```
cjellick, line 120: This entire section is not applicable to REK2.

docs/hardening_guide.md
briandowns: Remediated.

docs/hardening_guide.md
```diff
@@ -0,0 +1,52 @@
+# Hardening Guide
+
+This document provides prescriptive guidance for hardening a production installation of RKE2. It outlines the configurations and controls required to address Kubernetes benchmark controls from the Center for Information Security (CIS).
+
+If RKE2 is run with the profile flag set to "cis-1.5", it performs a number of system checks and applies a restrictive NetworkPolicy and PodSecurityPolicy.
```
cjellick, line 5:
```suggestion
If RKE2 is run with the `profile` flag set to "cis-1.5", it performs a number of system checks and applies a restrictive NetworkPolicy and PodSecurityPolicy. 
```

docs/hardening_guide.md
```diff
@@ -0,0 +1,52 @@
+# Hardening Guide
+
+This document provides prescriptive guidance for hardening a production installation of RKE2. It outlines the configurations and controls required to address Kubernetes benchmark controls from the Center for Information Security (CIS).
+
+If RKE2 is run with the profile flag set to "cis-1.5", it performs a number of system checks and applies a restrictive NetworkPolicy and PodSecurityPolicy. 
+
+RKE2 checks for 5 sysctl kernel parameters (see below) as well as the existence of the "etcd" user. If any of the sysctl kernel parameters aren't set to the expected value or the "etcd" users doesn't exist, RKE2 will immediately cease operation.
```
cjellick, line 7: I dropped this to four because thats all that the kubelet checks for. See: https://github.com/rancher/rke2/issues/389

Also as part of that issue, we now ensure the kubelet flag `protect-kernel-parameters` is set. Can you add that in here?

docs/hardening_guide.md
```diff
@@ -0,0 +1,52 @@
+# Hardening Guide
+
+This document provides prescriptive guidance for hardening a production installation of RKE2. It outlines the configurations and controls required to address Kubernetes benchmark controls from the Center for Information Security (CIS).
+
+If RKE2 is run with the profile flag set to "cis-1.5", it performs a number of system checks and applies a restrictive NetworkPolicy and PodSecurityPolicy. 
+
+RKE2 checks for 5 sysctl kernel parameters (see below) as well as the existence of the "etcd" user. If any of the sysctl kernel parameters aren't set to the expected value or the "etcd" users doesn't exist, RKE2 will immediately cease operation. 
+
+The restrictive NetworkPolicy allows for only namespace traffic with the exception of DNS and applies to `kube-system`, `kube-public`, and `default` namespaces. The restrictive PodSecurityPolicy addresses CIS controls defined in section 5.2. More details can be found below.
```
cjellick, line 9:
```suggestion
The restrictive NetworkPolicy allows for only traffic between pods in the same namespace with the exception of DNS. This restrictive policy only applies to the built-in namespaces: `kube-system`, `kube-public`, and `default` namespaces. Operators must manage network policies as normal for additional namespace.

The restrictive PodSecurityPolicy addresses CIS controls defined in section 5.2. More details can be found below.
```

docs/hardening_guide.md
```diff
@@ -0,0 +1,52 @@
...
+## Configure Kernel Runtime Parameters
+
+The following sysctl configuration is recommended for all nodes type in the cluster. Set the following parameters in `/etc/sysctl.d/90-kubelet.conf`.
+
+```sh
+vm.overcommit_memory=1
+vm.panic_on_oom=0
+kernel.panic=10
+kernel.panic_on_oops=1
+kernel.keys.root_maxbytes=25000000
```
cjellick, line 24: As per above, can drop this one. kubelet doesn't require it.

docs/hardening_guide.md
```diff
@@ -0,0 +1,52 @@
...
+
+The restrictive NetworkPolicy allows for only namespace traffic with the exception of DNS and applies to `kube-system`, `kube-public`, and `default` namespaces. The restrictive PodSecurityPolicy addresses CIS controls defined in section 5.2. More details can be found below.
+
+## Overview
+
+For more detail about evaluating a hardened cluster against the official CIS benchmark, refer to the [CIS Benchmark Rancher Self-Assessment Guide](cis_self_assessment.md).
+
+## Configure Kernel Runtime Parameters
+
+The following sysctl configuration is recommended for all nodes type in the cluster. Set the following parameters in `/etc/sysctl.d/90-kubelet.conf`.
```
cjellick, line 17: In one of the two install methods, i think the tar, I think @dweomer actually puts this file at a location somewhere in `/var/lib/rancher/rke2` ... so I think the instructions for this might be different for us. Actually, this is an inconsistency we might want to fix...need to talk to Jacob.

cjellick: It is the tar install method and its at `/usr/local/share/rke2/rke2-cis-sysctl.conf`

dweomer: Yes. An instruction to copy the sysctl content into `/etc/sysctl.d` should be sufficient.

docs/hardening_guide.md
```diff
@@ -0,0 +1,54 @@
...
+
+useradd -r -c "etcd user" -s /sbin/nologin -M etcd
+
+echo "Setting kernel parameters..."
+
+sysctl -w vm.panic_on_oom=0
+sysctl -w vm.overcommit_memory=1
+sysctl -w kernel.keys.root_maxbytes=25000000
+sysctl -w kernel.panic=10
+sysctl -w kernel.panic_on_oops=1
```
cjellick, line 49: I guess you'll need to update this to symlink from the file that we lay down.

cjellick (commented):
The content for this is spot on, but I have some thoughts for reorganizing it a bit. I'm going to copy the content, reorganize, and submit a new PR. I'm doing it this way, rather than merging and submitting a new PR, so that there's a single PR that introduces the content that others can review.

briandowns:
Changes look good.

cjellick (approved):

PR #394, opened 2020-09-29
[CIS-1.5] Correct validation issues and set kubelet arg properly

brandond:
#### Proposed Changes ####

Correcting issues from #389:
* Don't check kernel.keys.root_maxbytes
* Check to see if the user has set the protect-kernel-parameters to a value that conflicts with the profile's required setting.
* Set protect-kernel-defaults via existing CLI flag, instead of appending it to kubelet args.

#### Types of Changes ####

* CLI
* CIS Profile

#### Verification ####

* Start RKE2 with --profile=cis-1.5 flag while setting --protect-kernel-defaults=false; note error
* Start RKE2 with --profile=cis-1.5 flag with invalid kernel.keys.root_maxbytes value; note lack of error
* Start RKE2 with --profile=cis-1.5 flag; note --protect-kernel-defaults added to kubelet args

#### Linked Issues ####

#389

### DIFF
--- a/pkg/cli/cmds/root.go
+++ b/pkg/cli/cmds/root.go
@@ -46,6 +46,8 @@ var (
 	}
 )
 
+const pkdFlagName = "protect-kernel-defaults"
+
 func init() {
 	// hack - force "file,dns" lookup order if go dns is used
 	if os.Getenv("RES_OPTIONS") == "" {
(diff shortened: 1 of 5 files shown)


dweomer (approved):

pkg/cli/cmds/root.go
```diff
@@ -124,6 +125,16 @@ func validateCISreqs() error {
 	return nil
 }
 
+// setCISFlags validates and sets any CLI flags necessary to ensure
+// compliance with the profile.
+func setCISFlags(clx *cli.Context) error {
+	// If the user has specifically set this to false, raise an error
+	if clx.IsSet(pkdFlagName) && !clx.Bool(pkdFlagName) {
+		return fmt.Errorf("--%s must be true when using --profile=%s", pkgFlagName, profile)
```
dweomer, line 133:
```suggestion
		return fmt.Errorf("--%s must be true when using --profile=%s", pkdFlagName, profile)
```

pkg/cli/cmds/root.go
```diff
@@ -62,11 +64,10 @@ func init() {
 // kernel.panic_on_oops=1
 // kernel.keys.root_maxbytes=25000000
```
cjellick, line 65: drop maxbytes from comment

pkg/cli/cmds/root.go
brandond: thanks... not sure why my editor didn't catch that one.

pkg/cli/cmds/root.go
brandond: dropped all args from comments, no reason to duplicate them

cjellick (approved):

PR #397, opened 2020-09-29
Update Controls 1.2.21 - 1.2.35

briandowns:
#### Proposed Changes ####

Update Controls 1.2.21 - 1.2.35

#### Linked Issues ####

#327

### DIFF
--- a/docs/cis_self_assessment.md
+++ b/docs/cis_self_assessment.md
@@ -1210,30 +1210,16 @@ Where `etcd` encryption is used, it is important to ensure that the appropriate 
 Follow the Kubernetes documentation and configure a `EncryptionConfig` file.
 In this file, choose **aescbc**, **kms** or **secretbox** as the encryption provider.
 
-**Audit Script:** 1.2.34.sh
+**Audit:** 
+Run the below command on the master node.
 
-```
-#!/bin/bash -e
-
-check_file=${1}
-
-grep -q -E 'aescbc|kms|secretbox' ${check_file}
-if [ $? -eq 0 ]; then
-echo "--pass"
-exit 0
-else
-echo "fail: encryption provider found in ${check_file}"
-exit 1
-fi
+```bash
+grep aescbc /var/lib/rancher/rke2/server/cred/encryption-config.json
 ```
 
-**Audit Execution:**
+Run the below command on the master node.
 
-```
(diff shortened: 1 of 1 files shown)


docs/cis_self_assessment.md
```diff
@@ -1294,26 +1184,17 @@ etcd is a highly available key-value store used by Kubernetes deployments for pe
...
 ```
 
-**Expected result**:
+Verify that the `--encryption-provider-config` argument is set to a EncryptionConfigfile. Additionally, ensure that the `EncryptionConfigfile` has all the desired resources covered especially any secrets.
 
-```
-'--encryption-provider-config' is present
-```
+**Remediation:**
+By default, RKE2 sets the `--encryption-provider-config` argument explicitly. No manual remediation needed.
```
cjellick, line 1197:
```suggestion
By default, RKE2 sets the `--encryption-provider-config` argument explicitly. No manual remediation needed. RKE2's default encryption provider config file is located at `/var/lib/rancher/rke2/server/cred/encryption-config.json` and is configured to encrypt secrets.
```

docs/cis_self_assessment.md
```diff
@@ -1349,14 +1230,13 @@ fi
 **Audit Execution:**
 
 ```
```
cjellick, line 1232: 1.2.34 deviates far from the norm. it has the old style audit script and audit execetution. Is this inentoinal?

Also lloks like remediation is in there twice

cjellick: for the audit for this one, I would just give a command to cat the encryption config file and tell them to verify that the provider is aescbc

docs/cis_self_assessment.md
```diff
@@ -1367,10 +1247,19 @@ Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Not S
...
 
-**Remediation:**
+**Audit:**
+Run the below command on the master node.
 
+```bash
+/bin/ps -ef | grep kube-apiserver | grep -v grep
+```
+
+Verify that the `--tls-cipher-suites` argument is set as outlined in the remediation procedure below.
```
cjellick, line 1259: We could possibly "do better" and set this flag as the upstread CIS doc outlines. But that is not a blocker. Can you open a separate issue for implementing this, but leave this doc as-is for now, as what we are doing is acceptable for ga

cjellick (approved):

PR #409, opened 2020-10-01
Fix profile flag for static pods

brandond:
#### Proposed Changes ####

Stop ranging over CLI flags; the type checks are tedious and error-prone.

#### Types of Changes ####

* CLI
* Security Profile

#### Verification ####

* Start RKE2 with --profile cis-1.5; note etcd pod is running as etcd user and /var/lib/rancher/rke2/server/db/etcd is owned by etcd not root.
* Start RKE2 with --data-dir=/var/lib/rancher/test; note that this path is used instead of /var/lib/rancher/rke2

#### Linked Issues ####

#351
Fixes errors turned up in QA validation of #387.

#### Further Comments ####

This was in the original version of #387 but removed in the name of minimalism. Ranging over the flags is error-prone because there are many different types, and we have a mix of flag and *flag, and if you're not careful it's easy to get wrong. The urfave API has a simple interface for doing this that we should be using instead.

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -55,7 +55,7 @@ func NewAgentCommand() cli.Command {
 
 func AgentRun(clx *cli.Context) error {
 	switch profile {
-	case "cis-1.5":
+	case rke2.CISProfile:
 		if err := validateCISReqs("agent"); err != nil {
 			logrus.Fatal(err)
 		}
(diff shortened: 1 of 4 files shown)


cjellick:
We can just associate to the existing issue: https://github.com/rancher/rke2/issues/351

dweomer (approved):

brandond:
CI is having issues, merging anyway.

Issue #226, opened 2020-08-21
[Documentation] cgroups v2 support


davidnuzik:
If on an OS that supports cgroups v2 -- rke2 will need to run with v1 or a hybrid. v2 support doesn't come until containerd v1.4.0 and this is targeted for k8s v1.19.0 -- We may not have this when we GA so this should get called out in docs.

davidnuzik:
Closing. QA is going to take a holistic look at the docs and this will be included. QA will enter PRs as needed for any changes (not that I foresee the need for any).

PR #423, opened 2020-10-02
docs: cgroups v2 call-out

dweomer:
Provided a short description of the likelihood of, and mitigation
required, for running RKE2 on a cgroups v2 capable system due to
shipping with containerd v1.3.6 that does not support cgroups v2.

Addresses #226


davidnuzik (approved):
LGTM as far as spelling, grammar, and readability. The content seems pretty clear to me and to the point. Would still like at least 1 lgtm from a dev.

briandowns (approved):

PR #437, opened 2020-10-03
clarify agent vs server differences in quickstart docs

joshrwolf:
#### Proposed Changes ####

Clarify differences between server and agent install using the install script in the quickstart.md

#### Types of Changes ####

Documentation change

#### Further Comments ####

Install script will only create the service file of the corresponding `INSTALL_RKE2_TYPE`, will `server` being the default.

### DIFF
--- a/docs/install/quickstart.md
+++ b/docs/install/quickstart.md
@@ -16,7 +16,7 @@ RKE2 provides an installation script that is a convenient way to install it as a
 ```
 curl -sfL https://get.rke2.io | sh -
 ```
-This will install `rke2-server` and `rke2-agent` services onto your machine.
+This will install the `rke2-server` service and the `rke2` binary onto your machine.
 
 **Enable the rke2-server service**
 ```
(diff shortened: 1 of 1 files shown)


cjellick:
Thanks! And more docs will be landing today and tomorrow. Please continue to take a look if you have time

docs/install/quickstart.md
```diff
@@ -16,7 +16,7 @@ RKE2 provides an installation script that is a convenient way to install it as a
 ```
 curl -sfL https://get.rke2.io | sh -
 ```
-This will install `rke2-server` and `rke2-agent` services onto your machine.
+This will install the `rke2-server` service and the `rke2` binary onto your machine.
```
cjellick, line 19: I had both services here because they do both get installed.

Do you think that’s just too much noise for a QuickStart?

cjellick, line 19: I had both services here because they do both get installed.

Do you think that’s just too much noise for a QuickStart?

joshrwolf: It looks like only the service of the respective node type is being installed, at least with the auto detecting yum method on centos7/rhel7

cjellick: Yeah, my bad. It’s both on tar

dweomer, line 19:
```suggestion
This will install the `rke2-server` systemd service and the `rke2` binary onto your machine.
```

dweomer: It is both with tarball but for a quickstart I can see the argument for leaving out that detail.

dweomer (approved):

PR #440, opened 2020-10-03
Add rke2-upgrade docs

MonzElmasry:
This pr adds the possible upgrade scenarios for rke2 clusters

### DIFF
--- a/docs/basic_upgrade.md
+++ b/docs/basic_upgrade.md
@@ -7,12 +7,9 @@ You can upgrade rke2 by using the installation script, or by manually installing
 
 ### Release Channels
 
-Upgrades performed via the installation script or using our [automated upgrades](automated_upgrade.md) feature can be tied to different release channels. The following channels are available:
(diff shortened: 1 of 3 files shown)


docs/basic_upgrade.md
```diff
@@ -0,0 +1,62 @@
...
+
+>**Note:** When upgrading, upgrade server nodes first one at a time, then any worker nodes.
+
+### Release Channels
+
+Upgrades performed via the installation script or using our [automated upgrades](automated_upgrade.md) feature can be tied to different release channels. The following channels are available:
+
+| Channel |   Description  |
+|---------------|---------|
+|      stable     | (Default) Stable is recommended for production environments. These releases have been through a period of community hardening. |
```
cjellick, line 14: We are not going to have a stable channel upon initial release. Can you remove this table and instead have:

Currently, the `latest` channel is the only available channel. Once we have more releases and need to distinguish between the most recent release and the most stable release, we will add a `stable` channel and set it as the default.

cjellick (approved):

PR #448, opened 2020-10-04
Add installation requirements docs

galal-hussein:


docs/install/requirements.md
```diff
@@ -0,0 +1,51 @@
+RKE2 is very lightweight, but has some minimum requirements as outlined below.
+
+## Prerequisites
+
+Two nodes cannot have the same hostname.
+
+If all your nodes have the same hostname, use `--node-name` or `$RKE2_NODE_NAME` for each node you add to the cluster to have a different node name.
```
cjellick, line 7: i'm trying to make reference to config file params rather than cmd line args or env vars
```suggestion
If all your nodes have the same hostname, set the `node-name` parameter in the RKE2 config file for each node you add to the cluster to have a different node name.
```

galal-hussein: got it, will fix

docs/install/requirements.md
```diff
@@ -0,0 +1,51 @@
...
+## Hardware
+
+Hardware requirements scale based on the size of your deployments. Minimum recommendations are outlined here.
+
+*    RAM: 512MB Minimum (we recommend at least 1GB)
+*    CPU: 1 Minimum
+
+#### Disks
+
+RKE2 performance depends on the performance of the database, and since RKE2 runs etcd embeddedly and it stores the data dir on disk. we recommend using an SSD when possible to ensure optimal performance.
```
cjellick, line 27:
```suggestion
RKE2 performance depends on the performance of the database, and since RKE2 runs etcd embeddedly and it stores the data dir on disk, we recommend using an SSD when possible to ensure optimal performance.
```

docs/install/requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Networking
+
+The RKE2 server needs port 6443 and 9345 to be accessible by the nodes.
```
cjellick, line 31:
```suggestion
The RKE2 server needs port 6443 and 9345 to be accessible by other nodes in the cluster.
```

docs/install/requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+The nodes need to be able to reach other nodes over UDP port 8472 when Flannel VXLAN is used.
```
cjellick, line 33:
```suggestion
All nodes need to be able to reach other nodes over UDP port 8472 when Flannel VXLAN is used.
```

docs/install/requirements.md
briandowns, line 31: Should we include what uses each port?

docs/install/requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+If you wish to utilize the metrics server, you will need to open port 10250 on each node.
+
+> **Important:** The VXLAN port on nodes should not be exposed to the world as it opens up your cluster network to be accessed by anyone. Run your nodes behind a firewall/security group that disables access to port 8472.
```
cjellick, line 37: I dont like how the sections that start with `>` are rendered on our docs site.

```suggestion
**Important:** The VXLAN port on nodes should not be exposed to the world as it opens up your cluster network to be accessed by anyone. Run your nodes behind a firewall/security group that disables access to port 8472.
```

docs/install/requirements.md
galal-hussein: I elaborated on that in the table below

briandowns (approved):

cjellick (approved):

PR #459, opened 2020-10-05
Installation options

cjellick:


docs/install/install_options/install_options.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Options
```
davidnuzik, line 2: Nit:
(I missed this in my original) Can we just call it "Overview" instead of "Options" because in the nav tree this looks funny to me:
```
Install Options:
        Options
```

docs/install/install_options/install_options.md
```diff
@@ -0,0 +1,89 @@
...
+The simplest form of this command is as follows:
+```sh
+curl -sfL https://get.rke2.io | sh -
+```
+
+When using this method to install RKE2, the following environment variables can be used to configure the installation:
+
+| Environment Variable | Description |
+|-----------------------------|---------------------------------------------|
+| <span style="white-space: nowrap">`INSTALL_RKE2_VERSION`</span> | Version of RKE2 to download from GitHub. Will attempt to download from the latest channel if not specified. |
```
davidnuzik, line 28: You'll need to remove the `<span>` tags throughout install_options.md as I didn't get that in before you ran with what I had.

docs/install/install_options/install_options.md
```diff
@@ -0,0 +1,89 @@
...
+  --tls-san "foo.local"             \
+  --node-label "foo=bar"            \
+  --node-label "something=amazing"
+```
+
+It is also possible to use both a configuration file and CLI arguments.  In these situations, values will be loaded from both sources, but CLI arguments will take precedence.  For repeatable arguments such as `--node-label`, the CLI arguments will overwrite all values in the list.
+
+Finally, the location of the config file can be changed either through the cli argument `--config FILE, -c FILE`, or the environment variable `$RKE2_CONFIG_FILE`.
+
+### Configuring when Running the Binary Directly]
```
davidnuzik, line 79: Remove erroneous ]

davidnuzik (changes requested):

briandowns (approved):
Aside from @davidnuzik 's comments, LGTM

PR #411, opened 2020-10-01
set the default install channel= latest once v1.18.9+rke2r1 is released

MonzElmasry:
#### Proposed Changes ####

Change the default install channel for rke2 to latest when v1.18.9+rke2r1 is released

#### Types of Changes ####

Install script change 

#### Verification ####

When merged, run install.sh with no passed INSTALL_RKE2_CHANNEL param, should install the latest release 

#### Linked Issues ####

https://github.com/rancher/rke2/issues/403

Issue #403, opened 2020-09-30
Update default release channel to latest


davidnuzik:
https://github.com/rancher/rke2/blob/14738559266ef5fbb3fecc7a9d41c73d423720f1/install.sh#L63

We need to update the default release channel to latest post release.

davidnuzik:
We'll GA and then right after set latest to default.
Then, when we are ready to mark the release (or future release) stable, we'll do so via rancher/rke2#404

davidnuzik:
I can confirm the default channel is indeed latest now (based on testing) and I successfully installed the latest release (GA).

PR #460, opened 2020-10-05
install: latest should be the default channel

dweomer:
Addresses #403
Supplants #411


Oats87 (approved):
LGTM

briandowns (approved):

MonzElmasry (commented):
can we remove https://github.com/rancher/rke2/blob/e5dbd9be0a6b618aead50a6f4a6e657df8bcc829/install.sh#L136 as well?

Issue #309, opened 2020-09-11
Add node-external-ip to rke2


cjellick:
k3s supports the --node-external-ip flag, but rke2 doesn't.

We should add this field to RKE2, we just need to make it clear in the help text and docs that the field is ignored if you have explicitly set a cloud provider. In fact, we should probably error out if we see a combination of the node-external0p and cloud-provider flags

davidnuzik:
A second PR is necessary to enable the built-in cloud provider.

ShylajaDevadiga:
Validating from master passing --node-external-ip, also validated expected error while passing cloud-provider
```
 rke2 server --node-external-ip=PUBLICIP  2>&1

kubectl get nodes -o wide
NAME              STATUS   ROLES         AGE   VERSION                 INTERNAL-IP    EXTERNAL-IP      OS-IMAGE             KERNEL-VERSION   CONTAINER-RUNTIME
ip-172-31-38-48   Ready    etcd,master   55s   v1.18.10-dev+c5354973   172.31.38.48   PUBLICIP   Ubuntu 20.04.1 LTS   5.4.0-1024-aws   containerd://1.3.6-k3s2

rke2 server  --node-external-ip=PUBLICIP  --cloud-provider-name="aws" 2>&1
WARN[0000] not running in CIS 1.5 mode                  
INFO[0000] Runtime image "index.docker.io/rancher/rke2-runtime:v1.18.10-dev-c5354973" bin dir already exists at "/var/lib/rancher/rke2/data/v1.18.10-dev-c5354973-9445e60437bc/bin"; skipping extract 
FATA[0000] can't set node-external-ip while using cloud provider 
```

PR #494, opened 2020-10-19
Add support for node-external-ip flag

MonzElmasry:
#### Proposed Changes ####

- Add support for node-external-ip flag and
- Add validation for node-external-ip as it can not be used along with cloud-provider flags

#### Types of Changes ####
New Feature

#### Verification ####

Just like k3s, check the funtionality of `node-external-ip` flag, and try to set `node-external-ip` along with any of the cloud provider flags to make sure it errors out. 
#### Linked Issues ####

https://github.com/rancher/rke2/issues/309

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -1,6 +1,7 @@
 package rke2
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/rke2/rke2.go
```diff
@@ -86,7 +86,9 @@ func setup(clx *cli.Context, cfg Config) error {
 	agentImagesDir := filepath.Join(dataDir, "agent", "images")
 
 	managed.RegisterDriver(&etcd.ETCD{})
-
+	if clx.IsSet("node-external-ip") && (clx.IsSet("cloud-provider-config") || clx.IsSet("cloud-provider-name")) {
+		return fmt.Errorf("Can't set node-external-ip while using cloud provider")
```
briandowns, line 90: No need to use the format error creation method here since we're not formatting anything. Also, we don't want to capitalize error strings. Link below for reference.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

```suggestion
		return errors.New("can't set node-external-ip while using cloud provider")
```

brandond (approved):

brandond:
LGTM other than the stuff that Brian requested

briandowns (approved):
LGTM

Issue #410, opened 2020-10-01
Cannot properly configure audit-logging in the API Server pod


briandowns:
Per CIS requirements, we need to add the `--audit-policy-file` argument to the kube-apiserver. This argument needs to point at a file containing a valid Policy resource:
```yaml
apiVersion: audit.k8s.io/v1
kind: Policy
rules: []
```

We may want to provide a default file with no rules (as shown above) that the user can customize.

In order to pass this file through to the apiserver static pod, we will need to add it to the host path mounts. We will likely need to do the same for the audit log file. Suggested paths and audit options from the hardening guide are:

```
--kube-apiserver-arg=audit-policy-file=/etc/rancher/rke2/audit/policy.yaml
--kube-apiserver-arg=audit-log-path=/var/log/rke2/audit.log
--kube-apiserver-arg=audit-log-maxage=31
--kube-apiserver-arg=audit-log-maxbackup=100
--kube-apiserver-arg=audit-log-maxsize=100
```

cjellick:
We need to mount in a directory from the host that will have the audit policy file and we need to be able to get the audit log files off the host.

dweomer:
> We need to mount in a directory from the host that will have the audit policy file and we need to be able to get the audit log files off the host.

Should we be opinionated here and force/assume those locations?

briandowns:
> > We need to mount in a directory from the host that will have the audit policy file and we need to be able to get the audit log files off the host.
> 
> Should we be opinionated here and force/assume those locations?

I think it'd probably be safe with `/var/lib/rancher` as the assumed root.

galal-hussein:
I am going to add a flag for the policy file the same way we do for the registries.yaml file, the default place will be in the same config file /etc/rancher/rke2/policy.yaml file and will be an empty policy by default, as for the log file I will assume datadir/server/logs/audit.log the same way we do for kubelet logs under agent dir, and the rest will be hardcoded and can be overriden with extra args

brandond:
It needs to be a valid policy file, but the example above (with no rules) should work. Maybe don't overwrite it if it already exists so that users can customize it?

rancher-max:
#### Validated using `v1.18.10-alpha1+rke2r1`

- There continues to be no audit logging by default when NOT in CIS mode
[...]

PR #511, opened 2020-10-30
Add audit policy and log files to rke2 in cis mode

galal-hussein:
#### Proposed Changes ####

Adding both audit log and policy file to rke2, and adding a flag for overriding the default place for policy file for using custom policy

#### Types of Changes ####

- Add a flag for policy file --audit-policy-file
- adding a default policy file with only one rule to be able to pass cis check
```
apiVersion: audit.k8s.io/v1
kind: Policy
metadata:
  creationTimestamp: null
rules:
- level: Metadata
  resources:
  - resources:
    - pods
```
- mount both files within kubeapi pod (mounting the dir for logs because api renames the audit log files)

#### Verification ####

- run rke2 in cis mode
- check for audit logs in /var/lib/rancher/rke2/server/logs/audit.log
- also run rke2 with flag --audit-policy-file <custom-policy> and point point to a custom policy file

rke2 should be able to use this new custom policy file

#### Linked Issues ####

https://github.com/rancher/rke2/issues/410

#### Further Comments ####

this pr checked against the cis benchmarks  and passed the audit log cases.

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"net/http"
 	"os"
 	"os/exec"
@@ -20,6 +21,7 @@ import (
 	"github.com/rancher/rke2/pkg/staticpod"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"sigs.k8s.io/yaml"
 )
@@ -34,12 +36,13 @@ var (
 )
 
 type StaticPodConfig struct {
-	ManifestsDir  string
-	ImagesDir     string
-	Images        images.Images
-	CloudProvider *CloudProviderConfig
-	CISMode       bool
-	DataDir       string
+	ManifestsDir    string
+	ImagesDir       string
+	Images          images.Images
+	CloudProvider   *CloudProviderConfig
+	CISMode         bool
+	DataDir         string
+	AuditPolicyFile string
 }
 
 type CloudProviderConfig struct {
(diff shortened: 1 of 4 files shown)


briandowns (approved):
LGTM

pkg/podexecutor/staticpod.go
```diff
@@ -107,9 +118,9 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
 			Command:   "kube-apiserver",
 			Args:      args,
 			Image:     s.Images.KubeAPIServer,
-			Dirs:      ssldirs,
+			Dirs:      append(ssldirs, filepath.Dir(auditLogFile)),
```
brandond, line 121: Don't we need to mount the audit policy file into the pod as well as the log file?

pkg/podexecutor/staticpod.go
```diff
@@ -284,4 +295,45 @@ func chownr(path string, uid, gid int) error {
...
+}
+
+func writeDefaultPolicyFile(policyFilePath string) error {
+	auditPolicy := auditv1.Policy{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Policy",
+			APIVersion: "audit.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{},
+		Rules: []auditv1.PolicyRule{
```
brandond, line 307: I don't think we should configure any rules by default, should we? CIS just says you have to enable logging and configure a policy, it doesn't say that the policy needs to have any rules. If it **does** need to have rules, can we add a comment with a reference to what these are? Auditing just pods by default seems awfully arbitrary.

brandond (changes requested):

pkg/podexecutor/staticpod.go
galal-hussein: nah, the code automatically mount all argument files https://github.com/rancher/rke2/blob/aeb2f9e07e2cb622380b02a360b771bc31941dc9/pkg/staticpod/staticpod.go#L229

pkg/podexecutor/staticpod.go
galal-hussein: kubeapi doesnt accept a policy with no rule, it fatals with an error that you should add at least one rule

pkg/podexecutor/staticpod.go
brandond: Ah, I missed that bit of magic. Cool!

pkg/podexecutor/staticpod.go
```diff
@@ -284,4 +295,45 @@ func chownr(path string, uid, gid int) error {
...
+			auditv1.PolicyRule{
+				Level: "Metadata",
+				Resources: []auditv1.GroupResources{
+					auditv1.GroupResources{
+						Group:     "",
+						Resources: []string{"pods"},
+					},
+				},
```
brandond, line 315: Valid rule with no logging by default:

```suggestion
				Level: "None",
```

brandond (changes requested):

pkg/podexecutor/staticpod.go
galal-hussein: I will test it out and will change it if it works, thanks

brandond (approved):

Issue #463, opened 2020-10-05
Fix el7 versioning


davidnuzik:
Note: Right now there is functionally no difference between our el7 and el8 rpms, so this has no actual impact to v1.18.9+rke2r1 people upgrading in the future may see a bump from el7 to el8.

The install script will need to detect between el7 and el8 and get the appropriate RPMS.

davidnuzik:
@ShylajaDevadiga this already got tested - can we close?

Oats87:
@davidnuzik This still needs to be corrected from the install script. el8 should end up with el8 rpms, el7 should end up with el7 rpms and so forth

davidnuzik:
cc: @cjellick this needs assignment? Take out of To Test?

ShylajaDevadiga:
@davidnuzik We will need install script fix for rhel8. Its cosmetic. Works as expected but needs the fix.

Oats87:
We may want to take inspiration from what we're doing in our `install-docker` script:

https://github.com/rancher/install-docker/commit/9a21b08b6651c92e0b323c491835400f3c3e6bb7#diff-22b308d81eb1d73a60b7c5c5c33881031f5b188118eefb03d66ae789b539c7b9

ShylajaDevadiga:
Validated  as per PR in CentOS 7/8, Redhat 7/8 and Fedora 33
Version:
```
rke2 -v
rke2 version v1.18.10+rke2r1 (8e418583df2c62932657afd3c05973b577064cf5)
```
Redhat 7.8
```
[ec2-user@ip-172-31-46-205 ~]$ rpm -qa |grep rke2
rke2-common-1.18.10~rke2r1-2.el7.x86_64
rke2-selinux-0.3-1.el7_8.noarch
rke2-server-1.18.10~rke2r1-2.el7.x86_64
[ec2-user@ip-172-31-46-205 ~]$ sudo cat /etc/yum.repos.d/rancher-rke2-testing.repo 
[rancher-rke2-common-testing]
name=Rancher RKE2 Common (testing)
baseurl=https://rpm-testing.rancher.io/rke2/testing/common/centos/7/noarch
enabled=1
gpgcheck=1
gpgkey=https://rpm-testing.rancher.io/public.key
[rancher-rke2-1-18-testing]
name=Rancher RKE2 1.18 (testing)
baseurl=https://rpm-testing.rancher.io/rke2/testing/1.18/centos/7/x86_64
enabled=1
gpgcheck=1
[...]

PR #516, opened 2020-11-03
Switch EL repo version based on detected EL version

Oats87:
#### Proposed Changes ####
Update `install.sh` to install RPMs based on the detected EL version
#### Types of Changes ####
Bug Fix

#### Verification ####
Run the `install.sh` script on
CentOS 7 - `7`
CentOS 8 - `8`
RHEL 7 - `7`
RHEL 8 - `8`
Fedora 33 - `7`

Once installed, inspect the `RPM`s installed as well as the `/etc/yum.repos.d/rancher-rke2-*.repo` file and verify that the `7` or `8` is set as shown above.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/463

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -218,16 +218,29 @@ do_install_rpm() {
     if [ "${1}" = "testing" ]; then
         rpm_site="rpm-${1}.rancher.io"
     fi
+    maj_ver="7"
+    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+        dist_version="$(. /etc/os-release && echo "$VERSION_ID")"
+        maj_ver=$(echo "$dist_version" | sed -E -e "s/^([0-9]+)\.?[0-9]*$/\1/")
+        case ${maj_ver} in
+            7|8)
+                :
+                ;;
+            *) # In certain cases, like installing on Fedora, maj_ver will end up being something that is not 7 or 8
+                maj_ver="7"
+                ;;
+        esac
+    fi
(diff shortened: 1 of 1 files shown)


dweomer (approved):

install.sh
```diff
@@ -218,16 +218,29 @@ do_install_rpm() {
...
+    maj_ver="7"
+    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+        dist_version="$(. /etc/os-release && echo "$VERSION_ID")"
+        maj_ver=$(echo "$dist_version" | sed -E -e "s/^([0-9]+)\.?[0-9]*$/\1/")
+        case ${maj_ver} in
+            7|8)
+                :
+                ;;
+            *) # In certain cases, like installing on Fedora, maj_ver will end up being something that is not 7 or 8
+                maj_ver="7"
```
brandond, line 230: In case of Fedora or other bleeding-edge distros, shouldn't we default to the EL8 packages?

Oats87: There is a case to be made to rely on the EL8 packages for fedora, but due to the fact that you must use `2.124` or newer of `container-selinux` with the `el8` `rke2-selinux` compiled policy, I didn't want to force the hand in the event that someone was running an older Fedora version. The `el7` policies/rpms are identical to the `el8` rpms and should work on all fedora systems.

galal-hussein (approved):
LGTM

briandowns (approved):
LGTM

install.sh
brandond: As long as they work!

brandond (approved):

PR #548, opened 2020-11-23
install release from stable channel by default

MonzElmasry:
#### Proposed Changes ####

install release from stable channel by default 
#### Types of Changes ####

channelserver

#### Verification ####

 `curl -sfL https://get.rke2.io | sudo sh -`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -60,7 +60,7 @@ setup_env() {
 
     # --- make sure install channel has a value
     if [ -z "${INSTALL_RKE2_CHANNEL}" ]; then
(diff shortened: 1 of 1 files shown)


davidnuzik (approved):

davidnuzik:
QA has checked. When override install script to stable channel now, it works on Ubuntu and RHEL. As such, I'm going to go ahead and merge this in now that this is fixed. (we needed to publish rke2-packaging).

In order to prevent mixups like this in the future we'll need to document this process so it is more clear what is necessary before marking an rke2 release as stable.

PR #543, opened 2020-11-20
Add CLI flags to override images and kubelet binary

brandond:
#### Proposed Changes ####

Allow overriding images and kubelet binary path from CLI or config file

#### Types of Changes ####

* New feature

#### Verification ####

* `rke2 --help`  -- look for new CLI flags
* `rke2 server --kubelet-path=X`  -- after downloading custom kubelet to X
* `rke2 server --kube-apiserver-image=X` -- where X is an alternative apiserver Docker image name

#### Linked Issues ####

#542

PR #552, opened 2020-11-24
Add custom image / kubelet example script

brandond:
#### Proposed Changes ####

Add a Python script that generates config files and extracts binaries from a Release YAML manifest. The resulting configuration will replace core components with those from an alternative Kubernetes distribution.

#### Types of Changes ####

* contrib

#### Verification ####

Follow instructions in readme

#### Linked Issues ####

#543 

#### Further Comments ####
cc @cloudnautique

### DIFF
--- a/contrib/custom-image-kubelet/genconfig.py
+++ b/contrib/custom-image-kubelet/genconfig.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+import io
+import logging
+import click
+import requests
+import boto3
+import tarfile
+from base64 import b64decode
+from yaml import load, dump
+from os.path import basename, normpath
+from collections import defaultdict
+from urllib.parse import urlparse
+from shutil import copyfileobj
+
+try:
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
+
+
+IMAGES = ['kube-apiserver', 'kube-controller-manager', 'kube-scheduler', 'pause', 'etcd']
+COMPONENTS = ['kube-proxy', 'coredns', 'metrics-server']
+
+
+@click.command(context_settings={'max_content_width': 120})
+@click.option('--release-url', type=str, required=True, help='URL to a Kubernetes Release YAML')
+@click.option('--data-dir', type=str, default='/var/lib/rancher/rke2', help='RKE2 data directory', show_default=True)
+@click.option('--prefix', type=str, default='/', help='Prefix for output files created by this script', show_default=True)
+def main(release_url, data_dir, prefix):
+    release = get_release(release_url)
+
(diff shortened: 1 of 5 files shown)


contrib/custom-image-kubelet/genconfig.py
```diff
@@ -0,0 +1,217 @@
...
+    for component in COMPONENTS:
+        alt_image = get_image_artifact(release, component)
+        write_chart_config(component, alt_image, f'{prefix}/{data_dir}')
+
+    extract_archive(release, release_url, 'kubernetes-node-linux-amd64.tar.gz', f'{prefix}/{data_dir}')
+
+    try:
+        write_ecr_credentials(release, prefix)
+    except Exception as e:
+        logging.warning(f'Unable to write ECR credenials to registries.yaml: {e}')
```
briandowns, line 56: Should we exit at this point?

contrib/custom-image-kubelet/genconfig.py
```diff
@@ -0,0 +1,217 @@
...
+
+    write_rke2_config(rke2_config, prefix)
+
+
+def get_release(release_url):
+    response = requests.get(release_url)
+    response.raise_for_status()
```
briandowns, line 63: Should we catch this RequestException if raised here rather than let it go uncaught?

dweomer (approved):

contrib/custom-image-kubelet/genconfig.py
```diff
@@ -0,0 +1,217 @@
...
+                if '.ecr.' in registry:
+                    registries.add(registry)
+    return registries
+
+
+if __name__ == '__main__':
+    try:
+        logging.basicConfig(format="%(levelname).1s %(message)s", level=logging.INFO, stream=sys.stdout)
+        main()
+    except (KeyboardInterrupt, BrokenPipeError):
```
briandowns, line 214: We want to ignore ctrl-c?

brandond: No, at this point the exception has already broken out of main(). This just makes it exit cleanly instead of printing an error.

contrib/custom-image-kubelet/genconfig.py
brandond: It'll get caught and pretty-printed by the bit down at https://github.com/rancher/rke2/pull/552/files#diff-e81b4d596775d35ffea04790f3aca8eb5c7ef803b51e6a0860fd66fa7c65c71aR217

briandowns: Cool. Good enough going top level exception for this.

contrib/custom-image-kubelet/genconfig.py
brandond: Fixed the typo, moved earlier, and converted to an exception.

briandowns (approved):
LGTM

Issue #483, opened 2020-10-13
Pod stuck in blocked on CentOS 7/8 when running with cis 1.5 profile


curantes:
When I try to deploy anything I get:
```
Status:         Pending
Reason:         AppArmor
Message:        Cannot enforce AppArmor: AppArmor is not enabled on the host
```
Since I run on Centos 8 I have SELinux enforcing but no Apparmor. So I edited the default PodSecurityPolicy for cis 1.5 profile and removed the annotations for apparmor to get the pods running. I am not sure if its correct to fix but as of now the default behavior breaks for Centos 8 users.

https://github.com/rancher/rke2/blob/3ba3b9d5326be42d07efe8bc1bab8c2b4857e51c/pkg/rke2/psp_templates.go#L37-L40

ChristopherEdwards:
I've encountered this same issue on a clean CentOS 8 install.

cjellick:
should backport fix to 1.18 as well

Oats87:
This is observed as happening on CentOS 7 systems as well. 
```
[root@ck-centos7-0 rke2]# cat /etc/rancher/rke2/config.yaml
profile: cis-1.5
[root@ck-centos7-0 rke2]# /var/lib/rancher/rke2/data/v1.18.12-rke2r2-a10e0031baf8/bin/kubectl create deployment test --image oats87/diag
deployment.apps/test created
[root@ck-centos7-0 rke2]# /var/lib/rancher/rke2/data/v1.18.12-rke2r2-a10e0031baf8/bin/kubectl get pod
NAME                   READY   STATUS    RESTARTS   AGE
test-687d5f778-66xsd   0/1     Blocked   0          6s
[root@ck-centos7-0 rke2]# /var/lib/rancher/rke2/data/v1.18.12-rke2r2-a10e0031baf8/bin/kubectl describe pod
Name:           test-687d5f778-66xsd
Namespace:      default
Priority:       0
Node:           ck-centos7-0/172.16.134.19
Start Time:     Tue, 08 Dec 2020 17:18:46 -0800
Labels:         app=test
                pod-template-hash=687d5f778
Annotations:    container.apparmor.security.beta.kubernetes.io/diag: runtime/default
                kubernetes.io/psp: global-restricted-psp
                seccomp.security.alpha.kubernetes.io/pod: runtime/default
Status:         Pending
Reason:         AppArmor
Message:        Cannot enforce AppArmor: AppArmor is not enabled on the host
IP:
[...]

PR #584, opened 2020-12-09
Remove PSP Annotations and Update Documentation

briandowns:
#### Proposed Changes ####

Removes all annotations from PSP's and updates the documentation reflecting the change.

#### Verification ####

Run RKE2 on CentOS 7/8 in CIS mode and verify that pods in a deployment are successfully launched.

#### Linked Issues ####

#483

### DIFF
--- a/pkg/rke2/psp_templates.go
+++ b/pkg/rke2/psp_templates.go
@@ -33,11 +33,6 @@ const globalRestrictedPSPTemplate = `apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: %s
-  annotations:
-    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default,runtime/default'
-    apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
-    seccomp.security.alpha.kubernetes.io/defaultProfileName: 'runtime/default'
-    apparmor.security.beta.kubernetes.io/defaultProfileName: 'runtime/default'
 spec:
   privileged: false                # CIS - 5.2.1
   allowPrivilegeEscalation: false  # CIS - 5.2.5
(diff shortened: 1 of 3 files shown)


Oats87 (approved):
LGTM. I'm curious if this will apply retroactively to existing RKE2 clusters that upgrade...

galal-hussein (approved):
LGTM

pkg/rke2/psp_templates.go
```diff
@@ -33,11 +33,6 @@ const globalRestrictedPSPTemplate = `apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
```
brandond, line 33: This policy looks like it came from the second example policy here:  https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies

If so, that has a comment in it that says: 
>     # This policy assumes the nodes are using AppArmor rather than SELinux.
but we are pulling out all the AppArmor stuff, which would appear to leave it a bit toothless. I don't have a suggested alternative, but should we perhaps go with a completely different PSP?

ibrokethecloud: We have a flag for selinux, may we need one for apparmor and a combination could drive the psp template?

This will also mean that user will have more control on how the behaviour of the psp.

brandond: Sounds like CIS doesn't care either way, so we're fine to just drop the annotations and leave the seLinux rule open.

brandond (approved):

cjellick (approved):

briandowns:
> LGTM. I'm curious if this will apply retroactively to existing RKE2 clusters that upgrade...

No.Those who upgrade will still have the annotations and will need to delete them manually. The next release will call this out specifically.

Oats87 (approved):

cjellick (approved):

Issue #597, opened 2020-12-14
Uninstall does not work for cent/rhel systems-regression


ShylajaDevadiga:
Environment:
rke2 -v
rke2 version v1.18.13-rc1+rke2r1

Centos7.8
RHEL8.2

Describe the issue:
rke2-uninstall.sh does not delete all files.

```
/etc/yum.repos.d/rancher-rke2.repo
/etc/sysctl.d/60-rke2-cis.conf
/etc/sysconfig/rke2-server
/etc/selinux/targeted/active/modules/400/rke2
/etc/rancher/rke2
/etc/rancher/rke2/rke2.yaml
/var/lib/yum/repos/x86_64/7/rancher-rke2-1.18-testing
/var/lib/yum/repos/x86_64/7/rancher-rke2-common-testing
/usr/bin/rke2
/usr/bin/rke2-killall.sh
/usr/bin/rke2-uninstall.sh
/usr/share/rke2
/usr/lib/systemd/system/rke2-server.service
```

davidnuzik:
RKE2 will still get killed. The problem is that expected files are not cleaned up; thus the node cannot be re-used. We will known-issue this in release notes for 1.18.13

davidnuzik:
@ShylajaDevadiga nothing has changed in rke2-uninstall.sh since october 6th. Are you certain this is a new regression? You ran the script as root?

ShylajaDevadiga:
Attaching log 
[uninstall.log](https://github.com/rancher/rke2/files/5690077/uninstall.log)

ShylajaDevadiga:
@davidnuzik Same issue is seen in rke2 version v1.18.12+rke2r2 as well.

davidnuzik:
Ok, not a regression. This will be marked as a known-issue in release notes.

davidnuzik:
Assigned at least initially to @briandowns for initial investigation.
Bumped milestone out of v1.18.13+rke2r1 (our release we are trying to get out today). Put in the v1.19.5+rke2r1 milestone, but it should get fixed as soon as possible post v1.18.13 release since our dev-complete date for v1.19.5 is today. If this needs another day past dev-complete day this should be fine since it's such short notice to get a PR in for this.
We should not merge into master unless this is trivial to do because we need to get the 1.18.13 release out asap...

davidnuzik:
The PR exists and is in review. I applied the Peer Review label.

ShylajaDevadiga:
[...]

PR #598, opened 2020-12-14
Update Uninstall Script

briandowns:

#### Linked Issues ####

#597

### DIFF
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -40,7 +40,7 @@ uninstall_remove_files()
 
         yum remove -y rke2-*
     fi
-    
+
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
@@ -48,8 +48,6 @@ uninstall_remove_files()
     rm -rf /etc/rancher/rke2
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
-
-
 }
 
 uninstall_remove_self()
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -34,7 +34,13 @@ uninstall_disable_services()
 
 uninstall_remove_files()
 {
+    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+        rm -f /etc/yum.repos.d/rancher-rke2*.repo
+        rm -f /etc/sysconfig/rke2-server
 
+        yum remove -y rke2-*
+    fi
+    
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
@@ -43,12 +49,7 @@ uninstall_remove_files()
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
 
-    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
-        rm -f /etc/yum.repos.d/rancher-rke2.repo
-        rm -f /etc/sysconfig/rke2-server
 
-        yum remove rke2-server
-    fi
 }
 
 uninstall_remove_self()
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -7,7 +7,7 @@ if [ ! $(id -u) -eq 0 ]; then
     exit 1
 fi
 
-if [ -f "/etc/redhat-release" ]; then
+if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
     : "${INSTALL_RKE2_ROOT:="/usr"}"
 else
     : "${INSTALL_RKE2_ROOT:="/usr/local"}"
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -7,7 +7,11 @@ if [ ! $(id -u) -eq 0 ]; then
     exit 1
 fi
 
-: "${INSTALL_RKE2_ROOT:="/usr/local"}"
+if [ -f "/etc/redhat-release" ]; then
+    : "${INSTALL_RKE2_ROOT:="/usr"}"
+else
+    : "${INSTALL_RKE2_ROOT:="/usr/local"}"
+fi
 
 uninstall_killall()
 {
@@ -30,6 +34,7 @@ uninstall_disable_services()
 
 uninstall_remove_files()
 {
+
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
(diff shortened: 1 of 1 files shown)


bundle/bin/rke2-uninstall.sh
```diff
@@ -30,13 +34,21 @@ uninstall_disable_services()
 
 uninstall_remove_files()
 {
+
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
```
Oats87, line 39: These are all managed by the RPM, I'm not sure we should be out-of-band removing files... If we are, then we should probably do the `yum -y remove rke2-*` before we actually go through and do the `rm -f ` of things

bundle/bin/rke2-uninstall.sh
```diff
@@ -30,13 +34,21 @@ uninstall_disable_services()
...
     rm -rf "${INSTALL_RKE2_ROOT}/share/rke2"
     rm -rf /etc/rancher/rke2
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
+
+    if [ -f "/etc/redhat-release" ]; then
+        rm -f /etc/yum.repos.d/rancher-rke2.repo
+        rm -f /etc/sysconfig/rke2-server
+
+        yum remove rke2-server
```
Oats87, line 50: We should probably make this a `yum remove -y rke2-*` as we have RPMs for both `rke2-server`, `rke2-agent`, and also coincidentally have the `rke2-selinux` floating around.

bundle/bin/rke2-uninstall.sh
```diff
@@ -30,13 +34,21 @@ uninstall_disable_services()
...
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
     rm -rf "${INSTALL_RKE2_ROOT}/share/rke2"
     rm -rf /etc/rancher/rke2
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
+
+    if [ -f "/etc/redhat-release" ]; then
+        rm -f /etc/yum.repos.d/rancher-rke2.repo
```
Oats87, line 47: This should probably be
```
rm -f /etc/yum.repos.d/rancher-rke2*.repo
```
as before, we used to create channel-specific repo files

bundle/bin/rke2-uninstall.sh
```diff
@@ -7,7 +7,11 @@ if [ ! $(id -u) -eq 0 ]; then
     exit 1
 fi
 
-: "${INSTALL_RKE2_ROOT:="/usr/local"}"
+if [ -f "/etc/redhat-release" ]; then
```
brandond, line 10: We should probably match https://github.com/rancher/rke2/blob/master/install.sh#L222
```suggestion
if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
```

bundle/bin/rke2-uninstall.sh
```diff
@@ -30,13 +34,21 @@ uninstall_disable_services()
...
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
     rm -rf "${INSTALL_RKE2_ROOT}/share/rke2"
     rm -rf /etc/rancher/rke2
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
+
+    if [ -f "/etc/redhat-release" ]; then
+        rm -f /etc/yum.repos.d/rancher-rke2.repo
+        rm -f /etc/sysconfig/rke2-server
```
Oats87, line 48: Do we have an expectation that `/etc/sysconfig/rke2-server` is around?

bundle/bin/rke2-uninstall.sh
Oats87: and `rke2-common`

bundle/bin/rke2-uninstall.sh
Oats87: (This is in reference to the detection up above of the `/usr` vs `/usr/local` for `INSTALL_RKE2_ROOT`)

Oats87 (changes requested):

bundle/bin/rke2-uninstall.sh
```diff
@@ -30,13 +34,21 @@ uninstall_disable_services()
...
+
     find "${INSTALL_RKE2_ROOT}/lib/systemd/system" -name rke2-*.service -type f -delete
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2"
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
     rm -rf "${INSTALL_RKE2_ROOT}/share/rke2"
     rm -rf /etc/rancher/rke2
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
+
+    if [ -f "/etc/redhat-release" ]; then
```
brandond, line 46:
```suggestion
    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
```

brandond (changes requested):

bundle/bin/rke2-uninstall.sh
dweomer: This has been cargo-culted (by me mostly) because it is the default env config pattern for systemd units on older redhat systems.

briandowns: I think the expectation is to have a "clean" system.

Oats87: Do we (RKE2) create the file though? I would imagine if we don't create it we shouldn't need to remove it

dweomer (approved):

brandond (approved):

bundle/bin/rke2-uninstall.sh
```diff
@@ -7,7 +7,11 @@ if [ ! $(id -u) -eq 0 ]; then
     exit 1
 fi
 
-: "${INSTALL_RKE2_ROOT:="/usr/local"}"
+if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+    : "${INSTALL_RKE2_ROOT:="/usr"}"
+else
+    : "${INSTALL_RKE2_ROOT:="/usr/local"}"
+fi
```
dweomer, line 14: P.S. This will fail when uninstalling a tarball installation on a redhat-derived system.

brandond: Should we perhaps just check for /usr/local/bin/rke2 as an indicator of a tarball-based install?

bundle/bin/rke2-uninstall.sh
```diff
@@ -30,6 +34,13 @@ uninstall_disable_services()
 
 uninstall_remove_files()
 {
+    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+        rm -f /etc/yum.repos.d/rancher-rke2*.repo
+        rm -f /etc/sysconfig/rke2-server
```
brandond, line 39: Will this file only exist on RPM-based systems?

bundle/bin/rke2-uninstall.sh
Oats87: Maybe we can update to that logic later on. For now I think this covers most cases (are people really installing RKE2 using tarball that often?)

Oats87 (approved):

bundle/bin/rke2-uninstall.sh
briandowns: Do we want this change now or later? What's the cncensus?

bundle/bin/rke2-uninstall.sh
briandowns: Looks like it. Not on Ubuntu.

Oats87: Yes, only on RPM-based systems as it's created by the RPM: https://github.com/rancher/rke2-packaging/blob/master/rpm/centos7/server/rke2-server.spec#L31

bundle/bin/rke2-uninstall.sh
dweomer: I am fine with changing this later

davidnuzik: I just don't want to lose track of it. Can we open a small gh issue to track? I guess set v1.19.5 milestone to fix this soon? Easy to fix later when we have some more time?
Let's just fix it later - but create gh issue to track.

Issue #538, opened 2020-11-18
Install script should support versioned channels


Oats87:
Today, the install.sh script only supports `latest`, `stable`, and `testing` as channels. It should be able to support channels that are like `v1.18`, `v1.19`, `v1.18-testing`, `v1.18-latest`, and so forth.

ShylajaDevadiga:
Verifying in v1.18.13-rc1+rke2r1 (3ff7cff12a1a74c991abe7628e0041e4659fbc67)

v1.18-latest is not an expected channel. v1.18 should point to latest version v1.18.12+rke2r2, but v1.18-latest does point to latest.
```
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_CHANNEL=v1.18 sh -
Rancher RKE2 Common (v1.18)                                                        375  B/s | 388  B     00:01    
Errors during downloading metadata for repository 'rancher-rke2-common-v1.18':
  - Status code: 404 for https://rpm.rancher.io/rke2/v1.18/common/centos/8/noarch/repodata/repomd.xml (IP: 104.24.112.22)
Error: Failed to download metadata for repo 'rancher-rke2-common-v1.18': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```
with Channel set to v1.18-latest 
```
rke2 -v
rke2 version v1.18.12+rke2r1 (b63f92a2b66e8ccfe2d3fae4814fa0e0579da5a4)
```

davidnuzik:
@ShylajaDevadiga we discussed this. There is no <version> channels yet. This fails when you set a valid channel?

davidnuzik:
@Oats87 moved to next milestone per our discussion. Still more work to do to support (TBD) versioned channels.
v1.19.5 has dev complete date of today. Please work on the issue as soon as you are able to; pr can get in but don't merge yet until post 1.18.13 release.

ShylajaDevadiga:
Thats correct @davidnuzik. As shared above v1.18 needs a fix. Channels stable, latest, testing works as expected.

atsai1220:
Having problem running the quick-start script but using **INSTALL_RKE2_VERSION** allows the script to pass.
Running this results in a 404 for the checksum file
```
curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL=latest sh -

[INFO]  finding release for channel latest
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
[...]

PR #605, opened 2020-12-14
Handle `vX.Y` channels in `install.sh`

Oats87:
In the case we have a `vX.Y` channel, default to pulling RPMs from the `stable` channel

#### Proposed Changes ####
When we find a `vX.Y` channel, assume we are pulling RPMs from `stable`

#### Types of Changes ####
`install.sh`

#### Verification ####
When the corresponding channels are available, try installing using them.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/538

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -238,6 +238,7 @@ do_install_rpm() {
             rke2_rpm_channel=$(echo "${INSTALL_RKE2_CHANNEL}" | sed -E -e "s/^v[0-9]+\.[0-9]+-(.*)/\1/")
             # If our regex fails to capture a "sane" channel out of the specified channel, fall back to `stable`
             if [ "${rke2_rpm_channel}" = ${INSTALL_RKE2_CHANNEL} ]; then
+                info "using stable RPM repositories"
                 rke2_rpm_channel="stable"
             fi
             ;;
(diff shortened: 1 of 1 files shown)


briandowns (approved):

install.sh
```diff
@@ -236,6 +236,10 @@ do_install_rpm() {
             # We are operating with a version-based channel, so we should parse our version out
             rke2_majmin=$(echo "${INSTALL_RKE2_CHANNEL}" | sed -E -e "s/^v([0-9]+\.[0-9]+).*/\1/")
             rke2_rpm_channel=$(echo "${INSTALL_RKE2_CHANNEL}" | sed -E -e "s/^v[0-9]+\.[0-9]+-(.*)/\1/")
+            # If our regex fails to capture a "sane" channel out of the specified channel, fall back to `stable`
+            if [ "${rke2_rpm_channel}" = ${INSTALL_RKE2_CHANNEL} ]; then
```
galal-hussein, line 240: can we log that when falling back to stable so that the user understands what happened

Oats87: @galal-hussein I just pushed a commit that adds the message if we fall back to stable

galal-hussein (approved):
LGTM

Issue #535, opened 2020-11-17
Using alternate path for data-dir fails some pods to come up


ShylajaDevadiga:
Version:
rke2 version v1.18.12-rc1+rke2r1

Environment:
Ubuntu 20.04

Describe the issue:
Using alternate path for data-dir fails some pods to come up due to missing files in expected location.

Steps to reproduce:
1. Create a single node cluster using alternate path for data-dir
2. Check the nodes are up
3. Check the logs of kube-proxy
```
/mnt/bin/kubectl logs -n kube-system  kube-proxy-w65hn
F1116 23:55:11.550968       1 server.go:497] stat /var/lib/rancher/rke2/agent/kubeproxy.kubeconfig: no such file or directory
```

Additional info:
Pods are in ContainerCreating state
```
/mnt/bin/kubectl get pods -A
NAMESPACE     NAME                                         READY   STATUS              RESTARTS   AGE
kube-system   etcd-ip-172-31-3-76                          1/1     Running             0          106m
kube-system   helm-install-rke2-canal-r88z4                0/1     Completed           0          107m
kube-system   helm-install-rke2-coredns-w6j62              0/1     Completed           0          107m
kube-system   helm-install-rke2-ingress-nginx-lvwfd        0/1     ContainerCreating   0          107m
kube-system   helm-install-rke2-kube-proxy-q4bcp           0/1     Completed           0          107m
kube-system   helm-install-rke2-metrics-server-f7fjr       0/1     ContainerCreating   0          107m
kube-system   kube-apiserver-ip-172-31-3-76                1/1     Running             0          106m
kube-system   kube-controller-manager-ip-172-31-3-76       1/1     Running             0          107m
kube-system   kube-proxy-n5zfs                             0/1     CrashLoopBackOff    2          33s
kube-system   kube-scheduler-ip-172-31-3-76                1/1     Running             0          107m
kube-system   rke2-canal-bm6t8                             0/2     CrashLoopBackOff    54         106m
kube-system   rke2-coredns-rke2-coredns-649896d599-5pnxl   0/1     ContainerCreating   0          106m
```

brandond:
This ended up being intertwined with the other data-dir fix (cluster wouldn't come up without it) so I've pulled it in to the next milestone.

ShylajaDevadiga:
rke2 version v1.19.5-rc1+rke2r1
On CentOS, Cluster fails to come up while using alternate data-dir

```
[...]

Issue #474, opened 2020-10-06
Using alternate location data-dir crictl.yaml still resides in the default data-dir path


ShylajaDevadiga:
Version:
rke2 version v1.18.9+rke2r1

Describe the Issue:

While using /mnt dir as the param for data-dir all content of data dir is relocated to new /mnt dir, except for crictl file in the agent dir

```
# ls -R  /var/lib/rancher/rke2/
/var/lib/rancher/rke2/:
agent

/var/lib/rancher/rke2/agent:
etc

/var/lib/rancher/rke2/agent/etc:
crictl.yaml

# ls /mnt
agent  bin  data  server
```

Cluster is up
```
/mnt/bin/kubectl get nodes
NAME              STATUS   ROLES         AGE     VERSION
ip-172-31-23-79   Ready    <none>        2m14s   v1.18.9+rke2r1
ip-172-31-30-69   Ready    etcd,master   41m     v1.18.9+rke2r1
```

brandond:
This appears to be coming from here:

rancher/k3s@58aae57#diff-e86b77690979bd28019f01c993abec39R44

ShylajaDevadiga:
While validating on k3s version v1.19.3+k3s-58b5b21f, found the behavior is different while passing data-dir and passing data-dir through config file
Node OS: Ubuntu 20.04, sudo user
Using data-dir flag in cli, alternate location is used and /var/lib/rancher does not exist
``` 
ubuntu@ip-172-31-34-47:~$ ls /var/lib/rancher
ls: cannot access '/var/lib/rancher': No such file or directory
ubuntu@ip-172-31-34-47:~$ ls /mnt
agent                      client-kube-proxy.crt  containerd                kubelet.kubeconfig    server-ca.crt
client-ca.crt              client-kube-proxy.key  data                      kubeproxy.kubeconfig  serving-kubelet.crt
client-k3s-controller.crt  client-kubelet.crt     etc                       pod-manifests         serving-kubelet.key
client-k3s-controller.key  client-kubelet.key     k3scontroller.kubeconfig  server
```
Using config file, /var/lib/rancher contains k3s/data dir
```
cat /etc/rancher/k3s/config.yaml 
data-dir: /mnt
 ls /var/lib/rancher/k3s/data/
76b839842c85db78d477b73c3dcb656159d31d91081517102614dad69c3a340c  current
ls /mnt
[...]

PR #622, opened 2020-12-17
Fix data-dir handling for crictl and helm charts

brandond:
#### Proposed Changes ####

* Bump K3s version to pull through crictl.yaml fix from k3s-io/k3s#2475
* Update rancher/helm-controller to k3s-io/helm-controller to track with k3s (and update a couple docs to track the move as well)
* Pass through data-dir to helm charts as global.rke2DataDir so that templates can update host paths accordingly

#### Types of Changes ####

* K3s
* Helm chart bootstrap

#### Verification ####

* Start RKE2 with --data-dir in CLI flags or config file
* Note that /var/lib/rancher/rke2 does not exist
* Note that all pods start up successfully

#### Linked Issues ####

#474
#535
rancher/rke2-charts#36

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -18,9 +18,9 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	"github.com/k3s-io/helm-controller/pkg/helm"
 	errors2 "github.com/pkg/errors"
-	helmv1 "github.com/rancher/helm-controller/pkg/apis/helm.cattle.io/v1"
-	"github.com/rancher/helm-controller/pkg/helm"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/rancher/wrangler/pkg/schemes"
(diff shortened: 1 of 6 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -370,10 +371,16 @@ func rewriteChart(fileName string, info os.FileInfo, systemDefaultRegistry strin
 
 	// Generally we should avoid using Set on HelmCharts since it cannot be overridden by HelmChartConfig,
 	// but in this case we need to do it in order to avoid potentially mangling the ValuesContent field by
-	// blindly appending content to it in order to set the global.systemDefaultRegistry value.
+	// blindly appending content to it in order to set values.
 	if chart.Spec.Set == nil {
 		chart.Spec.Set = map[string]intstr.IntOrString{}
 	}
+
+	if chart.Spec.Set["global.rke2DataDir"].StrVal != dataDir {
```
briandowns, line 379: What's the behavior here if the objects at that key doesn't exist and we try to call `Strval` on it?

pkg/bootstrap/bootstrap.go
```diff
@@ -336,17 +337,17 @@ func setSystemDefaultRegistry(manifestsDir string, systemDefaultRegistry string)
...
 	}
 	return merr.NewErrors(errs...)
 }
 
-// rewriteChart applies systemDefaultRegistry settings to the file at fileName with associated info.
+// rewriteChart applies dataDir and systemDefaultRegistry settings to the file at fileName with associated info.
 // If the file cannot be decoded as a HelmChart, it is silently skipped. Any other IO error is considered
 // a failure.
-func rewriteChart(fileName string, info os.FileInfo, systemDefaultRegistry string, serializer *json.Serializer) error {
+func rewriteChart(fileName string, info os.FileInfo, dataDir string, systemDefaultRegistry string, serializer *json.Serializer) error {
```
briandowns, line 350:
```suggestion
func rewriteChart(fileName string, info os.FileInfo, dataDir, systemDefaultRegistry string, serializer *json.Serializer) error {
```

pkg/bootstrap/bootstrap.go
brandond: `chart.Spec.Set` is typed as `map[string]intstr.IntOrString{}`, so if that's empty the map gives you back a default IntOrString struct. StrVal on the default struct is just an empty string.

briandowns: Cool. Just wanted to make sure we weren't introducing a point of panic.

briandowns (approved):

galal-hussein (approved):
LGTM

Issue #471, opened 2020-10-06
tarball uninstall.sh doesnt remove /var/lib/rancher


cjellick:
it removes everythign under that directory, but not that directory

it also leaves behind /etc/rancher/...something something/none..something. is that intnetional?

rancher-max:
https://github.com/rancher/rke2/pull/626 was merged to address this, but on a freshly installed cluster on an Ubuntu 20.04 system, this is not yet resolved as I see the `/var/lib/rancher` directory still present even though I didn't have any other directories or files in it.

davidnuzik:
FYI @innobead @c3y1huang we need to follow proper process with GitHub. Please reach out to me on Slack if you have questions. Assigning this back to @c3y1huang and reopening.

brandond:
I have been unable to reproduce this with tarball or rpm install on EL7 or tarball on Ubuntu, using either the stable or testing channels.

```
+ uninstall_remove_files
+ [ -r /etc/redhat-release ]
+ [ -r /etc/centos-release ]
+ [ -r /etc/oracle-release ]
+ find /usr/local/lib/systemd/system -name rke2-*.service -type f -delete
+ rm -f /usr/local/bin/rke2
+ rm -f /usr/local/bin/rke2-killall.sh
+ rm -rf /usr/local/share/rke2
+ rm -rf /etc/rancher/rke2
+ rm -rf /etc/rancher/node
+ rm -d /etc/rancher
+ rm -rf /var/lib/kubelet
+ rm -rf /var/lib/rancher/rke2
+ rm -d /var/lib/rancher
+ uninstall_remove_self
+ rm -f /usr/local/bin/rke2-uninstall.sh


root@ubuntu01:~# find /var/lib/rancher
find: ‘/var/lib/rancher’: No such file or directory
```

rancher-max:
I can no longer recreate this either, so looks like it's working as intended now

PR #626, opened 2020-12-18
Uninstall remove relate directories if is empty

c3y1huang:
#### Proposed Changes ####

Run rke2-uninstall.sh also remove /etc/rancher/node and try to remove /var/lib/rancher and /etc/rancher. It should skip if the directory is not empty.

#### Types of Changes ####

Uninstall, Enhancement

#### Verification ####

* On a new deployed cluster, rke2-uninstall.sh should also remove /var/lib/rancher.
* Create a new file/directory in /var/lib/rancher, rke2-uninstall.sh should not remove /var/lib/rancher.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/471

#### Further Comments ####

`none`

### DIFF
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -45,8 +45,11 @@ uninstall_remove_files()
     rm -f "${INSTALL_RKE2_ROOT}/bin/rke2-killall.sh"
     rm -rf "${INSTALL_RKE2_ROOT}/share/rke2"
     rm -rf /etc/rancher/rke2
+    rm -rf /etc/rancher/node
+    rm -d /etc/rancher || true
     rm -rf /var/lib/kubelet
     rm -rf /var/lib/rancher/rke2
+    rm -d /var/lib/rancher || true
 }
 
 uninstall_remove_self()


bundle/bin/rke2-uninstall.sh
```diff
@@ -47,6 +47,16 @@ uninstall_remove_files()
...
+    remove_directory_if_empty "/var/lib/rancher"
+}
+
+remove_directory_if_empty()
+{
+    if [ -z "$(ls -A ${1})" ]; then
+        rmdir ${1}
+    else
+        echo "Skip remove directory: ${1} is not empty."
+    fi
```
dweomer, line 59:
```suggestion
    rm -d /var/lib/rancher
```

dweomer: Also, if we are doing this, should we not also do the same for `/etc/rancher` ?

briandowns: We may need to check with @ShylajaDevadiga and @rancher-max in case this breaks any automation on their side.

rancher-max: Our automation brings up new instances each time so this should be okay for us! Thank you for checking

brandond (approved):

bundle/bin/rke2-uninstall.sh
c3y1huang: Thanks @dweomer ! I wasn't sure if the /etc/rancher/node/password was kept on purpose. Addressed.

I'm a little concerned about using `rm -d`. When used, the function will exit on the first fail. That means if `/etc/rancher` is not empty and `/var/lib/rancher` is empty. Both directories will be kept, and all the following commends in that function won't be processed.

brandond: you can `|| true` or just `set +e` beforehand and then `set -e` back again afterwards.

c3y1huang: @brandond thanks!

briandowns: > Our automation brings up new instances each time so this should be okay for us! Thank you for checking

Cool. Just wanted to make sure since the expectation was that this is removed and this changes it. I know @ShylajaDevadiga had a list of files and directories to check for in the past.

brandond (approved):

briandowns (approved):

PR #627, opened 2020-12-18
Update automated_upgrade.md

atsai1220:
We ran into issues copying this example. The controller would not apply the Plan because we did not have `nodeSelector` as part of our Plan's manifest and took us a while to catch this.

#### Proposed Changes ####

Example needs to be updated to avoid first-time user error.

#### Types of Changes ####

Documentation clarification.

#### Verification ####

Looking at the previous version you can see that `nodeSelector` is not part of documentation's example.

### DIFF
--- a/docs/upgrade/automated_upgrade.md
+++ b/docs/upgrade/automated_upgrade.md
@@ -47,7 +47,8 @@ metadata:
 spec:
   concurrency: 1
   cordon: true
-   matchExpressions:
+  nodeSelector:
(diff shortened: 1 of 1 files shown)


cjellick (approved):

briandowns (approved):

brandond (approved):

c3y1huang (approved):

atsai1220:
Thank you! :D

PR #684, opened 2021-02-02
docs: tweak some of the verbiage and link on fips page

dweomer:
- Remove link out to goboring on Docker Hub which we do not produce nor use.
- Remove possibly confusing reference to BoringSSL

### DIFF
--- a/docs/security/fips_support.md
+++ b/docs/security/fips_support.md
@@ -6,13 +6,13 @@ FIPS 140-2 is a U.S. Federal Government security standard used to approve crypto
 
 ## Use of FIPS Compatible Go compiler.
 
-The Go compiler in use can be found [here](https://hub.docker.com/u/goboring). Each component of the system is built with the version of this compiler that matches the same standard Go compiler version that would be used otherwise.
+The Go compiler in use can be found [here](https://go.googlesource.com/go/+/dev.boringcrypto). Each component of the system is built with the version of this compiler that matches the same standard Go compiler version that would be used otherwise.
 
 This version of Go replaces the standard Go crypto libraries with the FIPS validated BoringCrypto module. See GoBoring's [readme](https://github.com/golang/go/blob/dev.boringcrypto/README.boringcrypto.md) for more details. This module has been revalidated as the [Rancher Kubernetes Cryptographic Library](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/3836) in order to ensure support on a wider range of systems.
 
 ### FIPS Support in Cluster Components
 
-Most of the components of the RKE2 system are statically compiled with the GoBoring Go compiler implementation that takes advantage of the BoringSSL library. RKE2, from a component perspective, is broken up in a number of sections. The list below contains the sections and associated components.
+Most of the components of the RKE2 system are statically compiled with the GoBoring Go compiler implementation. RKE2, from a component perspective, is broken up in a number of sections. The list below contains the sections and associated components.
 
 * Kubernetes
   * API Server


docs/security/fips_support.md
```diff
@@ -6,15 +6,15 @@ FIPS 140-2 is a U.S. Federal Government security standard used to approve crypto
 
 ## Use of FIPS Compatible Go compiler.
 
-The Go compiler in use can be found [here](https://hub.docker.com/u/goboring). Each component of the system is built with the version of this compiler that matches the same standard Go compiler version that would be used otherwise.
```
dweomer, line 9: We do not use this container nor do we produce it.

docs/security/fips_support.md
```diff
@@ -6,15 +6,15 @@ FIPS 140-2 is a U.S. Federal Government security standard used to approve crypto
...
 
 This version of Go replaces the standard Go crypto libraries with the FIPS validated BoringCrypto module. See the [readme](https://github.com/golang/go/blob/dev.boringcrypto/README.boringcrypto.md) for more details.
 
 Moreover, this module is currently being [revalidated](../assets/fips_engagement.pdf) as the Rancher Kubernetes Cryptographic Library for the additional platforms and systems supported by RKE2.
 
 ### FIPS Support in Cluster Components
 
-Most of the components of the RKE2 system are statically compiled with the GoBoring Go compiler implementation that takes advantage of the BoringSSL library. RKE2, from a component perspective, is broken up in a number of sections. The list below contains the sections and associated components.
+Most of the components of the RKE2 system are statically compiled with the GoBoring Go compiler implementation. RKE2, from a component perspective, is broken up in a number of sections. The list below contains the sections and associated components.
```
dweomer, line 17: This deletion clears up an inconsistency with "This version of Go replaces the standard Go crypto libraries with the FIPS validated BoringCrypto module." sentence two paragraphs earlier.

dweomer: The reasons that this is significant is that BoringSSL also uses the BoringCrypto module but BoringSSL is not distributed as part of RKE2 and RKE2 is not build with BoringSSL.

docs/security/fips_support.md
dweomer: @briandowns has reminded me that we did at one time use this base container for building.

docs/security/fips_support.md
```diff
@@ -6,15 +6,15 @@ FIPS 140-2 is a U.S. Federal Government security standard used to approve crypto
 
 ## Use of FIPS Compatible Go compiler.
 
-The Go compiler in use can be found [here](https://hub.docker.com/u/goboring). Each component of the system is built with the version of this compiler that matches the same standard Go compiler version that would be used otherwise.
+The Go compiler in use can be found [here](https://go.googlesource.com/go/+/dev.boringcrypto). Each component of the system is built with the version of this compiler that matches the same standard Go compiler version that would be used otherwise.
```
brandond, line 9: You might link to the branch readme: https://github.com/golang/go/blob/dev.boringcrypto/README.boringcrypto.md

docs/security/fips_support.md
brandond, line 17: > Most of the components of the RKE2 system are statically compiled with the GoBoring Go compiler implementation

Here's a list of things, some of them are built with GoBoring. Maybe we should call out which ones are?

brandond (changes requested):

docs/security/fips_support.md
dweomer: It is linked to in the next paragraph.

docs/security/fips_support.md
dweomer: This is good feedback and something the docs could benefit from but this changes was to just fix the link.

Issue #710, opened 2021-02-19
How to configure authenticated registry access to fetch RKE2 images "secure" from on-premise registry


Martin-Weiss:
It seems we have just "anonymous" access as option for fetching images from the on-premise registry in the RKE2 deployment process.

Is there any way to give RKE2 credentials to fetch the images non-anonymous from an on-premise registry?

Background: for security and auditing reasons "anonymous" access is not allowed.

P.S. yes - I know about the tarball image solution - but that is not a nice way to distribute images to new and existing servers and image transport standard is to use a registry ;-).

brandond:
cc @cjellick

brandond:
In case it wasn't clear from the status updates, this is being actively worked on.

Martin-Weiss:
> 
> 
> In case it wasn't clear from the status updates, this is being actively worked on.

Cool- that is great news!

Btw. I have seen this message during rke-server start:

`rancher-master-01 rke2[5844]: No matching credentials were found, falling back on anonymous`

so is this already implemented "somehow"?

brandond:
Yes, right now it will read creds from root's docker credentials at /root/.docker/config.json, but only when pulling the runtime image. This will be unified to pull creds from registries.yaml.

Martin-Weiss:
Ok - so as a temporary workaround I just need to create that config.json?
I do not have docker installed - but creating that file is not much of a challenge ;-)

rancher-max:
#### Validated using v1.20.4-rc2+rke2r1

- There is now the ability to use mirrors to fetch images from secure registry. By default, this is at `/etc/rancher/rke2/registries.yaml`. See below example registries file with dummy data:
```yaml
---
mirrors:
  docker.io:
    endpoint:
      - "my.private.registry"
  k8s.gcr.io:
    endpoint:
      - "my.private.registry"
configs:
  "my.private.registry":
    auth:
      username: fakeuser
      password: fakepass
    tls:
      cert_file: /home/ubuntu/cert.pem
      key_file:  /home/ubuntu/key.pem
      ca_file: /home/ubuntu/ca.pem
```
- Can store this file anywhere, and pass location when starting rke2 such as: `rke2 server --private-registry=/this/is/my/test/folder/auth.yaml` 
- If desired, can use this in conjunction with tarball and/or `--system-default-registry` flag (for anonymous access to same or different registry).

Issue #683, opened 2021-02-02
Make system-default-registry work with airgap images


brandond:
It is time to revisit the comments from #233:

> We're currently pulling images via go-containerregistry, we might need to hack something in to the transport if it doesn't have any direct mirror/endpoint override support.

> One path is to design and implement of registry mirroring inside of go-containerregistry, whether in a fork or directly upstream. Seeing that the moby PR has been open for 3 years with SUSE backing, don't think this can be assumed to be easy. The other path is to switch to a beefier registry client that does support mirroring, and importing the registries.yaml into this client's mirror config. I don't think go-containerregistry's remote.WithTransport() is a possible solution

This is in support of hybrid airgap image preload + local registry mirroring. Right now system-default-registry disables loading of images from tarballs, which is still desired in some situations (such as an in-cluster mirror).

rancher-max:
#### Validated in v1.20.4-rc2+rke2r1

- Can use system-default-registry and tarball together. Can also use the new registries.yaml mirror with all of these -- any combination of these three options works well.
- When using these two options together, images will appear to be pulled from the value from `--system-default-registry` (even if it is a completely invalid registry), as they are retagged:
```
# Log example from startup
INFO[0168] Imported docker.io/rancher/hardened-calico:v3.13.3-build20210223 
INFO[0168] Tagged myfake.registry.max/rancher/hardened-calico:v3.13.3-build20210223 

# Images on the host:
$ sudo /var/lib/rancher/rke2/bin/crictl -r unix:///run/k3s/containerd/containerd.sock images
IMAGE                                                                 TAG                              IMAGE ID            SIZE
docker.io/rancher/hardened-calico                                     v3.13.3-build20210223            736cae9d947ba       437MB
myfake.registry.max/rancher/hardened-calico                           v3.13.3-build20210223            736cae9d947ba       437MB
docker.io/rancher/hardened-coredns                                    v1.6.9-build20210223             6ab71f04f5474       128MB
myfake.registry.max/rancher/hardened-coredns                          v1.6.9-build20210223             6ab71f04f5474       128MB
docker.io/rancher/hardened-etcd                                       v3.4.13-k3s1-build20210223       271c0a695260e       126MB
myfake.registry.max/rancher/hardened-etcd                             v3.4.13-k3s1-build20210223       271c0a695260e       126MB
docker.io/rancher/hardened-flannel                                    v0.13.0-rancher1-build20210223   366c64051af85       261MB
[...]

Issue #676, opened 2021-01-27
RKE2 with SLES15SP2 - cannot download images from nexus based repo with authentification


peter-englmaier:
I am trying to get RKE2 v1.19.7+rke2r1 to work with SLES15SP2. I am trying to download images from a repository with authorization, which works with another SLES15SP2 host with docker, but with kubectl on the rke2 cluster, I get the following error:
```
"Failed to pull image "private.repo.ch:443/software/node:1.0": rpc error: code = NotFound desc = failed to pull and unpack image "private.repo.ch:443/software/node:1.0": failed to resolve reference "private.repo.ch:443/software/node:1.0":  private.repo.ch:443/software/node:1.0: not found
```
When I try to download the image using crictl pull --creds user:password private.repo.ch:443/software/node:1.0
I am getting the same error regardless of which creds I give, even without creds. It looks like it does not send the credentials. 

I can, however, side load the images using "ctn image import" and then it can be used in the cluster.

SLES15SP2 is not yet listed as validated, but it should work on any linux.

Related to #592 and maybe #647 (I did not specify a different default repository).

Best, Peter

peter-englmaier:
Tested with RKE2 v1.19.8+rke2r1 again. Does not work. 

```
ctr --debug image pull --user user:password private.repo.ch:443/software/node:1.0
DEBU[0000] fetching                                      image=" private.repo.ch:443/software/node:1.0"
DEBU[0000] resolving                                     host="private.repo.ch:443"
DEBU[0000] do request                                    host="private.repo.ch:443" request.header.accept="application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*" request.header.user-agent=containerd/v1.4.3-k3s3 request.method=HEAD url="https://private.repo.ch:443/v2/software/node/manifests/1.0"
DEBU[0000] fetch response received                       host="private.repo.ch:443" response.header.cache-control="must-revalidate,no-cache,no-store" response.header.connection=keep-alive response.header.content-length=485 response.header.content-type="text/html;charset=iso-8859-1" response.header.date="Fri, 26 Feb 2021 11:08:41 GMT" response.header.keep-alive="timeout=5" response.header.server=nginx response.header.strict-transport-security="max-age=7776000" response.status="404 Not Found" url="https://private.repo.ch:443/v2/software/node/manifests/1.0"
[...]

Issue #641, opened 2020-12-24
Feature Request: RKE2 Support for Mixed Mode Registry


rajivml:
HI,

w.r.t Airgapped installs, 

If Docker private registry is installed on the same RKE2 cluster where all your target applications are deployed, then the same registry cannot be configured via system-default-registry param and as a result we have to patch our helm templates at so many places using kustomize to make helm/kubectl aware of docker private registry
       
So, what we are trying to achieve here is we want to use default registry location (i.e /var/lib/rancher/rke2/agent/images) for bootstrapping the RKE2 cluster and once the cluster is up we want to setup docker private registry over there and use the same for rest of the helm/kubectl installs.

But if we set system-default-registry to the docker private registry IP, RKE2 will no longer look at the path "/var/lib/rancher/rke2/agent/images" where the images required to bootstrap the k8 cluster are located and as a result cluster can't boot up

Where as our expectation is RKE2 should look at both "/var/lib/rancher/rke2/agent/images" location as well as the registryIP that we have configured via system-default-registry property, so that we don't have to patch any of our helm-charts and it will make the life of the customers much easy.

In the absence of this support, customers has to resort to kustomize or use some other open source library like this, but the problem using these kind of open source plugins is lack of support and these plugins can get deprecated any time due to lack of committers 
https://github.com/phenixblue/imageswap-webhook

rajivml:
More context about the request can be found here: https://rancher-users.slack.com/archives/C3ASABBD1/p1608958174187100

rajivml:
I think it makes more sense to support this at the platform level , from last one week am just doing this one thing of patching our helm templates in different ways so that RKE2 is aware of those images.

If RKE2 supports this out of box i.e. look for images at  "/var/lib/rancher/rke2/agent/images" first followed by looking for images at the system-default-registry-ip configured before bailing out, then it will solve the issue for those customers who want to setup docker private registry on the same original cluster where all original workloads are deployed

rajivml:
any chance of this getting picked up ? if there are doubts, please let me know , I can share further details

brandond:
I believe #715 should address your concerns. You will still need to be careful with the manifests needed to bootstrap the in-cluster registry to ensure that the imagePullPolicy is set to IfNotPresent, but once that is merged you will be able to use an airgap tarball at the same time as --system-default-registry. Note that you will still probably want to use registries.yaml to redirect docker.io to your private registry to avoid having to update the manifests for everything else you're deploying.

[...]

PR #715, opened 2021-02-20
Improve boostrap compat with airgap + private registry

brandond:
#### Proposed Changes ####

One commit per change:

*   Rewrite image override handling
    >Use go-containerregistry image reference packages to handle overriding   image names and registries instead of just building strings at startup.

* Remove quoted formatting in error messages
    >Quotes in log messages are awkward because they end up backslash-escaped    and it's hard to read. They aren't really adding any value so get rid of    them.

* Improve performance of bootstrap image extraction
    >The boostrap image was getting scanned twice - once to extract the    binaries, and again to extract the charts. The content is pulled or read    from the tarball on-demand so this is somewhat expensive.
    >
    >Improve performance by caching content locally, and only iterating    across it once.

* Restore charts content on startup
    >Chart manifests are actively deleted from the filesystem when they are    disabled, so we need to restore them from the copy extracted from the    runtime image every time we start up, otherwise they might unexpectedly    reappear after an upgrade when they were previously removed from the    disable list.

* Retag airgap tarball content to appear to come from system-default-registry
    >If using system-default-registry, all the manifests are updated to    specify that the image is sourced from this registry, but the images in    the tarball still come from docker.io.
    >
    >Fix this by retagging all the images in the tarball to also appear to    come from the default registry. K3s actually does the work, we just have    to ask it to do so.

* Add support for loading bootstrap image from compressed tarballs
    >K3s supports loading images from compressed tarballs, but we hadn't    added support for this when loading the bootstrap image. This meant that    users needed to manually decompress the extremely-large airgap image    tarball after downloading it from github.
    >
    >Fix this by adding support for the same compression formats that K3s    support when importing images, and also providing a zstd-compressed    tarball that shaves another ~250MB off the gzip image.

* Use registries.yaml auth/endpoint/tls configuration for bootstrap image pull
    >The boostrap image pull always went directly to the configured registry,    and did not support http registries, authentication, or other advanced    functionality.
    >
    >Add support for loading both credentials and transport (tls/endpoint)    configuration from registries.yaml.
    >
    >NOTE: Bootstrap image pulls will only use the first configured endpoint    for a registry, due to limitations in the go-containerregistry code. I    don't think this should be much of an issue, as I have not seen anyone    actually using multiple endpoints.

#### Types of Changes ####

* Airgap
* Packaged Components

#### Verification ####

Start RKE2 using all possible combinations of:
* Airgap tarball
* Private registry via registries.yaml
* --system-default-registry set to same / different registry as endpoint in registries.yaml

Start RKE2 with --disable=coredns, then restart with no disabled components and confirm that coredns comes back.

#### Linked Issues ####

#641
#676
#683
#710
#729

#### Further Comments ####

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -2,6 +2,8 @@ package bootstrap
 
 import (
 	"archive/tar"
+	"compress/bzip2"
+	"compress/gzip"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
@@ -21,7 +23,11 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
 	"github.com/k3s-io/helm-controller/pkg/helm"
+	"github.com/klauspost/compress/zstd"
+	"github.com/pierrec/lz4"
 	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/agent/util"
+	"github.com/rancher/k3s/pkg/untar"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/rancher/wrangler/pkg/schemes"
@@ -32,7 +38,10 @@ import (
 
 var releasePattern = regexp.MustCompile("^v[0-9]")
 
-const bufferSize = 4096
+const (
+	bufferSize    = 4096
+	extensionList = ".tar .tar.lz4 .tar.bz2 .tbz .tar.gz .tgz .tar.zst .tzst" // keep this in sync with the decompressor list
+)
 
 // binDirForDigest returns the path to dataDir/data/refDigest/bin.
 func binDirForDigest(dataDir string, refDigest string) string {
@@ -367,7 +376,7 @@ func preloadBootstrapImage(dataDir string, imageRef name.Reference) (v1.Image, e
 
 	// Try to find the requested tag in each file, moving on to the next if there's an error
 	for fileName := range files {
-		img, err := tarball.ImageFromPath(fileName, &imageTag)
+		img, err := preloadFile(imageTag, fileName)
 		if img != nil {
 			logrus.Debugf("Found %s in %s", imageTag, fileName)
 			return img, nil
@@ -378,6 +387,70 @@ func preloadBootstrapImage(dataDir string, imageRef name.Reference) (v1.Image, e
 	}
 	logrus.Debugf("No local image available for %s: not found in any file in %s", imageTag, imagesDir)
 	return nil, nil
+}
+
+// preloadFile handles loading images from a single tarball.
+func preloadFile(imageTag name.Tag, fileName string) (v1.Image, error) {
+	var opener tarball.Opener
+	switch {
+	case util.HasSuffixI(fileName, ".txt"):
+		return nil, nil
+	case util.HasSuffixI(fileName, ".tar"):
+		opener = func() (io.ReadCloser, error) {
+			return os.Open(fileName)
+		}
+	case util.HasSuffixI(fileName, ".tar.lz4"):
+		opener = func() (io.ReadCloser, error) {
+			file, err := os.Open(fileName)
+			if err != nil {
+				return nil, err
+			}
+			zr := lz4.NewReader(file)
+			return SplitReadCloser(zr, file), nil
+		}
+	case util.HasSuffixI(fileName, ".tar.bz2", ".tbz"):
+		opener = func() (io.ReadCloser, error) {
+			file, err := os.Open(fileName)
+			if err != nil {
+				return nil, err
+			}
+			zr := bzip2.NewReader(file)
+			return SplitReadCloser(zr, file), nil
+		}
+	case util.HasSuffixI(fileName, ".tar.gz", ".tgz"):
+		opener = func() (io.ReadCloser, error) {
+			file, err := os.Open(fileName)
+			if err != nil {
+				return nil, err
+			}
+			zr, err := gzip.NewReader(file)
+			if err != nil {
+				return nil, err
+			}
+			return MultiReadCloser(zr, file), nil
+		}
+	case util.HasSuffixI(fileName, "tar.zst", ".tzst"):
+		opener = func() (io.ReadCloser, error) {
+			file, err := os.Open(fileName)
+			if err != nil {
+				return nil, err
+			}
+			zr, err := zstd.NewReader(file, zstd.WithDecoderMaxMemory(untar.MaxDecoderMemory))
+			if err != nil {
+				return nil, err
+			}
+			return ZstdReadCloser(zr, file), nil
+		}
+	default:
+		return nil, fmt.Errorf("unhandled file type; supported extensions: " + extensionList)
+	}
+
+	img, err := tarball.Image(opener, &imageTag)
+	if err != nil {
+		logrus.Debugf("Did not find %s in %s: %s", imageTag, fileName, err)
+		return nil, nil
+	}
+	return img, nil
 }
 
 // setChartValues scans the directory at manifestDir. It attempts to load all manifests
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -62,17 +62,18 @@ func dirExists(dir string) bool {
 	return false
 }
 
-// Stage extracts binaries and manifests from the runtime image specified in imageConf into the directory
-// at dataDir. It attempts to load the runtime image from a tarball at dataDir/agent/images,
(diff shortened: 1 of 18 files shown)


pkg/bootstrap/registries.go
```diff
@@ -0,0 +1,175 @@
...
+	req.Host = endpointURL.Host
+	req.URL.Host = endpointURL.Host
+	req.URL.Scheme = endpointURL.Scheme
+
+	tlsConfig, err := r.getTLSConfigForEndpoint(endpointURL)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get TLS config from private registry configuration")
+	}
+
+	transport := &http.Transport{
```
ibuildthecloud, line 73: I think you need to CloseIdleConnections in defer, else disable the idle connections.

brandond: This is just the default Transport from https://github.com/golang/go/blob/release-branch.go1.15/src/net/http/transport.go#L42. All I'm really doing beyond what you'd get from http.DefaultTransport is setting TLSClientConfig.

Since I'm not actually making the request, but rather letting the core code handle the actual round-tripping, I'm not responsible for closing anything, am I?

brandond: Oh I see, I'm creating a new transport per request which means that idle connections can hang out on that Transport even though I'm never going to reuse it. I should either disable idle connections, or make sure that I return the same transport instance for the same endpoint.

pkg/bootstrap/bootstrap.go
```diff
@@ -137,10 +156,48 @@ func Stage(dataDir string, imageConf images.Images) (string, error) {
 	return refBinDir, nil
 }
 
-// extract extracts image content to targetDir all content from reader where the filename is prefixed with prefix.
-// The imageName argument is used solely for logging.
-// The destination directory is expected to be nonexistent or empty.
-func extract(imageName string, targetDir string, prefix string, reader io.Reader) error {
+// copyAll recursively copies files from source to dest, creating directories
```
dweomer, line 159: So long as we are not mutating while we copy, aka require visitor-pattern logic, then we should consider leverage https://pkg.go.dev/github.com/containerd/continuity@v0.0.0-20210208174643-50096c924a4e/fs#CopyDir which should already be in scope.

pkg/bootstrap/bootstrap.go
```diff
@@ -220,93 +278,210 @@ func extractToDir(dir, prefix string, img v1.Image, imageName string) error {
...
 	return merr.NewErrors(errs...)
 }
 
-// preloadBootstrapImage attempts return an image named imageName from a tarball
+// preloadBootstrapFromRuntime tries to load the runtime image from tarballs, using both the
+// default registry, and the user-configured registry (on the off chance they've retagged the
+// images in the tarball to match their private registry).
+func preloadBootstrapFromRuntime(dataDir string, resolver *images.Resolver) (v1.Image, error) {
+	var refs []name.Reference
+	runtimeRef := resolver.MustGetReference(images.Runtime)
```
dweomer, line 351: The return signature of `preloadBootstrapFromRuntime` includes an error so we should return it instead of delegating to `logrus.Fatalf` in the `MustGetReference` wrapper.

pkg/bootstrap/bootstrap.go
```diff
@@ -220,93 +278,210 @@ func extractToDir(dir, prefix string, img v1.Image, imageName string) error {
...
+	if runtimeRef.Context().Registry.Name() == images.DefaultRegistry {
+		// If the image is from the default registry, only check for that.
+		refs = []name.Reference{runtimeRef}
+	} else {
+		// If the image is from a different registry, check the default first, then the configured registry.
+		defaultRef := resolver.MustGetReference(images.Runtime, images.WithRegistry(images.DefaultRegistry))
```
dweomer, line 357: See https://github.com/rancher/rke2/pull/715/files#r583921244

pkg/images/images.go
```diff
@@ -1,53 +1,213 @@
...
+	Pause                 = "pause-image"
 )
 
+// These defaults are overridden at build time and do not need to be updated here
 var (
-	KubernetesVersion = "v1.20.2"      // make sure this matches what is in the scripts/version.sh script
-	PauseVersion      = "3.2"          // make sure this matches what is in the scripts/build-images script
-	EtcdVersion       = "v3.4.13-k3s1" // make sure this matches what is in the scripts/build-images script
-	RuntimeImageName  = "rke2-runtime"
+	DefaultRegistry        = name.DefaultRegistry
```
dweomer, line 26: :+1:

pkg/images/images.go
```diff
@@ -1,53 +1,213 @@
...
-type Images struct {
-	SystemDefaultRegistry string `json:"system-default-registry"`
-	Runtime               string `json:"runtime"`
-	KubeAPIServer         string `json:"kube-apiserver"`
-	KubeControllManager   string `json:"kube-controller-manager"`
-	KubeScheduler         string `json:"kube-scheduler"`
-	ETCD                  string `json:"etcd"`
-	Pause                 string `json:"pause"`
+// ROption is an option to modify image resolution behavior.
+type ROption func(name.Reference) (name.Reference, error)
```
dweomer, line 34:
```suggestion
type ResolverOpt func(name.Reference) (name.Reference, error)
```

dweomer: A stylistic suggestion that enhances grokability.

scripts/build-binary
```diff
@@ -20,10 +20,14 @@ BUILDTAGS="selinux netgo osusergo no_stage static_build sqlite_omit_load_extensi
...
-        -X ${RKE2_PKG}/pkg/images.KubernetesVersion=${DOCKERIZED_VERSION}
+        -X ${K3S_PKG}/pkg/version.GitCommit=${REVISION}
         -X ${K3S_PKG}/pkg/version.Program=${PROG}
         -X ${K3S_PKG}/pkg/version.Version=${VERSION}
-        -X ${K3S_PKG}/pkg/version.GitCommit=${REVISION}
+        -X ${RKE2_PKG}/pkg/images.DefaultRegistry=${REGISTRY}
+        -X ${RKE2_PKG}/pkg/images.DefaultEtcdImage=rancher/hardened-etcd:${ETCD_VERSION}
+        -X ${RKE2_PKG}/pkg/images.DefaultKubernetesImage=${REPO}/hardened-kubernetes:${DOCKERIZED_VERSION}
+        -X ${RKE2_PKG}/pkg/images.DefaultPauseImage=rancher/pause:${PAUSE_VERSION}
+        -X ${RKE2_PKG}/pkg/images.DefaultRuntimeImage=${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}
```
dweomer, line 30: I appreciate the enhanced lexical ordering.

scripts/package-images
```diff
@@ -8,4 +8,5 @@ source ./scripts/version.sh
 mkdir -p dist/artifacts
 
 cp -f build/images.txt dist/artifacts/${PROG}-images.${PLATFORM}.txt
-gzip < build/images/${PROG}-airgap.tar > dist/artifacts/${PROG}-images.${PLATFORM}.tar.gz
+zstd -T0 -16 -f --long=25 build/images/${PROG}-airgap.tar -o dist/artifacts/${PROG}-images.${PLATFORM}.tar.zstd
+gzip -v -c build/images/${PROG}-airgap.tar > dist/artifacts/${PROG}-images.${PLATFORM}.tar.gz
```
dweomer, line 12: Would be a real nice to have for build times to run these in parallel but I realize that means job control, ugh.

pkg/bootstrap/bootstrap.go
brandond: I wish I'd known about this yesterday!

scripts/package-images
brandond: Yeah, same :/

scripts/build-binary
brandond: yay for sorting!

pkg/images/images.go
brandond: 🎉

brandond:
* got rid of MustGetReference in places where we're returning an error anyway
* extended name.Reference into a few places we were passing around strings
* Used the containerd directory copy function

dweomer (approved):

pkg/bootstrap/readcloser.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"io"
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/rancher/wrangler/pkg/merr"
+)
+
+// Explicit interface checks
+var _ io.ReadCloser = &zstdReadCloser{}
```
briandowns, line 11: I like that we're actively continuing this pattern.

brandond:
Based on feedback from @briandowns, cleaned up ReadCloser wrappers and added shortcut return in getDefaultImage

briandowns (approved):

Issue #705, opened 2021-02-17
Cut 1.19.8 release


davidnuzik:
Upstream K8s: https://github.com/kubernetes/sig-release/issues/1466
K3s: https://github.com/k3s-io/k3s/issues/2943

**Notice:**
Upstream plans to get patch releases out by 2/17/2021.

**Summary:**
Task covering patch release work.
Dev Complete: 2/12/2021

**List of required releases:**
_To release as soon as able for QA:_
- v1.19.8-rc1+k3s1 

_To release once have approval from QA:_
- v1.19.8+k3s1

**Prep work:**
- PM: Dev and QA team to be notified of the incoming releases - send calendar invite
- PM: Dev and QA team to be notified of the date we will mark the latest release as stable - send calendar invite
- QA: Review changes and understand testing efforts
- PM: Rancher2 team notified of the release (create rancher/rancher issue and put in milestone. They will need to update rancher2 Dockerfile, etc before their next planned Rancher release) rancher/rancher#30492 (comment)
- [X] Release Captain: Prepare release notes in our private release-notes repo (submit PR for changes taking care to carefully check links and the components, once merged, create the release in GitHub and mark as a draft and check the pre-release box, fill in title, set target release branch, leave tag version blank for now until we are ready to release)
- [x] QA: Validate and close out all issues in the release milestone.
- [x] PM: Create an issue to mark the latest release as stable (this is done typically not later than 1 days post patch release, 7 days post new minor release) https://github.com/rancher/rke2/issues/675

**Vendor and release work:**
- [X] Release Captain: Tag and release any necessary RCs for QA to test KDM
- [X] Release Captain: Tag and release when have QA approval

**Post-Release work:**
- [X] Release Captain: Once release is fully complete (CI is all green and all release artifacts exist), edit the release, uncheck "Pre-release", and save.
- [x] PM: Close the milestone in GitHub.

PR #725, opened 2021-02-24
Update hardened images

erikwilson:
#### Proposed Changes ####
Update dependent images to new base OS versions

#### Linked Issues ####
https://github.com/rancher/rke2/issues/705


brandond (approved):
It would be nice to guarantee that we will rebuild these in lockstep so that we can just use a single var for the upstream image build datestamp.

dweomer (approved):

briandowns (approved):

Issue #617, opened 2020-12-16
Documentation for 616 - Node Password


davidnuzik:
Reference https://github.com/rancher/rke2/issues/616

This change will require some special documentation. Update /docs accordingly in the appropriate location. Submit PR and anyone on the team should be able to give at least 1 LGTM to merge this change in.

PR #735, opened 2021-02-27
Update docs for node-password

erikwilson:
#### Linked Issues ####
For https://github.com/rancher/rke2/issues/617

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -68,6 +68,18 @@ RKE2 agents can be configured with the options `node-label` and `node-taint` whi
 
 If you want to change node labels and taints after node registration you should use `kubectl`. Refer to the official Kubernetes documentation for details on how to add [taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) and [node labels](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/#add-a-label-to-a-node).
 
+# How Agent Node Registration Works
+
+Agent nodes are registered with a websocket connection initiated by the `rke2 agent` process, and the connection is maintained by a client-side load balancer running as part of the agent process.
+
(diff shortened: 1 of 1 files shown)


docs/advanced.md
```diff
@@ -68,6 +68,18 @@ RKE2 agents can be configured with the options `node-label` and `node-taint` whi
...
 
+# How Agent Node Registration Works
+
+Agent nodes are registered with a websocket connection initiated by the `rke2 agent` process, and the connection is maintained by a client-side load balancer running as part of the agent process.
+
+Agents will register with the server using the node cluster secret along with a randomly generated password for the node, stored at `/etc/rancher/node/password`. The server will store the passwords for individual nodes as Kubernetes secrets, and any subsequent attempts must use the same password. Node password secrets are stored in the `kube-system` namespace with names using the template `<host>.node-password.rke2`.
+
+Note: Prior to RKE2 v1.20.2 servers stored passwords on disk at `/var/lib/rancher/rke2/server/cred/node-passwd`.
+
+If the `/etc/rancher/node` directory of an agent is removed, the password file should be recreated for the agent, or the entry removed from the server.
```
brandond, line 79: A few minor wording nits:
```suggestion
Agent nodes are registered via a websocket connection initiated by the `rke2 agent` process, and the connection is maintained by a client-side load balancer running as part of the agent process.

Agents register with the server using the cluster secret portion of the join token, along with a randomly generated node-specific password, which is stored on the agent at `/etc/rancher/node/password`. The server will store the passwords for individual nodes as Kubernetes secrets, and any subsequent attempts must use the same password. Node password secrets are stored in the `kube-system` namespace with names using the template `<host>.node-password.rke2`. These secrets are deleted when the corresponding Kubernetes node is deleted.

Note: Prior to RKE2 v1.20.2 servers stored passwords on disk at `/var/lib/rancher/rke2/server/cred/node-passwd`.

If the `/etc/rancher/node` directory of an agent is removed, the password file should be recreated for the agent prior to startup, or the entry removed from the server or Kubernetes cluster (depending on the RKE2 version).
```

brandond (changes requested):

brandond (approved):

dweomer (approved):

Issue #742, opened 2021-03-02
RKE2 v1.20 fails on startup when installed from tarball


brandond:
Startup fails with:

```
Mar 02 14:42:26 sles01 rke2[5437]: time="2021-03-02T14:42:26-08:00" level=warning msg="not running in CIS 1.5 mode"
Mar 02 14:42:26 sles01 rke2[5437]: time="2021-03-02T14:42:26-08:00" level=info msg="Pulling runtime image index.docker.io/rancher/rke2-runtime:v1.20.4-rc2-rke2r1"
Mar 02 14:42:26 sles01 rke2[5437]: time="2021-03-02T14:42:26-08:00" level=fatal msg="failed to get runtime image rancher/rke2-runtime:v1.20.4-rc2-rke2r1: .dockercfg: $HOME is not defined"
```

This is because systemd does not set a HOME directory, and the updated go-containerregistry code in v1.20 now returns an error if the environment does not specify a HOME directory.

The RPM install provides a default home, but the tarball install does not:
https://github.com/rancher/rke2-packaging/blob/82c058a3ed793e8f5e6ed56c755e0759dd764bda/rpm/centos7/server/rke2-server.env#L1

Can be worked around with:
```bash
echo HOME=/root >> /usr/local/lib/systemd/system/rke2-server.env
```

ShylajaDevadiga:
Validated using rke2 version v1.20.4-rc3+rke2r1, on ubuntu 20.04/SLES 15 SP2, installation from tarball is successful.
```
$ rke2 -v
rke2 version v1.20.4-rc3+rke2r1 (06ebdfaf63aa9757945d96556842570660c8ce97)
go version go1.15.8b5
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal
$ kubectl get nodes
NAME               STATUS   ROLES                       AGE     VERSION
ip-172-31-20-246   Ready    control-plane,etcd,master   3m30s   v1.20.4-rc3+rke2r1
ubuntu@ip-172-31-20-246:~$ 
```

Issue #737, opened 2021-03-01
in SLES2/OpenSUSE rke2 services don't get restarted after reboot


wrkode:
The issue is the unusual unpacking of the RKE2 tarball and unusual mount points of in default partitioning of SLES. In default installations of SLES/OpenSUSE we'll have /usr/local defined as a discrete mount point. This is an issue because systemd will try to bring the service up before mounting /usr/local.
to Keep RKE2 distro agnostic, In my opinion install.sh should unpack the .service files in /usr/lib/systemd/system/.

To solve the issue, after unpacking (run ./install.sh), I have moved the files to /usr/lib/systemd/system and ran "systemctl daemon-reload" and "systemctl enable rke2-server.service"

brandond:
You might look into adding a `x-systemd.required-by=rke2-server.service option to the fstab entry:

https://www.freedesktop.org/software/systemd/man/systemd.mount.html

wrkode:
> Why is it mounting /usr/local so late? You might look into adding a `x-systemd.required-by=rke2-server.service option to the fstab entry for /usr/local:
> 
> https://www.freedesktop.org/software/systemd/man/systemd.mount.html

wrkode:
> Why is it mounting /usr/local so late? You might look into adding a `x-systemd.required-by=rke2-server.service option to the fstab entry for /usr/local:
> 
> https://www.freedesktop.org/software/systemd/man/systemd.mount.html

Hi Brandon,
thanks for the pointer, but personally, I wonder why systemd services are being registered/installed in /usr/local. I think that's not an appropriate location

Martin-Weiss:
FYI - also created https://bugzilla.suse.com/show_bug.cgi?id=1182237

cjellick:
Spoke briefly with @martin-weiss about this.

This manifests itself as the agent not coming back after a reboot.

@rancher-max or @ShylajaDevadiga can you test this and attempt to reprdouce please?

brandond:
According to the systemd documentation, /usr/local/lib is an appropriate location for unit files - see https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Unit%20File%20Load%20Path

> /usr/local/lib/systemd/system | System units installed by the administrator
> /usr/lib/systemd/system | System units installed by the distribution package manager

The RKE2 units are NOT being installed by the distribution package manager, and should NOT go in /usr/lib
[...]

Issue #682, opened 2021-01-31
Avoid writing into /usr to support deploying on openSUSE/SUSE MicroOS


kailiu42:
I was testing on openSUSE MicroOS and noticed that the api server just couldn't start since the begining. After checked into `/var/lib/rancher/rke2/agent/logs/kubelet.log` I saw a few repeating `mkdir` failures on `/usr/libexec/kubernetes/...` and `/usr/share/ca-certificate`, as the root(including /usr) is read only in MicroOS.

This seems is the reason of the failure, as deployment on a regular SLES 15 SP2 is OK. SLES 15 SP2 is almost the same as the MicroOS version I tested, except that the root is writable as most regular distros.

I have no problem deploying k3s on MicroOS as it seems k3s writes everything under `/var`, which is a writable separate btrfs. Can rke2 be adapted to support running on MicroOS too?

thkukuk:
According to FHS RKE2 is not allowed to write to /usr on Linux, as this is purely for distributions. At least as long as RKE2 is not shipped as RPM.
RKE2 should use /opt as at least RKE1 is already doing for /opt/cni and /opt/containerd.
Certificates should never be modified or installed in /usr/share/ca-certificate, there is absolut no guarantee that this will lead to success. Linux has tools to import new certificates, which should work on every Linux distribution.

brandond:
Unfortunately upstream has a bad habit of writing to /usr/libexec/kubernetes/, and daemonsets that want to modify stuff in /usr/share to modify CA bundles.

ibuildthecloud:
@brandond I believe /usr/libexec/kubernetes/ has been removed upstream, is that true? Either way we should just disable flexvol support in k8s, it's an irrelevant feature at this point.  That will get rid of this issue with /usr/libexec.

Also regarding /usr/share/ca-certificates it is because of https://github.com/rancher/rke2/blob/master/pkg/podexecutor/staticpod.go#L122.  We should first iterate through that slice and see if the directory exists on the host before adding it to the pod spec.

sdrahn:
@brandond and @ibuildthecloud : Question on this from my side would be if there is a workaround available until it is addressed in RKE2. Could you comment on that?

The-Loeki:
Same problem on CoreOS / Flatcar /  rpm-ostree's; all mount /usr ro.

https://github.com/rancher/rke2/blob/master/pkg/podexecutor/staticpod.go#L34
[...]

PR #743, opened 2021-03-03
Fix tarball installs on systems with odd /usr/local

brandond:
#### Proposed Changes ####

When installing tarballs, if /usr/local is read-only or on a dediated partition, install to /opt/rke2 instead. If not installing to /usr/local, drop the systemd units into /etc/systemd/system

This should fix installations on distributions where /usr is mounted read-only, or where /usr/local is a separate partition that is mounted after systemd scans for units.

#### Types of Changes ####

tarball install 

#### Verification ####

Install on non-RHEL/CentOS host with dedicated /usr/local partition, or read-only /usr

#### Linked Issues ####

#737 
#742
#682 (partial)

#### Further Comments ####

Up for some discussion on how to properly handle these situations, but this seems like a good starting point.

### DIFF
--- a/bundle/lib/systemd/system/rke2-agent.env
+++ b/bundle/lib/systemd/system/rke2-agent.env
@@ -1,0 +1,1 @@
+HOME=/root
--- a/bundle/lib/systemd/system/rke2-server.env
+++ b/bundle/lib/systemd/system/rke2-server.env
@@ -1,0 +1,1 @@
+HOME=/root
--- a/install.sh
+++ b/install.sh
@@ -33,7 +33,12 @@ fi
 #     Version of the rke2 RPM release to install.
 #     Format would be like "1.el7" or "2.el8"
 #
+#   - INSTALL_RKE2_TAR_PREFIX
+#     Installation prefix when using the tar installation method.
+#     Default is /usr/local, unless /usr/local is read-only or has a dedicated mount point,
+#     in which case /opt/rke2 is used instead.
 
+DEFAULT_TAR_PREFIX=/usr/local
 
 # info logs the given argument at info log level.
 info() {
@@ -53,6 +58,18 @@ fatal() {
     fi
     exit 1
 }
+
+# check_target_mountpoint return success if the target directory is on a dedicated mount point
+check_target_mountpoint() {
+    mountpoint -q ${INSTALL_RKE2_TAR_PREFIX}
+}
+
+# check_target_ro returns success if the target directory is read-only
+check_target_ro() {
+    touch ${INSTALL_RKE2_TAR_PREFIX}/.rke2-ro-test && rm -rf ${INSTALL_RKE2_TAR_PREFIX}/.rke2-ro-test
+    test $? -ne 0
+}
+
 
 # setup_env defines needed environment variables.
 setup_env() {
@@ -75,6 +92,16 @@ setup_env() {
     # --- use yum install method if available by default
     if [ -z "${INSTALL_RKE2_METHOD}" ] && command -v yum >/dev/null 2>&1; then
         INSTALL_RKE2_METHOD=yum
+    fi
+
+    # --- install tarball to /usr/local by default, except if /usr/local is on a separate partition or is read-only
+    # --- in which case we go into /opt/rke2.
+    if [ -z "${INSTALL_RKE2_TAR_PREFIX}" ]; then
+        INSTALL_RKE2_TAR_PREFIX=${DEFAULT_TAR_PREFIX}
+        if check_target_mountpoint || check_target_ro; then
+            INSTALL_RKE2_TAR_PREFIX=/opt/rke2
+            warn "${DEFAULT_TAR_PREFIX} is read-only or a mount point; installing to ${INSTALL_RKE2_TAR_PREFIX}"
+        fi
     fi
 }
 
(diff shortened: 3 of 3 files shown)


brandond:
**NOTE:** If you see this, please check the K3s logs for "Mount cannot be satisfied for container" messages, as upstream has recently added more verbose logging to try to collect info on the root cause.

bundle/lib/systemd/system/rke2-server.env
```diff
@@ -0,0 +1 @@
+HOME=/root
```
Oats87, line 1: This is likely to cause issues with the RKE2-packaging repo for RPMs, because it will "override" the specific `RKE2_SELINUX=true` change we made: https://github.com/rancher/rke2-packaging/blob/master/rpm/centos7/server/rke2-server.env

bundle/lib/systemd/system/rke2-agent.env
```diff
@@ -0,0 +1 @@
+HOME=/root
```
Oats87, line 1: This is likely to cause issues with the RKE2-packaging repo for RPMs, because it will "override" the specific RKE2_SELINUX=true change we made: https://github.com/rancher/rke2-packaging/blob/master/rpm/centos7/agent/rke2-agent.env

install.sh
```diff
@@ -211,9 +232,15 @@ verify_tarball() {
...
-    mkdir -p /usr/local
-    tar xzf $TMP_TARBALL -C /usr/local
+    info "unpacking tarball file to ${INSTALL_RKE2_TAR_DIR}"
+    mkdir -p ${INSTALL_RKE2_TAR_DIR}
+    tar xzf $TMP_TARBALL -C ${INSTALL_RKE2_TAR_DIR}
+    if [ "${INSTALL_RKE2_TAR_DIR}" != "/usr/local" ]; then
+        info "updating tarball contents to reflect install path"
+        sed -i "s|/usr/local|${INSTALL_RKE2_TAR_DIR}|" ${INSTALL_RKE2_TAR_DIR}/lib/systemd/system/rke2-*.service ${INSTALL_RKE2_TAR_DIR}/bin/rke2-uninstall.sh
+        mv -vf ${INSTALL_RKE2_TAR_DIR}/lib/systemd/system/rke2-*.service /etc/systemd/system/
+        info "you may want to run:  export PATH=\$PATH:${INSTALL_RKE2_TAR_DIR}/bin"
```
Oats87, line 242: Can we provide better (i.e. more verbose) information telling the user where the tarball contents are ending up? i.e. between `/etc/systemd/system` and `/opt/rke2`

Oats87 (changes requested):

install.sh
brandond: We do that up here at https://github.com/rancher/rke2/pull/743/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR235

bundle/lib/systemd/system/rke2-agent.env
brandond: According to the spec file the RPM's .env file goes to `%{_sysconfdir}/sysconfig/rke2-agent`. Both files are loaded independently via EnvironmentFile in the unit, and we're not unsetting RKE2_SELINUX in this one, so I don't see how it would override it.

bundle/lib/systemd/system/rke2-server.env
brandond: I don't think this will override it; see other comment.

install.sh
Oats87: Sure, we say we're untarring to `/opt/rke2` but other than the `.service` files being moved with a `-v` there really isn't much context there to the end user. I think we should more clearly state that we are explicitly placing the unit file in the `/etc/systemd/system` directory as this diverts from existing behavior of the tarball, otherwise a user might assume that the unit file should be in `/opt/rke2/lib/systemd/system` even though systemd doesn't actually search this directory.

bundle/lib/systemd/system/rke2-agent.env
Oats87: I was concerned not because it's ending up as `sysconfig/rke2-agent` but because the RPM build process performs an `untar` of the contents to the `_builddir` and then cherry-picks content out of the `_builddir`; however, looking more closely at the RPM build process, it looks as if we will utilize the copy of `rke2-agent.env` from the `_sourcedir` rather than `_builddir` and as such, this should be OK.

bundle/lib/systemd/system/rke2-server.env
Oats87: My other comment addresses the resolution of this.

install.sh
brandond: Fair enough, how's it look now?

install.sh
```diff
@@ -211,9 +233,16 @@ verify_tarball() {
     fi
 }
 unpack_tarball() {
-    info "unpacking tarball file"
-    mkdir -p /usr/local
-    tar xzf $TMP_TARBALL -C /usr/local
+    info "unpacking tarball file to ${INSTALL_RKE2_TAR_DIR}"
+    mkdir -p ${INSTALL_RKE2_TAR_DIR}
+    tar xzf $TMP_TARBALL -C ${INSTALL_RKE2_TAR_DIR}
+    if [ "${INSTALL_RKE2_TAR_DIR}" != "/usr/local" ]; then
```
briandowns, line 239: Small nit but can we define a var for this and reference that throughout this script?

install.sh
Oats87: Looks great!

Oats87 (approved):
LGTM

install.sh
brandond: done!

briandowns (approved):

dweomer (approved):
Longer term, if we want to properly support alternate install paths we should make it less magic and more determinant up front. This means, to me, at install time we drop an envvar file somewhere that the scripts (uninstall, killall) know to look for and source, with of course an `INSTALL_RKE2_PREFIX` envvar that can be overridden (with the /usr/local sensible default subject to an introspection pass such as is done here).

install.sh
```diff
@@ -33,7 +33,12 @@ fi
 #     Version of the rke2 RPM release to install.
 #     Format would be like "1.el7" or "2.el8"
 #
+#   - INSTALL_RKE2_TAR_DIR
```
dweomer, line 36: I'd prefer this be called something like `INSTALL_RKE2_PREFIX` or `INSTALL_RKE2_ROOT`

dweomer: Or possibly `INSTALL_RKE2_TAR_PREFIX`

brandond: I wanted to make it clear that it was `tar` specific since it does not relocate RPM installs.

install.sh
```diff
@@ -54,6 +59,18 @@ fatal() {
     exit 1
 }
 
+# check_target_mountpoint return success if the target directory is on a dedicated mount point
+check_target_mountpoint() {
```
dweomer, line 63: Should we instead fallback to fstab inspection iff `mountpoint -q /usr/local` fails (or the tool isn't present)?

brandond: I'm fine with just using mountpoint directly, now that I know about it. I think we can expect util-linux or the busybox applet equivalent to exist pretty much everywhere.

dweomer (approved):

PR #753, opened 2021-03-06
add docs question to PR template

briandowns:
#### Proposed Changes ####

Adds an HTML comment to the Pull Request template asking if the change submitted requires an update to the documentation.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,8 @@
 
 <!-- Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. -->
 
+<!-- Does this change require an update to documentation? -->
+
 #### Types of Changes ####
 
 <!-- What types of changes does your code introduce to RKE2? Bugfix, New Feature, Breaking Change, etc -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,6 +4,8 @@
 
 <!-- Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. -->
 
+<!-- Does this change require an update to documentation? -->
```
brandond, line 7:
```suggestion
- [ ] This change requires an update to documentation.
```

brandond (approved):
Maybe make it a full-on checkbox rather than just a suggested topic?

.github/PULL_REQUEST_TEMPLATE.md
briandowns: This was agreed to be a comment. I'm happy to change it but can you sign off @davidnuzik

davidnuzik: I vote we just make it a comment. I like the idea of the checkbox -- but to be honest no one seems to use them. I use them in my gh issues for releases and no one checks them off as we go. I say go with the HTML comment for now. If it doesn't seem to be helping we can revisit this and modify as needed later.

davidnuzik (approved):
LGTM

PR #765, opened 2021-03-10
Update ssldirs handling and flexvol plugin path to avoid writing to /usr

brandond:
#### Proposed Changes ####
* Only mount cert dirs that actually exist on the host
    We support a bunch of different distro-specific cert dir paths, but were previously mounting (as DirOrCreate) all of them, even on distros that don't use them. Only mount the directories that actually exist on the node, to avoid creating the ones that don't exist.
* Change the flexvol plugin path from `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` to `/var/lib/kubelet/volumeplugins`.
    This brings us in line with RKE1, and is required for systems where `/usr` is read-only.

This should be mentioned in the release notes, in case anyone was using the old flexvol plugin path.

#### Types of Changes ####

Bugfix

#### Verification ####

* Start RKE2
* Note that no content is created at /usr/libexec/kubernetes

#### Linked Issues ####

Requires rancher/rke2-charts#56
Related to #682

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -29,6 +29,7 @@ import (
 var (
 	ssldirs = []string{
 		"/etc/ssl/certs",
+		"/etc/pki/tls/certs",
 		"/etc/ca-certificates",
 		"/usr/local/share/ca-certificates",
 		"/usr/share/ca-certificates",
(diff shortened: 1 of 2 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -158,6 +159,17 @@ func (s *StaticPodConfig) Scheduler(apiReady <-chan struct{}, args []string) err
 	})
 }
 
+// onlyExisting filters out paths from the list that cannot be accessed
+func onlyExisting(paths []string) []string {
+	existing := []string{}
```
briandowns, line 164:
```suggestion
	existing := make([]string{}, len(paths))
```

briandowns (approved):

erikwilson (approved):

Issue #745, opened 2021-03-03
Metrics API is not available when metrics server is not running on local API server when running with profile: cis-1.5


Martin-Weiss:
When deploying rke2 with agents (had three masters and three workers) and having the metric server running on the workers the command "kubectl top nodes" / "kubectl top pods" does not work.

`Error from server (ServiceUnavailable): the server is currently unable to handle the request (get nodes.metrics.k8s.io)`

Root cause seems to be that the API server is running with host network and due to that the default-network-policy does not help much..

My "workaround" was to create an additional network policy:

default-network-metrics-policy.yaml:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-network-metrics-policy
  namespace: kube-system
spec:
  ingress:
  - from:
    - podSelector: {}
  - ports:
    - port: 8443
      protocol: TCP
  podSelector:
    matchLabels:
      app: rke2-metrics-server
  policyTypes:
  - Ingress
```

--> could we get this or a similar network policy auto-created in case we specify the profile cis-1.5?

Oats87:
In a multi-node control plane cluster, when the metrics server is not on the local node, the api server is unable to communicate with the metrics server and thus will lead to API failures. This has a cascading effect such as not being able to `kubectl top` or `kubectl get pod -A` in a reasonable amount of time on various control plane hosts.

briandowns:
@Oats87  and I spent some time pairing on this to figure out what's going on. It appears as though traffic is unable to be passed from host network to the pod network destined for pods on remote nodes. This is a side-effect of:  https://kubernetes.io/docs/concepts/services-networking/network-policies/#what-you-can-t-do-with-network-policies-at-least-not-yet 

Traffic originating from the host network needs to traverse the flannel interface which results in that traffic being blocked as opposed to traffic originating from the
10.42.0.X/24 subnet.

To resolve this issue, we need to adjust the default NetworkPolicy to get all of the node IP's in the cluster and update the NetworkPolicy with an ingress rule that has
the IP of the flannel interface, e.g. 10.42.0.0, etc.

brandond:
Rather than adding a new controller to keep the policies in sync with the node list, could we just open up access to the metrics server pod? Are there negative security implications to allowing access to the metric server?

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-metrics-server-policy
  namespace: kube-system
spec:
  ingress:
  - ports:
[...]

PR #771, opened 2021-03-11
Add CIS Network Policy Controller

Oats87:
#### Proposed Changes ####
This PR adds the CIS Network Policy controller, which watches nodes and mutates a network policy in the three system namespaces when RKE2 is operating in CIS mode with a flannel-backed CNI.

#### Types of Changes ####
Add a controller to RKE2

#### Verification ####
Run RKE2 in cis mode, and verify that the network policy list in `kube-system`, `kube-public`, and `default` include the `rke2-flannel-host-networking` network policy, and validate that the ingress rules of the network policy include the `.0` of each flannel interface. 

#### Linked Issues ####
https://github.com/rancher/rke2/issues/745

#### Further Comments ####

### DIFF
--- a/pkg/controllers/cisnetworkpolicy/controller.go
+++ b/pkg/controllers/cisnetworkpolicy/controller.go
@@ -1,0 +1,155 @@
+package cisnetworkpolicy
+
+import (
+	"context"
+	"net"
+	"sort"
+
+	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/server"
+	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	core "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+const (
+	flannelPresenceLabel         = "flannel.alpha.coreos.com/public-ip"
+	flannelHostNetworkPolicyName = "rke2-flannel-host-networking"
+)
+
+func CISNetworkPolicyController(ctx context.Context, sc *server.Context) error {
+	return register(ctx, sc.Core.Core().V1().Node(), sc.K8s)
+}
+
+func register(ctx context.Context,
+	nodes coreclient.NodeController,
+	k8s kubernetes.Interface,
+) error {
+	h := &handler{
+		ctx: ctx,
+		k8s: k8s,
+	}
+	logrus.Debugf("CISNetworkPolicyController: Registering controller hooks")
+	nodes.OnChange(ctx, "cisnetworkpolicy-node", h.handle)
+	nodes.OnRemove(ctx, "cisnetworkpolicy-node", h.handle)
+	return nil
+}
+
+type handler struct {
+	ctx context.Context
+	k8s kubernetes.Interface
+}
+
+func (h *handler) handle(key string, node *core.Node) (*core.Node, error) {
+	if node == nil {
+		return nil, nil
+	}
+
+	return h.reconcileFlannelHostNetworkPolicy(key, node)
+}
+
+func (h *handler) reconcileFlannelHostNetworkPolicy(nodeName string, node *core.Node) (*core.Node, error) {
+	var np *netv1.NetworkPolicy
+	var npIR *netv1.NetworkPolicyIngressRule
+
+	npIR, err := h.generateHostNetworkPolicyIngressRule()
+	if err != nil {
+		return nil, err
+	}
+
+	np = h.generateHostNetworkingNetworkPolicy(npIR)
+
+	var namespaces = []string{
+		metav1.NamespaceSystem,
+		metav1.NamespaceDefault,
+		metav1.NamespacePublic,
+	}
+
+	for _, namespace := range namespaces {
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if npGet, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Get(h.ctx, flannelHostNetworkPolicyName, metav1.GetOptions{}); err != nil {
+				if apierrors.IsNotFound(err) {
+					if _, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Create(h.ctx, np, metav1.CreateOptions{}); err != nil {
+						if !apierrors.IsAlreadyExists(err) {
+							logrus.Errorf("CISNetworkPolicyController: Error creating network policy in namespace %s: %v", namespace, err)
+							return err
+						}
+					}
+				} else {
+					logrus.Errorf("CISNetworkPolicyController: Error getting network policy network policy in namespace %s: %v", namespace, err)
+					return err
+				}
+			} else {
+				npGet = npGet.DeepCopy()
+				npGet.Spec.Ingress[0] = *npIR
+				_, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Update(h.ctx, npGet, metav1.UpdateOptions{})
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, errors.Wrapf(err, "CISNetworkPolicyController: error working on network policy in namespace %s", namespace)
+		}
+	}
+	logrus.Debugf("CISNetworkPolicyController: Handled node change")
+	return nil, nil
+}
+
+func (h *handler) generateHostNetworkingNetworkPolicy(networkPolicyIngressRule *netv1.NetworkPolicyIngressRule) *netv1.NetworkPolicy {
+	np := &netv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: flannelHostNetworkPolicyName,
+		},
+		Spec: netv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{},
+			Ingress: []netv1.NetworkPolicyIngressRule{
+				*networkPolicyIngressRule,
+			},
+			PolicyTypes: []netv1.PolicyType{
+				netv1.PolicyTypeIngress,
+			},
+		},
+	}
+
+	return np
+}
+
+func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIngressRule, error) {
+	var npIR netv1.NetworkPolicyIngressRule
+
+	nodes, err := h.k8s.CoreV1().Nodes().List(h.ctx, metav1.ListOptions{})
+	if err != nil {
+		return &npIR, errors.Wrap(err, "CISNetworkPolicyController: problem listing nodes")
+	}
+
+	for _, node := range nodes.Items {
+		if _, ok := node.Annotations[flannelPresenceLabel]; !ok {
+			logrus.Debugf("CISNetworkPolicyController: node=%v doesn't have flannel label, skipping", node.Name)
+			continue
+		}
+		podCIDRFirstIP, _, err := net.ParseCIDR(node.Spec.PodCIDR)
+		if err != nil {
+			logrus.Debugf("CISNetworkPolicyController: node=%+v", node)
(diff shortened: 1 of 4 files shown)


pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+				if errors.IsNotFound(err) {
+					if _, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Create(h.ctx, np, metav1.CreateOptions{}); err != nil {
+						if !errors.IsAlreadyExists(err) {
+							logrus.Errorf("CISNetworkPolicyController: Error creating network policy in namespace %s: %v", namespace, err)
+							return nil, err
+						}
+					}
+				} else {
+					logrus.Errorf("CISNetworkPolicyController: Error getting network policy network policy in namespace %s: %v", namespace, err)
+					return nil, err
```
galal-hussein, line 94: I think this will display the error twice, we can wrap the error here and return it instead of displaying it and then return the original error

galal-hussein (approved):
LGTM

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	FlannelHostNetworkPolicyName = "rke2-flannel-host-networking"
+)
+
+func CISNetworkPolicyController(ctx context.Context, sc *server.Context) error {
+	Register(ctx, sc.Core.Core().V1().Node(), sc.K8s)
+
+	return nil
+}
+
+func Register(ctx context.Context,
```
dweomer, line 30: exporting this seems unnecessary (granted we break api compatibility at will with no regards to semver whatsoever)

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+					return nil, err
+				}
+			} else {
+				npGet = npGet.DeepCopy()
+				npGet.Spec.Ingress[0] = *npIR
+				_, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Update(h.ctx, npGet, metav1.UpdateOptions{})
+				if err != nil {
+					if errors.IsConflict(err) {
+						logrus.Errorf("CISNetworkPolicyController: Conflict updating network policy network policy in namespace %s: %v", namespace, err)
+						continue
```
brandond, line 103: Should this use a retry-with-backoff strategy instead of just tightly looping until it succeeds? I think there are some helpers for doing that.

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	return nil
+}
+
+type handler struct {
+	ctx context.Context
+	k8s kubernetes.Interface
+}
+
+func (h *handler) onRemove(key string, node *core.Node) (*core.Node, error) {
+	return h.reconcileFlannelHostNetworkPolicy(key, node)
```
dweomer, line 50:
```suggestion
	if node == nil {
		return nil, nil
	}
	return h.reconcileFlannelHostNetworkPolicy(key, node)
```

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return np, nil
+}
+
+func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIngressRule, error) {
+	var npIR netv1.NetworkPolicyIngressRule
+
+	nodes, err := h.k8s.CoreV1().Nodes().List(h.ctx, metav1.ListOptions{})
+	if err != nil {
+		logrus.Infof("CISNetworkPolicyController: problem listing nodes %v", err)
```
brandond, line 139: Should this be error or debug, or just not logged at all? I second Hussein's comment about not logging errors if we're also returning them.

brandond (changes requested):

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: This came from: https://github.com/k3s-io/k3s/blob/master/pkg/node/controller.go#L39-L49

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,166 @@
...
+				_, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Update(h.ctx, npGet, metav1.UpdateOptions{})
+				if err != nil {
+					if errors.IsConflict(err) {
+						logrus.Errorf("CISNetworkPolicyController: Conflict updating network policy network policy in namespace %s: %v", namespace, err)
+						continue
+					}
+				}
+				applied = true
+
+			}
```
dweomer, line 108: Shouldn't we be using https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict here? With multiple servers, races will happen.

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: This should be an error, but if you look at our internal thread on this, I have a reason for specifically calling `Errorf`.

dweomer: `logrus.Warnf` ?

Oats87: It's an error, and should be stated as such.

pkg/controllers/cisnetworkpolicy/controller.go
dweomer: Well, that code is wrong too. Event handlers can and will be invoked with a `nil` object.

pkg/controllers/cisnetworkpolicy/controller.go
dweomer: `. o ( bubble it up? )`

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: Yes, pushing a change with this now.

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: Yeah, you're right. I'm going to just return a `fmt.Errorf`

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: No longer exported

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: I consolidated into a `handle` function

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: Now using retry

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: It's an error now and bubbled up

pkg/controllers/cisnetworkpolicy/controller.go
Oats87: Now using retry.

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	for _, namespace := range namespaces {
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if npGet, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Get(h.ctx, FlannelHostNetworkPolicyName, metav1.GetOptions{}); err != nil {
+				if errors.IsNotFound(err) {
+					if _, err := h.k8s.NetworkingV1().NetworkPolicies(namespace).Create(h.ctx, np, metav1.CreateOptions{}); err != nil {
+						if !errors.IsAlreadyExists(err) {
+							logrus.Errorf("CISNetworkPolicyController: Error creating network policy in namespace %s: %v", namespace, err)
+							return err
+						}
```
dweomer, line 86:
```suggestion
						return err
```

Oats87: This is done very specifically for the purposes of being able to clearly debug this without a bunch of scaffolding around the message, from prior troubleshooting experience.

dweomer (approved):

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	np, err = h.generateHostNetworkingNetworkPolicy(npIR)
+	if err != nil {
+		return nil, err
+	}
+
+	var namespaces = []string{
+		metav1.NamespaceSystem,
+		metav1.NamespaceDefault,
+		metav1.NamespacePublic,
```
brandond, line 75: 💯

brandond (approved):

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	core "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	FlannelPresenceLabel         = "flannel.alpha.coreos.com/public-ip"
```
briandowns, line 23: Does this need to be exported from this package?

pkg/rke2/rke2.go
```diff
@@ -7,10 +7,13 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/rancher/rke2/pkg/controllers/cisnetworkpolicy"
+
 	"github.com/rancher/k3s/pkg/agent/config"
 	"github.com/rancher/k3s/pkg/cli/agent"
 	"github.com/rancher/k3s/pkg/cli/cmds"
 	"github.com/rancher/k3s/pkg/cli/server"
+	rawServer "github.com/rancher/k3s/pkg/server"
```
briandowns, line 16: What does "rawServer" mean by comparison to the other server import?

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return np, nil
+}
+
+func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIngressRule, error) {
+	var npIR netv1.NetworkPolicyIngressRule
+
+	nodes, err := h.k8s.CoreV1().Nodes().List(h.ctx, metav1.ListOptions{})
+	if err != nil {
+		return &npIR, fmt.Errorf("CISNetworkPolicyController: problem listing nodes %v", err)
```
briandowns, line 134: Can we use errors.Wrap here?

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("CISNetworkPolicyController: error working on network policy: %v", err)
+		}
+	}
+	logrus.Debugf("CISNetworkPolicyController: Handled node change")
+	return nil, nil
+}
+
+func (h *handler) generateHostNetworkingNetworkPolicy(networkPolicyIngressRule *netv1.NetworkPolicyIngressRule) (*netv1.NetworkPolicy, error) {
```
briandowns, line 110: Does this method return an error to satisfy an interface? If not, we should update it to not return an error that's never created.

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+func register(ctx context.Context,
+	nodes coreclient.NodeController,
+	k8s kubernetes.Interface,
+) error {
+	h := &handler{
+		ctx: ctx,
+		k8s: k8s,
+	}
+	logrus.Debugf("CISNetworkPolicyController: Registering controller hooks")
```
briandowns, line 39: Does this need an "\n" added or use Debugln ?

briandowns (approved):

Issue #784, opened 2021-03-15
RKE2 Agent does not extract images with „system-default-registry“


Martin-Weiss:
Agents try to download images from on-premise registry even though they are included in the images tar ball

```
# rke2 --version
rke2 version v1.20.4+rke2r1 (e0a5163fef9227fa6c19e5f597101d4939077793)
go version go1.15.8b5
```
In server and agent config.yaml:

`system-default-registry: registry.domain.de`

On servers `crictl images` shows this:

```
docker.io/rancher/pause                                           3.2                              e004ddc1b078f       686kB
registry.domain.de/rancher/pause                                     3.2                              e004ddc1b078f       686kB
docker.io/rancher/rke2-runtime                                    v1.20.4-rke2r1                   2f48f0df56b54       270MB
registry.domain.de/rancher/rke2-runtime                              v1.20.4-rke2r1                   2f48f0df56b54       270MB
registry.domain.de/rancher/hardened-kube-proxy                       v1.20.4-build20210302            fe40618686a2f       235MB
docker.io/rancher/hardened-kube-proxy                             v1.20.4-build20210302            fe40618686a2f       235MB
```

On agents `crictl images` shows this:

```
docker.io/rancher/pause                                           3.2                              e004ddc1b078f       686kB
docker.io/rancher/rke2-runtime                                    v1.20.4-rke2r1                   2f48f0df56b54       270MB
docker.io/rancher/hardened-kube-proxy                             v1.20.4-build20210302            fe40618686a2f       235MB
```

--> Due to that the agent tries to fetch the pause image registry.domain.de/rancher/pause:3.2 from the on premise registry which fails due to missing namespace mapping support.. 

Why do agents act different compared to servers for these images?
Is this a bug that can be fixed?

brandond:
Yep, looks like this is a bug. As a workaround, until this is fixed, you should be able to add `airgap-extra-registry: registry.domain.de` to the config on the agent,

Martin-Weiss:
> 
> 
> Yep, this looks like a bug.
> 
> As a workaround, until this is fixed, you should be able to add `airgap-extra-registry: registry.domain.de` to the config on the agent to get the images retagged properly.

Cool - thanks  - did not know that this parameter exists.. do you have some details about this parameter? I could not find it here.. https://docs.rke2.io/install/install_options/agent_config/

Our workaround for the moment was this:

[...]

PR #787, opened 2021-03-15
Honor system-default-registry retagging behavior on agents

brandond:
#### Proposed Changes ####

Fix retagging tarball images on agents to appear to come from system-default-registry. The flag was only being passed through on servers.

No docs change necessary; should be called out in release notes.

#### Types of Changes ####

Bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

#784

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -48,12 +48,6 @@ func Server(clx *cli.Context, cfg Config) error {
 		return err
 	}
 
-	// If system-default-registry is set, retag images imported from tarballs to appear to
-	// come from this registry.
-	if cfg.Images.SystemDefaultRegistry != "" {
-		clx.Set("airgap-extra-registry", cfg.Images.SystemDefaultRegistry)
-	}
-
 	// Disable all disableable k3s packaged components. In addition to manifests,
 	// this also disables several integrated controllers.
 	disableItems := strings.Split(cmds.DisableItems, ",")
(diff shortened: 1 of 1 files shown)


dweomer (approved):

pkg/rke2/rke2.go
```diff
@@ -86,6 +80,12 @@ func setup(clx *cli.Context, cfg Config) error {
 		auditPolicyFile = defaultAuditPolicyFile
 	}
 
+	// If system-default-registry is set, retag images imported from tarballs to appear to
+	// come from this registry.
+	if cfg.Images.SystemDefaultRegistry != "" {
```
briandowns, line 85: Are there any relevant validations we should be doing on the contents of `cfg.Images.SystemDefaultRegistry`?

erikwilson (approved):

galal-hussein (approved):

pkg/rke2/rke2.go
brandond: yeah, `images.NewResolver(cfg.Images)` a couple lines down validates it.

I suppose this could go just above the `defaults.Set` call below but the the effective result is the same if the input is invalid - an error is displayed and rke2 exits.

briandowns: We probably want to do the validation prior to the assignment/storage but not a blocker at this point.

briandowns (approved):

Issue #782, opened 2021-03-14
systemctl restart rke2-server causes kubelet to restart forever 


Quantas:
I just setup a testing cluster using RKE2 on friday, and I noticed a server systemctl service had been restarted for some reason or another, and in the events log in kubernetes, it keeps repeating that the kubelet has been restarted on that node until I eventually just reboot the server, then it comes up normally.  Is this expected? Anyone else notice this? Are there some logs you would like me to obtain? Thanks!

Quantas:
Just a note, it is a 6 node cluster on CentOS 7, 3 are server and 3 are agent, using round robin DNS for the cluster url.

brandond:
The journald logs from rke2-server or rke2-agent on the node in question would be a good start.

Quantas:
Sounds good, I will grab those logs when I'm back at work on Monday. As an aside, I just tested a similar scenario at home with ubuntu without SELinux and didn't have any issues, my cluster at work has SELinux enabled. I will get those logs on Monday though :) Thanks!

andrew-landsverk-win:
Here is what I've discovered, after doing `systemctl restart rke2-server` I get a message like the following:

> Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: F0315 08:00:52.262027    4635 server.go:162] listen tcp 0.0.0.0:10250: bind: address already in use
Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: goroutine 859 [running]:
Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: k8s.io/klog/v2.stacks(0xc0004c8d01, 0xc000290500, 0x64, 0x241)
Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: /go/src/kubernetes/vendor/k8s.io/klog/v2/klog.go:1026 +0xb9
Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: k8s.io/klog/v2.(*loggingT).output(0x6f76da0, 0xc000000003, 0x0, 0x0, 0xc000982310, 0x6e01433, 0x9, 0xa2, 0x410900)
Mar 15 08:00:52 dev-rke2-master-1.dev.internal rke2[4618]: /go/src/kubernetes/vendor/k8s.io/klog/v2/klog.go:975 +0x19b
[...]

PR #791, opened 2021-03-17
Extract runtime image contents direct to destination

brandond:
#### Proposed Changes ####

* Extract runtime image contents directly to destination.
  This prevents issues caused by binaries getting labeled with the wrong context when they are created in a temp dir. We're already using a unique parent directory per runtime image so the temp dir isn't really necessary to begin with.
* Update go-containerregistry.

No docs change, should be called out in release notes.

#### Types of Changes ####

bugfix

#### Verification ####

See steps in linked issue

#### Linked Issues ####

#782


pkg/bootstrap/bootstrap.go
```diff
@@ -181,24 +180,22 @@ func extract(targetDir string, dirs map[string]string, reader io.Reader) error {
...
+		}
 
 		switch h.Typeflag {
 		case tar.TypeDir:
-			if _, ok := dirs[targetDir]; !ok {
-				continue
-			}
-			if _, err := os.Stat(targetName); err != nil {
-				if err := os.MkdirAll(targetName, 0755); err != nil {
+			if _, err := os.Stat(targetDir); err != nil {
```
briandowns, line 191: Do we care what the error is or just the existance?

brandond: Yeah I'm not sure how defensive we need to be - if it's a permissions error or something else it'll bubble up when the mkdir fails?

briandowns: I hate to prolong this but for clarity of intent, it'd probably be good to perform an explicit existence check, make the directory if it doesn't, and log or return otherwise.

brandond: Turns out MkdirAll will actually stat the target path for you and just returns without error if it already exists.

briandowns: Should we remove the stat call all together then?

briandowns (approved):

erikwilson (approved):

PR #794, opened 2021-03-17
Activate systemd with tarball installation

zaggash:
#### Proposed Changes ####

Edit documentation about Tarball systemd activation

#### Types of Changes ####

Quickstart page modification about systemd enablement.

### DIFF
--- a/docs/install/quickstart.md
+++ b/docs/install/quickstart.md
@@ -24,6 +24,11 @@ This will install the `rke2-server` service and the `rke2` binary onto your mach
 ```
 systemctl enable rke2-server.service
 ```
+```  
+For a Tarball installation, systemd service is not automatically installed.  
+Activate it like so:
+```
+systemctl enable /usr/local/lib/systemd/system/rke2-server.service
 
 #### 3. Start the service
 ```


brandond:
This shouldn't be necessary. What distro are you on that doesn't search for units in /usr/local/lib/systemd/system?

zaggash:
Default Ubuntu `ubuntu-18-04-x64` from DigitalOcean.

Using the quickstart command up to the `systemctl enable` where is says there is no such unit file.

brandond:
Hmm, on my Ubuntu systems this is a valid path:

```
(python38) brandond@seago:~$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"

(python38) brandond@seago:~$ man systemd | grep -A4 'System unit directories'
       System unit directories
           The systemd system manager reads unit configuration from various directories. Packages that want to install unit files shall place them in the
           directory returned by pkg-config systemd --variable=systemdsystemunitdir. Other directories checked are /usr/local/lib/systemd/system and
           /lib/systemd/system. User configuration always takes precedence.  pkg-config systemd --variable=systemdsystemconfdir returns the path of the system
           configuration directory. Packages should alter the content of these directories only with the enable and disable commands of the systemctl(1) tool.
```

zaggash:
I just tried to reproduce the issue twice and I can't.
Maybe my previous node was broken.

Sorry for the noise.

Issue #786, opened 2021-03-15
NetworkManager interferes with network related components


rancher-max:
This is mostly visible when running CIS scans through rancher as it will hang forever. It is possible to see if the issue is present or not but running `ip route` on a server node after adding all nodes to the cluster, then pinging each `.0` CIDR. Each should return a response.
Related rancher docs issue: https://github.com/rancher/docs/issues/1052

Potential Workarounds:
- https://dev.to/flrichar/netmanager-leave-calico-interfaces-alone-38de
- Upon rebooting the affected node(s), you might get lucky and have the issue fixed. Conversely, if there is no issue and you reboot the node, it might cause a problem.

@Oats87 to add more info as needed.

Oats87:
Potential workaround:

Create a file: `/etc/NetworkManager/conf.d/canal.conf`
with the contents:
```
[keyfile]
unmanaged-devices=interface-name:cali*;interface-name:flannel*
```

Oats87:
@rancher-max 

`systemctl reload NetworkManager` should suffice to reload the configuration into place.

Oats87:
Verified workaround:
```
cat << EOF > /etc/NetworkManager/conf.d/rke2-canal.conf
[keyfile]
unmanaged-devices=interface-name:cali*;interface-name:flannel*
EOF
systemctl reload NetworkManager
```

brandond:
Is this common enough that we should handle it in the installer?

Oats87:
@brandond I think it is common enough that it may make sense to handle it in the installer, but is only necessary on `rke2-canal` enabled systems. Past this, I am also weary of applying a NetworkManager configuration change and reloading NetworkManager in an automated fashion, as this could accidentally orphan a node if there are other configurations we are not aware of.

briandowns:
Should it be an opt-in in the installer with docs saying you need to but giving the customer a means of not doing if they choose? (maybe they have some other work around)

PR #836, opened 2021-04-01
[docs] Add call out for networkmanager workaround

Oats87:
https://github.com/rancher/rke2/issues/786

Docs change to add NetworkManager call out.

### DIFF
--- a/docs/install/airgap.md
+++ b/docs/install/airgap.md
@@ -1,5 +1,7 @@
 # Air-Gap Install
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
+
 RKE2 can be installed in an air-gapped environment with two different methods.
 You can either deploy via the `rke2-airgap-images` tarball release artifact, or by using a private registry.
 
--- a/docs/install/methods.md
+++ b/docs/install/methods.md
@@ -2,6 +2,8 @@
 title: Installation Methods
 ---
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
+
 RKE2 can be installed to a system in a number of ways, two of which are the preferred and supported methods. Those methods are tarball and RPM. The install script referenced in the Quick Start is a wrapper around these two methods.
 
 This document explains these installation methods in greater detail. 
--- a/docs/install/quickstart.md
+++ b/docs/install/quickstart.md
@@ -9,6 +9,7 @@ This guide will help you quickly launch a cluster with default options.
 ### Prerequisites
 
 Make sure your environment fulfills the [requirements.](https://docs.rke2.io/install/requirements/)
+If NetworkManager is installed and enabled on your hosts, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
 
 ### Server Node Installation
 --------------
--- a/docs/install/requirements.md
+++ b/docs/install/requirements.md
@@ -29,6 +29,8 @@ RKE2 performance depends on the performance of the database, and since RKE2 runs
 
 ## Networking
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
+
 The RKE2 server needs port 6443 and 9345 to be accessible by other nodes in the cluster.
 
 All nodes need to be able to reach other nodes over UDP port 8472 when Flannel VXLAN is used.
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -6,6 +6,16 @@ This section contains current known issues and limitations with rke2. If you com
 
 Firewalld conflicts with RKE2's default Canal (Calico + Flannel) networking stack. To avoid unexpected behavior, firewalld should be disabled on systems running RKE2.
 
+## NetworkManager
+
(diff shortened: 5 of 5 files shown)


docs/known_issues.md
```diff
@@ -6,6 +6,16 @@ This section contains current known issues and limitations with rke2. If you com
...
 
+## NetworkManager
+
+NetworkManager, in stock configuration, does not work well with RKE2 and Canal. As such, if installing RKE2 on a NetworkManager enabled system, it is highly recommended to configure NetworkManager to ignore calico/flannel related network interfaces. In order to do this, create a configuration file called `rke2-canal.conf` in `/etc/NetworkManager/conf.d` with the contents:
+```bash
+[keyfile]
+unmanaged-devices=interface-name:cali*;interface-name:flannel*
+```
+
+If you have not yet installed RKE2, a simple `systemctl reload NetworkManager` will suffice to install the configuration. If performing this configuration change on a system that already has RKE2 installed, a reboot of the node is necessary to effectively apply the changes.
```
rancher-max, line 17: Instead of reboot of the node, can we say "restart of the rke2-canal pods"? Appears to be working for me when doing this instead of fully rebooting the nodes.

briandowns (approved):

docs/known_issues.md
brandond: @rancher-max its probably easier to tell people to just restart the node, vs how to find and kill the canal pods on the affected host.

docs/install/airgap.md
```diff
@@ -1,5 +1,7 @@
 # Air-Gap Install
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure the interface ignore workaround is applied.](https://docs.rke2.io/known_issues/#networkmanager)
```
brandond, line 3:
```suggestion
**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
```

docs/install/methods.md
```diff
@@ -2,6 +2,8 @@
 title: Installation Methods
 ---
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure the interface ignore workaround is applied.](https://docs.rke2.io/known_issues/#networkmanager)
```
brandond, line 5:
```suggestion
**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
```

docs/install/quickstart.md
```diff
@@ -9,6 +9,7 @@ This guide will help you quickly launch a cluster with default options.
 ### Prerequisites
 
 Make sure your environment fulfills the [requirements.](https://docs.rke2.io/install/requirements/)
+If on a NetworkManager enabled system, [ensure the interface ignore workaround is applied.](https://docs.rke2.io/known_issues/#networkmanager)
```
brandond, line 12:
```suggestion
If NetworkManager is installed and enabled on your hosts, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
```

docs/install/requirements.md
```diff
@@ -29,6 +29,8 @@ RKE2 performance depends on the performance of the database, and since RKE2 runs
 
 ## Networking
 
+**Important:** If your node has NetworkManager installed and enabled, [ensure the interface ignore workaround is applied.](https://docs.rke2.io/known_issues/#networkmanager)
```
brandond, line 32:
```suggestion
**Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
```

brandond (changes requested):

docs/known_issues.md
Oats87: I agree with @brandond , it's a little bit of a "cannon" approach but will essentially ensure the issue will not reoccur. Otherwise, it's a multi-step reload of `NetworkManager` then restart of the `rke2-canal` pods.

docs/install/airgap.md
Oats87: I will implement these commit suggestions manually.

docs/install/methods.md
Oats87: I will implement these commit suggestions manually.

docs/install/quickstart.md
Oats87: I will implement these commit suggestions manually.

docs/install/requirements.md
Oats87: I will implement these commit suggestions manually.

brandond (approved):

docs/known_issues.md
davidnuzik: @Oats87 @brandond 
The second sentence in line 17 is not necessary. We tested and you do not need to restart the daemon or reboot the node. Only a reload is necessary. Additionally, I have a question. I thought this workaround was just for Rancher - is this not a correct assumption?
- If so, then we should clearly call out this is just for if you will be installing Rancher.
- If not, we still should mention one stipulation we noticed which is that if you are applying this workaround to run Rancher, you will need to do it first before attempting to install Rancher or Rancher will not work.

davidnuzik: I synced with Max and does sound like it wouldn't be a workaround for Rancher only. This is fine but can we add a mention that if you plan to install Rancher on the cluster, to do this step before doing that? This should be clearly called out to the reader.

davidnuzik: This is because Shylaja tested in a couple of scenarios and found that we must always ensure this is done first as a pre-req before installation of Rancher or Rancher will not work. Perhaps belongs in prereqs section.

davidnuzik:
Merging, but I'd like to add in some notes with regards to Rancher installs soon. Rancher does not yet list these RKE2 versions in the support matrix though. Let's plan on updating the docs tomorrow.

Issue #866, opened 2021-04-09
Node fails to join the cluster when disable-etcd flag is passed 


ShylajaDevadiga:
Environment:
rke2 version: v1.20.5-dev+0a114cc6

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04

Cluster Configuration:
Three node cluster, 1 etcd-only, 2 all roles except etcd role, 3 all roles

**Describe the bug:**
Node with disable-etcd flags fails to join the cluster

**Steps To Reproduce:**
Create a 3 node cluster
1. etcd-only node
```
cat /etc/rancher/rke2/config.yaml
disable-api-server: true
disable-scheduler: true
disable-controller-manager: true

curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=0a114cc67c3c8400ceed5e88699e99922f525e41 sh -
```
2. node without etcd
```
$ cat /etc/rancher/rke2/config.yaml
token: <token>
server:  <server fqdn:9345>
disable-etcd: true

$ curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=0a114cc67c3c8400ceed5e88699e99922f525e41 sh -

```
3. node with all components
```
$ cat /etc/rancher/rke2/config.yaml
token: <token>
server:  <server fqdn:9345>
$ curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=0a114cc67c3c8400ceed5e88699e99922f525e41 sh -
```
**Expected behavior:**
All three nodes should join the cluster with respective roles

**Actual behavior:**
Node with disable-etcd flag does not join the cluster expecting /var/lib/rancher/rke2/server/db/etcd dir

**Additional context / logs:**
```
Apr 09 19:35:36 ip-172-31-20-25 rke2[10732]: time="2021-04-09T19:35:36Z" level=fatal msg="open /var/lib/rancher/rke2/server/db/etcd/name: no such file or directory"
Apr 09 19:35:36 ip-172-31-20-25 systemd[1]: rke2-server.service: Main process exited, code=exited, status=1/FAILURE
Apr 09 19:35:36 ip-172-31-20-25 systemd[1]: rke2-server.service: Failed with result 'exit-code'.
Apr 09 19:35:41 ip-172-31-20-25 systemd[1]: rke2-server.service: Scheduled restart job, restart counter is at 277.
[...]

PR #867, opened 2021-04-09
Fix disable etcd flag in rke2

galal-hussein:
#### Proposed Changes ####

check if disable etcd flag is passed and dont use etcd name file if its disabled

#### Types of Changes ####
bug fix

#### Linked Issues ####
https://github.com/rancher/rke2/issues/866

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -45,6 +45,7 @@ type StaticPodConfig struct {
 	DataDir         string
 	AuditPolicyFile string
 	KubeletPath     string
+	DisableETCD     bool
 }
 
 type CloudProviderConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -127,14 +128,18 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
 			args = append(args[:i], args[i+1:]...)
 		}
 	}
+	files := []string{etcdNameFile(s.DataDir)}
```
briandowns, line 131: I think we want to reverse this so we check to see if disabled and if it's not, we assign, rather than assigning and then emptying. Here and below.

galal-hussein: fixed

briandowns (approved):
LGTM

brandond (approved):

PR #901, opened 2021-04-20
[docs] Update ha.md to use upstream standard node taint for controlplane nodes.

estenrye:
The current example breaks the Rancher Monitoring package when applied during the installation.

```yaml
node-taint:
  - "CriticalAddonsOnly=true:NoExecute"
```

This breaks the pushprox-kube-etcd-client which has a toleration expecting a taint of `node-role.kubernetes.io/master` with an effect of `NoSchedule`

```
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
        operator: Equal
```

#### Proposed Changes ####

Change example to:

```yaml
node-taint:
  - "node-role.kubernetes.io/master:NoSchedule"
```

#### Types of Changes ####

Updates documentation to use upstream standard taints.

#### Challenges to Accepting This Change ####

After reconfiguring my Ansible automation for building rke2 clusters I found out that if one applies this change, it breaks the helm controller integration due to the fact that the controller has its taints for the jobs it creates hard coded.

https://github.com/k3s-io/helm-controller/blob/4ed68f1ecd2ba59cdf5090fb2c166f7055f8cf99/pkg/helm/controller.go#L287

Applying this change without changing the behavior of the helm controller would break the RKE2 bootstrapping process.

### DIFF
--- a/docs/install/ha.md
+++ b/docs/install/ha.md
@@ -50,7 +50,7 @@ tls-san:
 By default, server nodes will be schedulable and thus your workloads can get launched on them. If you wish to have a dedicated control plane where no user workloads will run, you can use taints. The `node-taint` parameter will allow you to configure nodes with taints. Here is an example of adding a node taint to the configuration file:
 ```yaml
 node-taint:
-  - "CriticalAddonsOnly=true:NoExecute"
+  - "node-role.kubernetes.io/master:NoSchedule"
 ```
 
 ### 3. Launch additional server nodes


brandond:
In addition to CriticalAddonsOnly being the expected taint, upstream has deprecated the master role label. Control-plane is the new standard role.

estenrye:
@brandond is there a larger feature here across the rke2 and the helm controller to standardize on upstream taints or would this be best resolved by enhancing the Rancher Monitoring helm chart to put the expected toleration if `CriticalAddonsOnly` when it detects it is being deployed to an RKE2 cluster?

brandond:
Lets track this in #903 - I don't think changing the taints on this side is the correct fix.

estenrye:
Sounds good, I'll close this PR for now and we'll see what they come up with.

Issue #886, opened 2021-04-15
Removing etcd role from a node needs killall to be executed to delete static pods


ShylajaDevadiga:
**Environmental Info:**
rke2 Version: 
rke2 version v1.20.5-dev+dc9b0919

Node(s) CPU architecture, OS, and Version: 
Sles 15 SP2

Cluster Configuration:
4 node cluster
1- etcd-only
2- all roles except etcd role
3- all roles
4- all roles

**Describe the bug:**
While removing etcd-role from a node that has all roles, and restarting rke2 fails as the static pods are still running. etcd-role is removed only after running rke2-killall to kill the pods and starting service again.

**Steps To Reproduce:**

- Create the cluster with 4 nodes with above config
- On node4 remove the etcd-role with disable-etcd: true in config.yaml
- Restart the rke2 service

**Expected behavior:**
etcd-role should be removed from the node with just restarting service

**Actual behavior:**
we need to run killall to remove the etcd-role as pods are still running.

**Additional context / logs:**
```
Apr 15 18:41:20 ip-172-31-5-170 rke2[660]: time="2021-04-15T18:41:20Z" level=info msg="Starting rke2 v1.20.5-dev+dc9b0919 (dc9b091934def8e62396e441d90e217044fa3404)"
Apr 15 18:41:20 ip-172-31-5-170 rke2[660]: time="2021-04-15T18:41:20Z" level=info msg="Managed etcd cluster bootstrap already complete and initialized"
Apr 15 18:41:20 ip-172-31-5-170 rke2[660]: time="2021-04-15T18:41:20Z" level=warning msg="Error starting load balancer: listen tcp 127.0.0.1:2379: bind: address already in use"
Apr 15 18:41:20 ip-172-31-5-170 rke2[660]: time="2021-04-15T18:41:20Z" level=fatal msg="starting kubernetes: preparing server: listen tcp 127.0.0.1:2379: bind: address already in use"
Apr 15 18:41:20 ip-172-31-5-170 systemd[1]: rke2-server.service: Main process exited, code=exited, status=1/FAILURE
Apr 15 18:41:20 ip-172-31-5-170 systemd[1]: rke2-server.service: Unit entered failed state.
Apr 15 18:41:20 ip-172-31-5-170 systemd[1]: rke2-server.service: Failed with result 'exit-code'.
```

brandond:
[...]

PR #923, opened 2021-04-23
start kubelet and containerd on start to delete old pods

galal-hussein:
#### Proposed Changes ####

The pr will remove any manifest for a disabled item and then start kubelet and containerd to clean up old pods before starting the server

#### Types of Changes ####

bug fix

#### Linked Issues ####

https://github.com/rancher/rke2/issues/886

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -212,9 +212,9 @@ func removeOldPodManifests(dataDir string, disabledItems map[string]bool) error 
 		tmpAddress := filepath.Join(os.TempDir(), "containerd.sock")
 
 		// start containerd
-		go startContainerd(ctx, dataDir, containerdErr, tmpAddress, containerdCmd)
+		go startContainerd(dataDir, containerdErr, tmpAddress, containerdCmd)
 		// start kubelet
-		go startKubelet(ctx, dataDir, kubeletErr, tmpAddress, kubeletCmd)
+		go startKubelet(dataDir, kubeletErr, tmpAddress, kubeletCmd)
 
 		// check for any running containers from the disabled items list
 		go checkForRunningContainers(ctx, tmpAddress, disabledItems, kubeletErr, containerdErr)
@@ -250,7 +250,7 @@ func isCISMode(clx *cli.Context) bool {
 	return profile == CISProfile15 || profile == CISProfile16
 }
 
-func startKubelet(ctx context.Context, dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
+func startKubelet(dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
 	logrus.Infof("starting kubelet to clean up old static pods")
 	args := []string{
 		"--fail-swap-on=false",
@@ -266,7 +266,7 @@ func startKubelet(ctx context.Context, dataDir string, errChan chan error, tmpAd
 	errChan <- cmd.Run()
 }
 
-func startContainerd(ctx context.Context, dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
+func startContainerd(dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
 	args := []string{
 		"-a", tmpAddress,
 		"--root", filepath.Join(dataDir, "agent", "containerd"),
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -186,10 +186,11 @@ func podManifestsDir(dataDir string) string {
 }
 
 func removeOldPodManifests(dataDir string, disabledItems map[string]bool) error {
-	var (
-		kubeletStandAlone         bool
-		containerdErr, kubeletErr chan error
-	)
+	var kubeletStandAlone bool
+
+	kubeletErr := make(chan error)
+	containerdErr := make(chan error)
+
 	ctx := context.Background()
 
 	for component, disabled := range disabledItems {
@@ -215,24 +216,32 @@ func removeOldPodManifests(dataDir string, disabledItems map[string]bool) error 
 		go startKubelet(ctx, dataDir, kubeletErr, tmpAddress, kubeletCmd)
 
 		// check for any running containers from the disabled items list
-		go checkForRunningContainers(ctx, tmpAddress, disabledItems, kubeletCmd, containerdCmd)
+		go checkForRunningContainers(ctx, tmpAddress, disabledItems, kubeletErr, containerdErr)
 
 		go func() {
-			logrus.Infof("containerd 4 %v", containerdCmd)
+			// kill both temp processes after 5 minutes anyway
+			time.Sleep(5 * time.Minute)
+			logrus.Info("Timeout reached, exiting kubelet and containerd")
+			kubeletErr <- fmt.Errorf("timeout reached")
+			containerdErr <- fmt.Errorf("timeout reached")
+		}()
+
+		for {
+			time.Sleep(5 * time.Second)
 			select {
 			case err := <-kubeletErr:
 				logrus.Infof("kubelet Exited: %v, exiting Containerd", err)
 				// exits containerd if kubelet exits
 				containerdCmd.Process.Kill()
+				kubeletCmd.Process.Kill()
+
 			case err := <-containerdErr:
 				logrus.Infof("Containerd Exited: %v, exiting kubelet", err)
 				kubeletCmd.Process.Kill()
-			case <-time.After(5 * time.Minute):
-				// kill both temp process after 5 minutes anyway
-				kubeletCmd.Process.Kill()
 				containerdCmd.Process.Kill()
 			}
-		}()
+			break
+		}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/rke2/rke2.go
```diff
@@ -166,5 +176,145 @@ func setup(clx *cli.Context, cfg Config) error {
...
+}
+
+func checkForRunningContainers(ctx context.Context, tmpAddress string, disabledItems map[string]bool, kubeletCmd, containerdCmd *exec.Cmd) {
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+	for range ticker.C {
+		var gc bool
+		for item, disabled := range disabledItems {
+			if disabled {
+				isRunning, err := isContainerRunning(ctx, item, tmpAddress)
```
brandond, line 300: Each call to isContainerRunning results in a ListContainers call. While not especially expensive, it seems like it would be more efficient just call ListContainers once per ticker iteration, and then look at all the names in the list and see if they're in disabledItems.

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,145 @@ func setup(clx *cli.Context, cfg Config) error {
...
+			select {
+			case err := <-kubeletErr:
+				logrus.Infof("kubelet Exited: %v, exiting Containerd", err)
+				// exits containerd if kubelet exits
+				containerdCmd.Process.Kill()
+			case err := <-containerdErr:
+				logrus.Infof("Containerd Exited: %v, exiting kubelet", err)
+				kubeletCmd.Process.Kill()
+			case <-time.After(5 * time.Minute):
+				// kill both temp process after 5 minutes anyway
```
brandond, line 235: needs a logged error or warning indicating that the kill timed out

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,145 @@ func setup(clx *cli.Context, cfg Config) error {
...
+	}
+	cmd.Args = append(cmd.Args, args...)
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+
+	errChan <- cmd.Run()
+}
+
+func startContainerd(ctx context.Context, dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
+	// starting containerd first then kubelet
```
brandond, line 261: does this comment belong here?

brandond (changes requested):
Very cool! The approach looks good.

I had one nit about the efficiency of how containers are listed and searched for that I think should be addressed. The new functions also need doc comments.

pkg/rke2/rke2.go
galal-hussein: fixing

pkg/rke2/rke2.go
galal-hussein: fixing

pkg/rke2/rke2.go
galal-hussein: yeah that was before the code cleanup, deleting it now

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,145 @@ func setup(clx *cli.Context, cfg Config) error {
...
+		// start containerd
+		go startContainerd(ctx, dataDir, containerdErr, tmpAddress, containerdCmd)
+		// start kubelet
+		go startKubelet(ctx, dataDir, kubeletErr, tmpAddress, kubeletCmd)
+
+		// check for any running containers from the disabled items list
+		go checkForRunningContainers(ctx, tmpAddress, disabledItems, kubeletCmd, containerdCmd)
+
+		go func() {
+			logrus.Infof("containerd 4 %v", containerdCmd)
```
galal-hussein, line 225: removing old comment

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,145 @@ func setup(clx *cli.Context, cfg Config) error {
...
+		"--root", filepath.Join(dataDir, "agent", "containerd"),
+	}
+	cmd.Args = append(cmd.Args, args...)
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+
+	errChan <- cmd.Run()
+}
+
+func isContainerRunning(ctx context.Context, name string, address string) (bool, error) {
```
briandowns, line 273:
```suggestion
func isContainerRunning(ctx context.Context, name, address string) (bool, error) {
```

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,151 @@ func setup(clx *cli.Context, cfg Config) error {
...
+		"kube-apiserver":          disableAPIServer,
+		"kube-scheduler":          disableScheduler,
+		"kube-controller-manager": disableControllerManager,
+		"etcd":                    disableETCD,
+	}
+	return removeOldPodManifests(dataDir, disabledItems)
+}
+
+func podManifestsDir(dataDir string) string {
+	return filepath.Join(dataDir, "agent", defaultPodManifestPath)
```
brandond, line 189: Consolidate with this:
https://github.com/rancher/rke2/blob/c8283cd8e9bf7e8edba69c1916d24e9316c42f33/pkg/rke2/rke2.go#L130

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,151 @@ func setup(clx *cli.Context, cfg Config) error {
...
+	var kubeletStandAlone bool
+
+	kubeletErr := make(chan error)
+	containerdErr := make(chan error)
+
+	ctx := context.Background()
+
+	for component, disabled := range disabledItems {
+		if disabled {
+			manifestName := filepath.Join(podManifestsDir(dataDir), component+".yaml")
```
brandond, line 202: Don't call `podManfestsDir` inside the loop
```suggestion
	agentManifestsDir := podManifestsDir(dataDir)
	for component, disabled := range disabledItems {
		if disabled {
			manifestName := filepath.Join(agentManifestsDir, component+".yaml")
```

pkg/rke2/rke2.go
```diff
@@ -166,5 +176,151 @@ func setup(clx *cli.Context, cfg Config) error {
...
+		go startContainerd(ctx, dataDir, containerdErr, tmpAddress, containerdCmd)
+		// start kubelet
+		go startKubelet(ctx, dataDir, kubeletErr, tmpAddress, kubeletCmd)
+
+		// check for any running containers from the disabled items list
+		go checkForRunningContainers(ctx, tmpAddress, disabledItems, kubeletErr, containerdErr)
+
+		go func() {
+			// kill both temp processes after 5 minutes anyway
+			time.Sleep(5 * time.Minute)
```
brandond, line 227: Rather than running a sleeping goroutine to enforce the timeout, consider doing this this via `context.WithDeadline`
https://golang.org/pkg/context/#WithDeadline

brandond (changes requested):

pkg/rke2/rke2.go
```diff
@@ -166,5 +175,150 @@ func setup(clx *cli.Context, cfg Config) error {
...
+					break
+				}
+			}
+		}
+		if gc {
+			continue
+		}
+		// if all disabled items containers has been removed
+		// then terminate kubelet and containerd
+		containerdErr <- fmt.Errorf("static pods deleted exiting containerd")
```
brandond, line 320: Since you're wrapping these error channels up above, doesn't this result in a weird log message?
> kubelet Exited: static pods deleted exiting containerd, exiting Containerd

pkg/rke2/rke2.go
```diff
@@ -166,5 +175,150 @@ func setup(clx *cli.Context, cfg Config) error {
...
+		resp, err := c.ListContainers(ctx, &runtimeapi.ListContainersRequest{})
+		if err != nil {
+			logrus.Warnf("Failed to list containers: %v", err)
+			continue
+		}
+		var gc bool
+		for item, disabled := range disabledItems {
+			if disabled {
+				if isContainerRunning(ctx, item, resp) {
+					logrus.Infof("old static pod %s is running, waiting to delete", item)
```
brandond, line 309:
```suggestion
					logrus.Infof("Waiting for deletion of %s static pod", item)
```

pkg/rke2/rke2.go
```diff
@@ -166,5 +175,150 @@ func setup(clx *cli.Context, cfg Config) error {
...
+				containerdCmd.Process.Kill()
+			}
+			break
+		}
+	}
+
 	return nil
 }
+
+func startKubelet(ctx context.Context, dataDir string, errChan chan error, tmpAddress string, cmd *exec.Cmd) {
```
brandond, line 252: You're not actually using the Context that you pass in to this or startContainerd - if they don't use the context don't bother passing it in.

pkg/rke2/rke2.go
```diff
@@ -166,5 +175,150 @@ func setup(clx *cli.Context, cfg Config) error {
...
+				kubeletStandAlone = true
+				if err := os.Remove(manifestName); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	if kubeletStandAlone {
+		kubeletCmd := exec.Command("kubelet")
+		containerdCmd := exec.Command("containerd")
```
brandond, line 214: Use CommandContext, it will kill the processes for you when the context is cancelled.
```suggestion
		kubeletCmd := exec.CommandContext(ctx, "kubelet")
		containerdCmd := exec.CommandContext(ctx, "containerd")
```

brandond (changes requested):

briandowns (approved):

pkg/rke2/rke2.go
brandond: This needs addressed, other than that LGTM

brandond (changes requested):

PR #939, opened 2021-04-30
add list of registry mirrors

paraglade:
testing seeder cache cluster

### DIFF
--- a/scripts/test-helpers
+++ b/scripts/test-helpers
@@ -488,9 +488,15 @@ export -f provision-cluster
 
 provision-registry-proxy() {
     set -e -o pipefail
+    local image="docker.io/library/registry:2.7.1"
+    local prefix="docker-registry-"
+    local registries="docker.io:registry-1.docker.io k8s.gcr.io gcr.io quay.io ghcr.io"
     local registries_yaml="$TEST_DIR/registries.yaml"
+    local seeder_registry="[https://seeder.int.rancher.io:5000/]"
+    local docker_info_registry_config=$(docker info --format '{{ .RegistryConfig.Mirrors  }}')
 
-    cat << EOF > $registries_yaml
+    if [ "${seeder_registry}" == "${docker_info_registry_config}" ]; then
+        cat << EOF > $registries_yaml
 mirrors:
   docker.io:
     endpoint:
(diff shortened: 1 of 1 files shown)


brandond:
If we're going to do something like this it needs to be made conditional, and augment but NOT replace the current mirror creation behavior - right now the logic also works for local dev builds if you set the right variable.

scripts/test-helpers
```diff
@@ -492,35 +492,57 @@ provision-registry-proxy() {
...
-                -e "REGISTRY_HTTP_SECRET=shared-secret" \
-                -e "REGISTRY_PROXY_REMOTEURL=https://$registry_endpoint" \
-                -e "REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR=inmemory" \
-                -e "REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/var/lib/registry/$registry_name" \
-                $image
-        fi
-        echo -e "  $registry_name:\n    endpoint:\n    - http://172.17.0.1:$hostport" >> $registries_yaml
-    done
-
+    local seeder_registry="[https://seeder.int.rancher.io:5000/]"
```
brandond, line 495: Rather than embedding a specific registry configuration that is automatically enabled by introspecting the local docker instance, could we perhaps host the registry configuration somewhere on the seeder, and point at that with an environment variable? That way we only need to support a single variable (REGISTRIES_CONFIG_URL or something) that would be curl'd and used as the test's registries.yaml. This would also let you make modifications or take the mirrors down for maintenance without breaking our CI or needing to put in a PR.

brandond (changes requested):

scripts/test-helpers
```diff
@@ -492,35 +492,57 @@ provision-registry-proxy() {
...
+  gcr.io:
+    endpoint:
+    - https://seeder.int.rancher.io:5002
+  quay.io:
+    endpoint:
+    - https://seeder.int.rancher.io:5004
+  ghcr.io:
+    endpoint:
+    - https://seeder.int.rancher.io:5005
+EOF
```
brandond, line 516:
```suggestion
    if [ -z "${REGISTRY_CONFIG_URL:-}" ]; then
        curl -Lkso $registries_yaml ${REGISTRY_CONFIG_URL}
```

dereknola:
@paraglade Please respond to changes requested by brandond or close this PR as out-of-date.

briandowns:
@paraglade Is this PR still in progress or can it be closed?

oxr463:
> @paraglade Is this PR still in progress or can it be closed?

I don't think this should be closed. This seems to be related to the build timeouts we're seeing on Drone Publish.

PR #954, opened 2021-05-05
Integrate Multus in rke2

manuelbuil:
#### Proposed Changes ####

Allow multus to be deployed using the --cni flag

#### Types of Changes ####

New feature

#### Verification ####

--cni=$CNI_plugin+multus (e.g. canal+multus)

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -21,11 +21,10 @@ var (
 	config = rke2.Config{}
 
 	serverFlag = []cli.Flag{
-		&cli.StringFlag{
+		&cli.StringSliceFlag{
 			Name:   "cni",
-			Usage:  "(networking) CNI Plugin to deploy, one of none, " + strings.Join(CNIItems, ", "),
+			Usage:  "(networking) CNI Plugins to deploy, one of none, " + strings.Join(CNIItems, ", ") + "; optionally with multus as the first value to enable the multus meta-plugin (default: canal)",
 			EnvVar: "RKE2_CNI",
-			Value:  "canal",
 		},
 	}
 
@@ -142,13 +141,38 @@ func ServerRun(clx *cli.Context) error {
 }
 
 func validateCNI(clx *cli.Context) {
-	cni := clx.String("cni")
-	switch {
-	case cni == "none":
+	cnis := []string{}
+	for _, cni := range clx.StringSlice("cni") {
+		for _, v := range strings.Split(cni, ",") {
+			cnis = append(cnis, v)
+		}
+	}
+
+	switch len(cnis) {
+	case 0:
+		cnis = append(cnis, "canal")
 		fallthrough
-	case slice.ContainsString(CNIItems, cni):
+	case 1:
+		if cnis[0] == "multus" {
+			logrus.Fatal("invalid value provided for --cni flag: multus must be used alongside another primary cni selection")
+		}
+		clx.Set("disable", "rke2-multus")
+	case 2:
+		if cnis[0] == "multus" {
+			cnis = cnis[:1]
+		} else {
+			logrus.Fatal("invalid values provided for --cni flag: may only provide multiple values if multus is the first value")
+		}
+	default:
+		logrus.Fatal("invalid values provided for --cni flag: may not provide more than two values")
+	}
+
+	switch {
+	case cnis[0] == "none":
+		fallthrough
+	case slice.ContainsString(CNIItems, cnis[0]):
 		for _, d := range CNIItems {
-			if cni != d {
+			if cnis[0] != d {
 				clx.Set("disable", "rke2-"+d)
 				clx.Set("disable", "rke2-"+d+"-crd")
 			}
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -58,6 +58,17 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-vsphere.txt
     ${REGISTRY}/rancher/mirrored-k8scsi-livenessprobe:v2.1.0
 EOF
 
+xargs -n1 -t docker image pull --quiet << EOF > build/images-multus.txt
+    ${REGISTRY}/rancher/hardened-multus-cni:v3.7.1-build20210416
+    ${REGISTRY}/rancher/hardened-sriov-network-operator:v1.0.0-build20210429
+    ${REGISTRY}/rancher/hardened-sriov-network-config-daemon:v1.0.0-build20210429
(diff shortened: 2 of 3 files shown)


pkg/cli/cmds/server.go
```diff
@@ -23,7 +23,7 @@ var (
 	serverFlag = []cli.Flag{
 		&cli.StringFlag{
 			Name:   "cni",
-			Usage:  "(networking) CNI Plugin to deploy, one of none, " + strings.Join(CNIItems, ", "),
+			Usage:  "(networking) CNI Plugin to deploy, one of none, " + strings.Join(CNIItems, ", ") + ". Add the suffix +multus to the CNI plugin to install multus as well",
```
brandond, line 26:
```suggestion
			Usage:  "(networking) CNI Plugin to deploy, one of none, " + strings.Join(CNIItems, ", ") + "; +multus may be appended to enable the Multus meta-plugin.",
```

pkg/cli/cmds/server.go
```diff
@@ -143,6 +143,10 @@ func ServerRun(clx *cli.Context) error {
 
 func validateCNI(clx *cli.Context) {
 	cni := clx.String("cni")
+	if !strings.Contains(cni, "multus") {
+		clx.Set("disable", "rke2-multus")
+	}
+	cni = strings.Split(cni, "+")[0]
```
brandond, line 149: Lets be a little more picky about the suffix
```suggestion
	if strings.HasSuffix(cni, "+multus") {
		cni = strings.TrimSuffix(cni, "+multus")
	} else {
		clx.Set("disable", "rke2-multus")
	}
```

brandond (approved):

brandond (changes requested):
LGTM once we get the additional image list created and tarred up

dweomer:
I realize that this may have already passed design but I am not a fan of the `---cni=calico+multus` formulation. Would it not be simpler to make it such that multus is off by default unless it is enabled via `--enable rke2-multus` (which implies the addition of the converse of the --disable flag)?

brandond:
> Would it not be simpler to make it such that multus is off by default unless it is enabled via `--enable rke2-multus`

We don't have a --enable flag, and I would prefer to avoid creating yet another CLI flag and list of things to enable (un-disable). The current logic that we've built for CNI and vSphere retains the convention of there only being a disable list, to which things are added if not wanted. I also kind of like the UX of it, in terms of asking for 'something plus multus' which is the only way it can be deployed.

I'm glad to put it up for a vote though - cc @cjellick @ibuildthecloud

scripts/build-images
```diff
@@ -58,6 +58,17 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-vsphere.txt
     ${REGISTRY}/rancher/mirrored-k8scsi-livenessprobe:v2.1.0
 EOF
 
+xargs -n1 -t docker image pull --quiet << EOF >> build/images-multus.txt
```
brandond, line 61: Otherwise we get another set of images appended every time we build locally
```suggestion
xargs -n1 -t docker image pull --quiet << EOF > build/images-multus.txt
```

brandond (changes requested):

scripts/build-images
manuelbuil: Good point! Thanks

pkg/cli/cmds/server.go
```diff
@@ -143,6 +143,11 @@ func ServerRun(clx *cli.Context) error {
 
 func validateCNI(clx *cli.Context) {
 	cni := clx.String("cni")
+	if strings.HasSuffix(cni, "+multus") {
```
cjellick, line 146: `multus,primary`

cjellick: brad to open an issue for helm-controller ordering bit for future work

brandond: Ordering issue: https://github.com/rancher/rke2/issues/960

@manuelbuil let me know if you need pointers on converting this over to a StringSliceFlag. StringSliceFlags have the unfortunate downside of not clearing the default value from the list if you provide values on the CLI, so you have to hack around it not providing a default value in the flag, adding the default text to the usage string, and manually setting the default if the list is empty.

Probably looking at something like:
```go
cnis := []string{}
for _, cni := range clx.StringSlice("cni") {
  for _, v := range strings.Split(",") {
    cnis = append(cnis, v)
  }
}

switch len(cnis) {
case 0:
  cnis = append(cnis, "canal")
  clx.Set("disable", "rke2-multus")
case 1:
  if cnis[0] == "multus" {
    logrus.Fatal("invalid value provided for --cni flag: multus must be used alongside another primary cni selection")
  }
case 2:
  if cnis[0] == "multus" {
    cnis = cnis[:1]
  } else {
    logrus.Fatal("invalid values provided for --cni flag: may only provide multiple values if multus is the first value")
  }
default:
  logrus.Fatal("invalid values provided for --cni flag: may not provide more than two values")
}  
```
and then switch the bit down below to operating on `cnis[0]` instead of `cni`

brandond (changes requested):

pkg/cli/cmds/server.go
manuelbuil: Thanks!

pkg/cli/cmds/server.go
```diff
@@ -142,13 +141,37 @@ func ServerRun(clx *cli.Context) error {
...
+	}
+
+	switch len(cnis) {
+	case 0:
+		cnis = append(cnis, "canal")
+		clx.Set("disable", "rke2-multus")
+	case 1:
+		if cnis[0] == "multus" {
+			logrus.Fatal("invalid value provided for --cni flag: multus must be used alongside another primary cni selection")
+		}
```
brandond, line 158: Whoops, also need to disable multus on single arg.
```suggestion
		fallthrough
	case 1:
		if cnis[0] == "multus" {
			logrus.Fatal("invalid value provided for --cni flag: multus must be used alongside another primary cni selection")
		}
		clx.Set("disable", "rke2-multus")
```

scripts/build-images
brandond: Still needs changed

brandond (changes requested):
The image list redirection still needs to be fixed, other than that looks good.

manuelbuil:
> The image list redirection still needs to be fixed, other than that looks good.

Weird, I fixed it in one of the updates. Thanks for pointing this out (again)

brandond (approved):

PR #957, opened 2021-05-05
Fix remove old pod manfiests

galal-hussein:
#### Proposed Changes ####

This change will make sure that all static pods are removed when any of the disable flags or cluster reset flag is passed, the pr will make sure components are removed before attempting to start the server process

#### Types of Changes ####

bug fix

#### Linked Issues ####
https://github.com/rancher/rke2/issues/886

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -200,24 +200,25 @@ func removeOldPodManifests(dataDir string, disabledItems map[string]bool, cluste
 			manifestName := filepath.Join(manifestDir, component+".yaml")
 			if _, err := os.Stat(manifestName); err == nil {
 				kubeletStandAlone = true
-				if err := os.Remove(manifestName); err != nil {
-					return err
-				}
 			}
 		}
 	}
+
 	if clusterReset {
 		// deleting old etcd if cluster reset is passed
 		manifestName := filepath.Join(manifestDir, "etcd.yaml")
 		if _, err := os.Stat(manifestName); err == nil {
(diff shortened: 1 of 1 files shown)


brandond (approved):
Just to confirm - this essentially mimics some of the behavior from K3s, wherein all the control-plane components get killed on shutdown so that they can come up cleanly the next restart.

Since RKE2 runs them as static pods, we need to instead run through a short cleanup cycle on startup wherein we have the kubelet terminate all the static pods **when disabling components that were previously enabled** so that everything can be brought back up in the correct order.

dweomer (approved):

pkg/rke2/rke2.go
```diff
@@ -316,13 +317,11 @@ func checkForRunningContainers(ctx context.Context, tmpAddress string, disabledI
...
 		}
 		var gc bool
-		for item, disabled := range disabledItems {
-			if disabled {
-				if isContainerRunning(ctx, item, resp) {
-					logrus.Infof("Waiting for deletion of %s static pod", item)
-					gc = true
-					break
-				}
+		for item, _ := range disabledItems {
```
brandond, line 320: @galal-hussein I don't see where we're adding anything to the disabledItems list - if we're shutting down everything shouldn't we be populating this with all the basenames of manifests in the static pods dir?

galal-hussein: The disabled item list is already populated with all names, its actually a map of string booleans, right now we just check all of them and delete all manifests

Issue #959, opened 2021-05-05
Use --system-default-registry flag from K3s


brandond:
* Pull through work from k3s-io/k3s#3281
* Handle `--system-default-registry` as flag inherited from K3s, instead of as native RKE2 flag

rancher-max:
#### Validated in v1.21.1-alpha5+rke2r1

- system-default-registry continues to work in rke2 in airgap setups

Issue #958, opened 2021-05-05
Use rancher/wharfie code for bootstrap


brandond:
* Pull through work from k3s-io/k3s#3280.
* Remove code from RKE2 that has been moved to rancher/wharfie; use wharfie instead.
* Use `--image-credential-provider` flags to enable credential provider support when pulling bootstrap images.

rancher-max:
This is validated on v1.21.1-alpha5+rke2r1

- All mirror private registry functionality still working as expected
- system-default-registry still working as expected
- Setting airgap-extra-registry still working as expected
- Airgap installs via tarball still working as expected

No new functionality to report here so steps remain the same and leaving them out here for brevity.

Note the `image-credential-provider` functionality encounters the same errors as noticed in the linked k3s issue. Closing this in favor of https://github.com/rancher/rke2/issues/929 for those.

PR #971, opened 2021-05-11
Use rancher/wharfie code shared with k3s for for bootstrap

brandond:
#### Proposed Changes ####

* Remove bootstrap code that has been moved to rancher/wharfie
* Support kubelet image credential plugins (both for bootstrap pull, and for kubelet)
* Use --system-default-registry flag from k3s

Removal of --system-default-registry from agent flags should be called out in release notes, and removed from docs.

#### Types of Changes ####

* Bootstrap
* Private registry

#### Verification ####

* Set --system-default-registry on server, note that agents use same setting
* Configure kubelet image credential plugins, note that bootstrap image and core images can be pulled from ECR if ECR is specified as --system-default-registry

#### Linked Issues ####

For #958 and #959


dweomer (approved):

briandowns (approved):

MonzElmasry:
Includes a fix for https://github.com/rancher/rke2/issues/1019

PR #1012, opened 2021-05-19
Update cilium's chart

manuelbuil:

#### Proposed Changes ####

There is a bug in Cilium chart that gets fixed with this PR ==> https://github.com/rancher/rke2-charts/pull/87

This PR in rke2 points to that fix

#### Types of Changes ####

Bugfix

#### Verification ####

--cni cilium

#### Linked Issues ####

https://github.com/rancher/rke2-charts/pull/87

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,7 +129,7 @@ ARG CHART_REPO="https://rke2-charts.rancher.io"
 ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
 RUN echo ${CACHEBUST}>/dev/null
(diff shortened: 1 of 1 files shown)


brandond:
@manuelbuil I think we don't need this - and the chart change should probably be rolled back as well. The input validation in the chart caught a legitimate issue with helm-controller's handling of empty strings.

manuelbuil:
> @manuelbuil I think we don't need this - and the chart change should probably be rolled back as well. The input validation in the chart caught a legitimate issue with helm-controller's handling of empty strings: [k3s-io/helm-controller#98](https://github.com/k3s-io/helm-controller/pull/98)

Agree, I was too fast fixing this here

PR #1055, opened 2021-05-25
Initial build working on Windows.

phillipsj:
#### Proposed Changes ####

These changes breaks up the rke2.go file into shared, linux, and windows versions. This was required to isolate the k3s dependencies that would not build on Windows.

There is no impact to documentation.

#### Types of Changes ####

There are no breaking changes that should occur from this change.

#### Verification ####

Builds both run successfully on Linux and Windows.

#### Linked Issues ####

This issue is the origin of the change: rancher/windows#52

#### Further Comments ####

This change is in preparation for adding Windows support to RKE2. We are attempting to keep the Windows experience as close as possible as the Linux experience.


luthermonson:
just testing the stubs for rke2.go and the build/merge process.

PR #1058, opened 2021-05-25
Ensure kubelet and containerd are in path for temporary startup

brandond:
#### Proposed Changes ####

Ensure that we are properly setting up the paths and environment for the kubelet and containerd processes that are temporarily started to clean up static pods during reconfiguration (cluster reset, role change, etc).

Because we now defer runtime bootstrap (which has a side effect of modifying the PATH environment variable to include the runtime bin dir), we need to use the bin dir symlink at this stage of the startup. We should also display the kubelet and containerd logs so that we can see what's going on as we wait for static pods to get cleaned up.

This also closes a gap that existed around checking for static pod termination by examining the containerd container names. Container names are not unique; we need to look at the kubernetes labels instead.

Long term we should move this into the static pod executor module, as we've copy-pasted too much code out of there (and out of K3s), but this will do for now.

#### Types of Changes ####

cluster reset

#### Linked Issues ####

For #1059

#### Further Comments ####

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -2,13 +2,14 @@ package rke2
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	containerdk3s "github.com/rancher/k3s/pkg/agent/containerd"
 	"github.com/rancher/rke2/pkg/controllers/cisnetworkpolicy"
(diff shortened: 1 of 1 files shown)


pkg/rke2/rke2.go
```diff
@@ -20,16 +21,20 @@ import (
...
 	rawServer "github.com/rancher/k3s/pkg/server"
 	"github.com/rancher/rke2/pkg/cli/defaults"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/rancher/rke2/pkg/podexecutor"
 	"github.com/urfave/cli"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
+var containerdSock = "/run/k3s/containerd/containerd.sock"
```
briandowns, line 36: Can this be a const?

brandond: Sure. Ideally we would have exposed it from K3s somewhere (and not hardcoded k3s in it) but this'll do for now.

briandowns (approved):

galal-hussein (approved):
LGTM, Thanks for fixing the containers lookup thing, I was struggling to find a proper way to check it

dweomer (approved):

PR #1086, opened 2021-06-01
[docs] Stop RKE2 before performing a snapshot restore

dkeightley:
#### Proposed Changes ####

Add suggested wording and command to stop `rke2-server` before starting a snapshot restore

#### Types of Changes ####

docs

### DIFF
--- a/docs/backup_restore.md
+++ b/docs/backup_restore.md
@@ -30,9 +30,10 @@ rke2 server --cluster-reset
 
 When RKE2 is restored from backup, the old data directory will be moved to `/var/lib/rancher/rke2/server/db/etcd-old-%date%/`. Then RKE2 will attempt to restore the snapshot by creating a new data directory, then starting etcd with a new RKE2 cluster with one etcd member.
 
-To restore the cluster from backup, run RKE2 with the `--cluster-reset` option, with the `--cluster-reset-restore-path` also given:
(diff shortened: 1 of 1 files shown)


Oats87 (approved):

docs/backup_restore.md
```diff
@@ -30,9 +30,10 @@ rke2 server --cluster-reset
 
 When RKE2 is restored from backup, the old data directory will be moved to `/var/lib/rancher/rke2/server/db/etcd-old-%date%/`. Then RKE2 will attempt to restore the snapshot by creating a new data directory, then starting etcd with a new RKE2 cluster with one etcd member.
 
-To restore the cluster from backup, run RKE2 with the `--cluster-reset` option, with the `--cluster-reset-restore-path` also given:
+To restore the cluster from backup, first you need to stop RKE2 service if its enabled via systemd. Once stopped, run RKE2 with the `--cluster-reset` option, with the `--cluster-reset-restore-path` also given:
```
briandowns, line 33: I would probably include also that the services started by RKE2 should be stopped as well, e.g. kube-apiserver, etc. This can be done via the `rke2-killall.s` script.

brandond: @briandowns the restore code should now handle stopping all the control-plane components before running the restore; running killall will terminate workload pods as well which may not be desired.

briandowns (approved):

brandond (approved):

PR #1109, opened 2021-06-08
The CNI parameter got introduced in 1.20.7 and not starting with 1.21

Martin-Weiss:
Not sure why we write 1.21 here while 1.21 is not yet released and why we write 1.21 for adding the cni parameter while it was added in 1.20.7.
I am also missing the details on how to deploy cilium in 1.20.6, now - so I assume further adjustments are required?

### DIFF
--- a/docs/install/network_options.md
+++ b/docs/install/network_options.md
@@ -32,7 +32,7 @@ For more information about the full options of the Canal config please refer to 
 
 # Using Cilium or Calico instead of Canal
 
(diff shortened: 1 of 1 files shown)


brandond:
The CNI flag is was backported to 1.20 as part of some other enhancement work, but not fully supported until 1.21. We will continue to document it as available and supported as part of the 1.21 release.

Martin-Weiss:
> The CNI flag is was backported to 1.20 as part of some other enhancement work, but is not fully supported until 1.21. We will continue to document it as available and supported as part of the 1.21 release.

Where can I find the information if / how far this is supported in 1.20.7?

Basically I stepped into the trap realizing that the way how to deploy cilium is different in 1.20.7 vs. 1.20.6 and before.. (the way of 1.20.6 does not work anymore).

I believe in this case we need the documentation for "cilium" with 1.20.7 vs 1.20.6 vs. 1.21. 
I am also not sure why we have documented how to configure it in 1.21 even though 1.21 is not released, yet... and it is difficult to find the documentation for pre-1.21...

dereknola:
Closing this PR as official support for replacing Canal was added in 1.21 and 1.21 in now widely available.

Issue #989, opened 2021-05-13
Add systemd notify support 


brandond:
From @ibuildthecloud 

The rke2 systemd units should use type=notify, and not notify systemd until critical components return OK to the readyz endpoint. Critical components would be:
* kubelet for agents
* apiserver and controller-manager for control-plane nodes
* etcd for etcd nodes

cjellick:
Second related PR should make it in today.

briandowns:
PR https://github.com/k3s-io/k3s/pull/3430 merged.

ShylajaDevadiga:
Validated using commit id. Waiting on rc
```
rke2 version v1.21.1-dev+178370a3
```

ShylajaDevadiga:
Validated using rc, v1.21.2-rc1+rke2r1, readyz check passed before the returning to the prompt using API health endpoints
```
$ /var/lib/rancher/rke2/data/v1.21.2-rc1-rke2r1-776c38e6c32a/bin/kubectl get --raw=/readyz?verbose
[+]ping ok
[+]log ok
[+]etcd ok
[+]informer-sync ok
[+]poststarthook/start-kube-apiserver-admission-initializer ok
[+]poststarthook/generic-apiserver-start-informers ok
[+]poststarthook/priority-and-fairness-config-consumer ok
[+]poststarthook/priority-and-fairness-filter ok
[+]poststarthook/start-apiextensions-informers ok
[+]poststarthook/start-apiextensions-controllers ok
[+]poststarthook/crd-informer-synced ok
[+]poststarthook/bootstrap-controller ok
[+]poststarthook/rbac/bootstrap-roles ok
[+]poststarthook/scheduling/bootstrap-system-priority-classes ok
[+]poststarthook/priority-and-fairness-config-producer ok
[+]poststarthook/start-cluster-authentication-info-controller ok
[+]poststarthook/aggregator-reload-proxy-client-cert ok
[+]poststarthook/start-kube-aggregator-informers ok
[+]poststarthook/apiservice-registration-controller ok
[+]poststarthook/apiservice-status-available-controller ok
[+]poststarthook/kube-apiserver-autoregistration ok
[+]autoregister-completion ok
[+]poststarthook/apiservice-openapi-controller ok
[+]shutdown ok
readyz check passed

```

PR #1122, opened 2021-06-09
update systemd service files to include type notify

briandowns:

#### Proposed Changes ####

Update type to be notify so that systemd will only be notified of rke2 being up when it writes to the systemd notify socket.

#### Linked Issues ####

#989

### DIFF
--- a/bundle/lib/systemd/system/rke2-agent.service
+++ b/bundle/lib/systemd/system/rke2-agent.service
@@ -9,6 +9,7 @@ Conflicts=rke2-server.service
 WantedBy=multi-user.target
 
 [Service]
+Type=notify
 EnvironmentFile=-/etc/default/%N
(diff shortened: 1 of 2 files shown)


Oats87 (approved):
This LGTM; however, is this going to cause a change in behavior for our users where `systemctl start rke2-server` is going to start hanging for a decent period of time?

briandowns:
@Oats87 This does change the expected behavior. On server startup, it will wait until the server (supervisor components) is up and running and will do so for the agent as well but will require a little reworking of the k3s agent code to do so. I'm currently working on the agent code now.

brandond:
Is this safe to merge without the agent code merged yet, or will it just hang on startup?

briandowns:
No hangs as far as I noticed in testing.

brandond (approved):

Issue #1053, opened 2021-05-25
Pod communication not working on RHEL 8.4 on cloud providers


Oats87:
**Environmental Info:**
RKE2 Version: v1.20.7+rke2r1

Node(s) CPU architecture, OS, and Version: 
`Linux hostname 4.18.0-305.el8.x86_64 #1 SMP Thu Apr 29 08:54:30 EDT 2021 x86_64 x86_64 x86_64 GNU/Linux`
`Red Hat Enterprise Linux release 8.4 (Ootpa)`

Cluster Configuration: Single node.

**Describe the bug:**
Pod communication is not possible when running on a RHEL 8.4 host on AWS, regardless of selinux status.

**Steps To Reproduce:**
```
mkdir -p /etc/rancher/rke2/
systemctl stop firewalld
cat > /etc/rancher/rke2/config.yaml <<EOF
selinux: true
write-kubeconfig-mode: "0644"
EOF
curl -sfL https://get.rke2.io --output install.sh
chmod +x install.sh
INSTALL_RKE2_CHANNEL=stable ./install.sh
systemctl enable rke2-server.service --now
```

Wait for RKE2 server to start and pods to start, and try to `ping` any pod IP in the `10.42.x.x` range.

**Expected behavior:**
Pods can talk.

**Actual behavior:**
Pods don't talk.

**Additional context / logs:**
Looking at a `tcpdump` from within the container, you can see the corresponding ICMP request/reply, so the pod is definitely receiving the traffic. It seems to be getting lost heading back to the host network namespace.

This does not occur on VMs running in our test environment that are running RHEL 8.4

selinux status does not seem to affect the issue.

Oats87:
Investigating this with @manuelbuil we found that we did not have issues on RHEL 8.3 AMI.

We also did not have any issues with our internal lab environment running 8.3 and 8.4.

We discovered a difference in the two AMIs, that there are new services (`nm-cloud-setup.service`, `nm-cloud-setup.timer`) that are enabled on the nodes. Disabling these i.e. `systemctl disable nm-cloud-setup.service nm-cloud-setup.timer` and rebooting the node restored connectivity on the node.

More work will need to be done to investigate why `nm-cloud-setup` is causing us issues when it is running.

brandond:
It's Always NetworkManager.

manuelbuil:
[...]

PR #1149, opened 2021-06-16
Fail to deploy if nm-setup-cloud is enabled

manuelbuil:

#### Proposed Changes ####

Even though we are documenting that nm-cloud-setup service breaks the rke2 cluster, it is a good idea to verify this when starting the rke2 service and not allow it to start if the user forgot to disable that service. If the OS is not running systemd, it will deploy rke2.

This PR also documents this as a known issue

#### Types of Changes ####

New feature. Prevents rke2 server or agent to start if nm-setup-cloud exists

#### Verification ####

Deploy RHEL8.4 which use nm-cloud-setup by default and run rke2. It should fail to start with the following log in journactl:
```
sh[1118717]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
sh[1118717]: + exit 1
```
#### Linked Issues ####

https://github.com/rancher/rke2/issues/1053

### DIFF
--- a/bundle/lib/systemd/system/rke2-agent.service
+++ b/bundle/lib/systemd/system/rke2-agent.service
@@ -22,6 +22,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
 ExecStart=/usr/local/bin/rke2 agent
--- a/bundle/lib/systemd/system/rke2-server.service
+++ b/bundle/lib/systemd/system/rke2-server.service
@@ -22,6 +22,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
 ExecStartPre=-/sbin/modprobe br_netfilter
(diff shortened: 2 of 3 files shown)


bundle/lib/systemd/system/rke2-agent.service
```diff
@@ -22,6 +22,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '/usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service && exit 1 || exit 0'
```
brandond, line 25:
```suggestion
ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
```

bundle/lib/systemd/system/rke2-server.service
```diff
@@ -22,6 +22,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '/usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service && exit 1 || exit 0'
```
brandond, line 25:
```suggestion
ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
```

brandond (changes requested):
LGTM, but I think you can negate the test more simply

bundle/lib/systemd/system/rke2-agent.service
manuelbuil: great idea! Thanks Brad

brandond (approved):

briandowns (approved):

Issue #1164, opened 2021-06-19
 bug: rke2/docs/install/quickstart.md link to known issues for networking results in 404


FruityWelsh:
**Describe the bug:**
 bug: rke2/docs/install/quickstart.md link to known issues for networking results in 404
 the link is https://docs.rke2.io/known_issues/#networkmanager but this 404s
 the link to known issues that works is https://docs.rke2.io/known_issues.html

cjellick:
just a change to the docs site, so i think we can close

cjellick:
Nah, QA can validate this I think.

ShylajaDevadiga:
Validated https://docs.rke2.io/known_issues/ link works and there are no broken links on the main page.

PR #1165, opened 2021-06-21
update link to known issues network manager

briandowns:

#### Proposed Changes ####

Updates the version of mkdocs in the Dockerfile to be 7.1.7 which is the version before "latest".

#### Linked Issues ####

#1164

### DIFF
--- a/Dockerfile.docs
+++ b/Dockerfile.docs
@@ -1,4 +1,4 @@
-FROM squidfunk/mkdocs-material
+FROM squidfunk/mkdocs-material:7.1.7
 RUN pip install mkdocs-markdownextradata-plugin
 RUN apk add -U git openssh
 


Oats87 (dismissed):
This shouldn't be necessary -- the links all worked before. Did something change in the way we publish our docs that broke these links? The link right above this one is also broken.

briandowns:
Not aware of any changes but updated the other broken link as well.

brandond (approved):

briandowns:
This issue seems to be related to the latest release of mkdocs, 7.1.8. I've put the previous, working version tag in the Dockerfile that generates the docs.

brandond:
Looks like the mkdocs version bump is the culprit:
```
brandond@dev01:~$ docker run --rm -it squidfunk/mkdocs-material:7.1.7 --version
mkdocs, version 1.1.2 from /usr/local/lib/python3.9/site-packages/mkdocs (Python 3.9)
brandond@dev01:~$ docker run --rm -it squidfunk/mkdocs-material:7.1.8 --version
mkdocs, version 1.2.1 from /usr/local/lib/python3.9/site-packages/mkdocs (Python 3.9)
```

https://www.mkdocs.org/about/release-notes/#version-12-2021-06-04
Specifically:
> The use_directory_urls configuration option will be forced to false if site_url is set to an emtpy string. In that case, if use_directory_urls is not explicitly set to false, a warning will be issued (https://github.com/mkdocs/mkdocs/issues/2189).

Rather than downgrading, you might try just setting `use_directory_urls: true` and see if it gets the old behavior.

briandowns:
My intent was to bring this back to the expected version and behavior and evaluate how we wanted to move forward whether it be this additional change or update the links themselves.

brandond (approved):

Oats87 (approved):

Issue #1166, opened 2021-06-21
Multi-OS Support for RKE2 runtime and binary builds (Adding Windows)


rosskirkpat:
Issue to track adding Windows (multi-OS) support to RKE2

1.21 backport issue: https://github.com/rancher/rke2/issues/1299
Manifest support for runtime images: https://github.com/rancher/rke2/issues/1127

rosskirkpat:
cc: @sirredbeard @luthermonson @phillipsj

rancher-max:
Closing in favor of https://github.com/rancher/rke2/issues/1289#issuecomment-926316897

briandowns:
Validated on master.

PR #1168, opened 2021-06-21
 Multi-OS Support for RKE2 runtime and binary builds (Adding Windows)

rosskirkpat:
#### Proposed Changes ####
- Add windows binary build scripts via cross platform builds (no Windows build hosts are required at this time)
- Add windows runtime image build
- add a dockerfile for the windows runtime and rename the existing (linux) dockerfile to reflect multi-OS state of rke2
- rename `build-binary` to `build-linux-binary` and updated makefile to reflect multi-OS state of rke2 
- updated relevant docs to reference `Dockerfile.linux` usage

Big Picture: These changes are required for the RKE2 Windows release as part of 2.6.0

Documentation will need updating to indicate that RKE2 will support Windows on 2.6.0/1.21.x. We will also need to document the guard rails for initial windows support on rke2.

Additionally, Air-gap will work for rke2 on windows in 2.6.0 but it will require it's own script to pull the images on a Windows host. The pause images for windows would require running 3 Windows drone nodes (1809/2004/20H2) and we decided that documenting this and providing a script for pulling the Windows air-gap images would meet requirements for 2.6.0

#### Types of Changes ####
New Feature

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1166


scripts/package-binary
```diff
@@ -8,3 +8,4 @@ source ./scripts/version.sh
 mkdir -p dist/artifacts
 
 install -s bin/${PROG} dist/artifacts/${RELEASE}
+install -s bin/${PROG}.exe dist/artifacts/${RELEASE}
```
briandowns, line 11: Should this have some guard to only run on Windows?

scripts/build-windows-binary
```diff
@@ -0,0 +1,61 @@
...
+  CGO_ENABLED=1 CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc \
+  go build \
+    -tags "${GO_BUILDTAGS}" \
+    ${GO_GCFLAGS} ${GO_BUILD_FLAGS} \
+    -o bin/${PROG}.exe \
+    -ldflags "${GO_LDFLAGS} ${VERSION_FLAGS}" \
+    ${GO_TAGS}
+
+# assert that rke2 is fully statically linked
+#if type -a go-assert-static.sh >/dev/null 2>&1; then
```
briandowns, line 54: Should this be commented out?

scripts/build-images
```diff
@@ -27,7 +27,7 @@ EOF
 xargs -n1 -t docker image pull --quiet << EOF > build/images-canal.txt
     ${REGISTRY}/rancher/hardened-calico:v3.19.1-build20210611
     ${REGISTRY}/rancher/hardened-flannel:v0.14.0-build20210611
-EOF
+    EOF
```
dweomer, line 30: `EOF` on a heredoc should not be tabbed
```suggestion
EOF
```

scripts/build-windows-binary
```diff
@@ -0,0 +1,61 @@
...
+"
+
+#STATIC_FLAGS='-extldflags "-static -Wl,--fatal-warnings"'
+### ON GLIBC SYSTEMS YOU WILL SEE LINKER WARNINGS #####################################################################
+# github.com/rancher/rke2
+#/usr/bin/ld: /tmp/go-link-743463534/000027.o: in function `nvmlInit_dl':
+#/home/jacob/go/pkg/mod/github.com/mindprince/gonvml@v0.0.0-20190828220739-9ebdce4bb989/bindings.go:152: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
+### THESE CAN BE SAFELY IGNORED UNLESS YOU ARE TRYING TO LEVERAGE NVIDIA METRICS FROM CADVISOR AND YOU HAPPEN TO BE ###
+### RUNNING RKE2 ON A SYSTEM OTHER THAN THE ONE WHICH IS WAS BUILT THAT ALSO HAS AN INCOMPATIBLE GLIBC ################
+#STATIC_FLAGS='-extldflags "-static"'
```
dweomer, line 40: Better to drop this comment block completely.

scripts/build-windows-binary
```diff
@@ -0,0 +1,61 @@
...
+
+# assert that rke2 is fully statically linked
+#if type -a go-assert-static.sh >/dev/null 2>&1; then
+#    go-assert-static.sh bin/${PROG}
+#fi
+
+# assert that rke2 is linked to goboring
+#if type -a go-assert-boring.sh >/dev/null 2>&1; then
+#    go-assert-boring.sh bin/${PROG}
+#fi
```
dweomer, line 61: Better to drop these commented blocks completely.

scripts/package-binary
dweomer: This one is weird to me. The reason for my original use of `install` is that I wanted correct file mode/perms and symbol stripping in a single operation. Does this even work on Windows?

scripts/build-windows-binary
dweomer: goboring is linux-only afaik

scripts/build-windows-binary
```diff
@@ -0,0 +1,61 @@
...
+GO_LDFLAGS="${STATIC_FLAGS} ${EXTRA_LDFLAGS}"
+echo ${DEBUG_GO_GCFLAGS}
+GOOS=windows \
+  CGO_ENABLED=1 CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc \
+  go build \
+    -tags "${GO_BUILDTAGS}" \
+    ${GO_GCFLAGS} ${GO_BUILD_FLAGS} \
+    -o bin/${PROG}.exe \
+    -ldflags "${GO_LDFLAGS} ${VERSION_FLAGS}" \
+    ${GO_TAGS}
```
dweomer, line 51: It would be nice if the specialization of `scripts/build-binary` for windows wasn't a copied-but-slightly-modified version. The more we copy the more we have to maintain. Can this script be generalizaed to support windows and linux?

scripts/build-windows-binary
rosskirkpat: @dweomer that's correct. There is no alternative for goboring on Windows either and was not part of our requirements for Windows on RKE2

scripts/build-windows-binary
rosskirkpat: Removed comment block

scripts/build-windows-binary
rosskirkpat: Removed these commented out lines

scripts/build-windows-binary
rosskirkpat: Done

scripts/build-windows-binary
rosskirkpat: Fair point - this script is meant to be a placeholder until we are at a point where we need Windows drone runners for more than one task. The rke2 binary build for Windows will be moved to a Windows build host in the near future. Do you mind if we keep the script as-is for now based on our future plans?

dweomer: Well, my concern around maintaining multiple build scripts is still relevant. Is it really expected to be that different? We are using go to build a binary executable after all.

dweomer (approved):
Approved with the caveat that having a separate build script for linux and windows is a recipe for failure in the indefinite future when we fix a build-time issue for linux but not windows (or vice versa). Therefore, we should plan to reconcile/unify them at some point in the near to mid term.

dweomer:
> Approved with the caveat that having a separate build script for linux and windows is a recipe for failure in the indefinite future when we fix a build-time issue for linux but not windows (or vice versa). Therefore, we should plan to reconcile/unify them at some point in the near to mid term.

@rosskirkpat a thought just occurred to me: have we tried cross-compiling? I think it could result in less bifurcation.

rosskirkpat:
> 
> 
> > Approved with the caveat that having a separate build script for linux and windows is a recipe for failure in the indefinite future when we fix a build-time issue for linux but not windows (or vice versa). Therefore, we should plan to reconcile/unify them at some point in the near to mid term.
> 
> @rosskirkpat a thought just occurred to me: have we tried cross-compiling? I think it could result in less bifurcation.

We are cross-compiling for Windows on Linux right [now](https://github.com/rancher/rke2/pull/1168/files/f9d7dca659b1e2ce31ea3b17c4ae3eef8ba6402c#diff-139d90a59aa01994da92fb793aaec0276586aad0a5dd180763fc573b1b30b468R44-R46)

.drone.yml
```diff
@@ -16,7 +16,9 @@ steps:
       from_secret: gcloud_auth
   commands:
   - docker pull --quiet rancher/hardened-build-base:v1.16.4b7
-  - dapper -f Dockerfile --target dapper make dapper-ci
+  - docker pull --quiet alpine:latest
```
brandond, line 19: why is this necessary? Won't this be pulled on-demand for Dockerfile.windows?

Dockerfile.windows
```diff
@@ -1,27 +1,63 @@
 FROM alpine AS build
```
brandond, line 1: This should be pinned, for example `alpine:3.12` as used by the golang image in the other Dockerfiles

brandond (commented):
I'm not a huge fan of moving scripts and dockerfiles into Linux variants - I think Linux should continue to be the (unlabeled) default, with Windows being the odd variant that gets special variants. The assumption is that Kubernetes is built on and for Linux unless you go out of your way to do otherwise. Willing to defer to the rest of the team on this though.

.drone.yml
dweomer: > why is this necessary? Won't this be pulled on-demand for Dockerfile.windows?

my working assumption here is that this is an attempt to more fully pre-cache images used in the build. the exact reason for the force pull of rancher/hardened-build-base escapes me but it is there for a reason

dweomer:
> > > Approved with the caveat that having a separate build script for linux and windows is a recipe for failure in the indefinite future when we fix a build-time issue for linux but not windows (or vice versa). Therefore, we should plan to reconcile/unify them at some point in the near to mid term.
> > 
> > 
> > @rosskirkpat a thought just occurred to me: have we tried cross-compiling? I think it could result in less bifurcation.
> 
> We are cross-compiling for Windows on Linux right [now](https://github.com/rancher/rke2/pull/1168/files/f9d7dca659b1e2ce31ea3b17c4ae3eef8ba6402c#diff-139d90a59aa01994da92fb793aaec0276586aad0a5dd180763fc573b1b30b468R44-R46)

ah, of course we are. apologies. /badbrain

ibuildthecloud:
I too would prefer we do not rename files to `.linux`.  Regarding dockerfiles it is possible to embed windows into the same Dockerfile as a different target in a multistage build. Just do `FROM windows AS foo` and then `docker build --target foo`.  Not saying you have to do that, but uber dockerfiles are possible.

Dockerfile.windows
rosskirkpat: Done, pinned `alpine:3.14`

.drone.yml
rosskirkpat: @dweomer that's right, I wasn't sure of the reason but I wanted to mimic the same process in case there was actually an issue caused by not pre-caching the images.

rosskirkpat:
> 
> 
> I too would prefer we do not rename files to `.linux`. Regarding dockerfiles it is possible to embed windows into the same Dockerfile as a different target in a multistage build. Just do `FROM windows AS foo` and then `docker build --target foo`. Not saying you have to do that, but uber dockerfiles are possible.

Thanks for the feedback, I switched back to using `Dockerfile` only. I'll check with the Windows and rke2 teams on if we want to keep the Dockerfiles separated or use a multistage build moving forward

Dockerfile.windows
```diff
@@ -1,4 +1,4 @@
-FROM alpine AS build
+FROM alpine:3.14 AS build
```
brandond, line 1: We're using 3.12 everywhere else, is there any particular reason you went with 3.14 for Windows?

brandond:
It looks like you removed references to Dockerfile.linux but didn't rename it back yet?

brandond:
This appears to be a new failure unrelated to anything you're changing. This should probably have a dedicated issue opened for it, and be fixed in a separate commit that you can rebase on top of.

```console
586 | /go/bin/golangci-lint
587 | pkg/controllers/cisnetworkpolicy/controller.go:25:6: func name will be used as cisnetworkpolicy.CISNetworkPolicyController by other packages, and that stutters; consider calling this Controller (golint)
588 | func CISNetworkPolicyController(ctx context.Context, sc *server.Context) error {
589 | ^
590 | make: *** [Makefile:64: validate] Error 1
```

rosskirkpat:
Closing this PR in favor of https://github.com/rancher/rke2/pull/1212. All feedback here has been incorporated and the new PR was created due to multiple changes in master

PR #1203, opened 2021-06-25
changes to reduce build time

paraglade:
changes `gzip` to using multi threaded gzip compression; adjustments to zstd compression aggressiveness.

### DIFF
--- a/scripts/package-images
+++ b/scripts/package-images
@@ -11,7 +11,8 @@ for FILE in build/images*.txt; do
     BASE=$(basename ${FILE} .txt)
     TARFILE=build/images/${PROG}-${BASE}.tar
     cp -f ${FILE} dist/artifacts/${PROG}-${BASE}.${PLATFORM}.txt
-    zstd -T0 -8 -f --exclude-compressed --no-progress ${TARFILE} -o dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.zst
+    # zstd --long=25 required for device compatibility: https://github.com/rancher/wharfie/blob/main/pkg/tarfile/tarfile.go#L28-L33
+    zstd -T0 -8 -f --long=25 --exclude-compressed --no-progress ${TARFILE} -o dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.zst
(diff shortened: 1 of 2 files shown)


scripts/package-images
```diff
@@ -11,8 +11,8 @@ for FILE in build/images*.txt; do
     BASE=$(basename ${FILE} .txt)
     TARFILE=build/images/${PROG}-${BASE}.tar
     cp -f ${FILE} dist/artifacts/${PROG}-${BASE}.${PLATFORM}.txt
-    zstd -T0 -16 -f --long=25 --no-progress ${TARFILE} -o dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.zst
-    gzip -v -c ${TARFILE} > dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.gz
+    zstd -T0 -8 -f --exclude-compressed --no-progress ${TARFILE} -o dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.zst
+    pigz -v -c ${TARFILE} > dist/artifacts/${PROG}-${BASE}.${PLATFORM}.tar.gz
```
brandond, line 15: I'm OK with the change to pigz, but the zstd settings (`-16 --long=25`) need to be retained in order balance compression factor against the amount of memory used by RKE2 when decompressing the resulting tarball. `--exclude-compressed` should be safe to add, but I would like to see a comparison of the archive sizes before and after.

brandond (changes requested):

scripts/package-images
paraglade: From the testing I did comparing various options to zstd I was seeing around 2% difference in compression from a `-16` to `-8` on the compression level which I would suspect would have very little impact on RKE2. However this little change has resulted in over 10 minute faster build times. I am thinking any improvement to build times would be welcome here especially something as simple as this.

brandond: What about the `--long=25` bit, what's the impact of that? It needs to be in sync with
https://github.com/rancher/wharfie/blob/main/pkg/tarfile/tarfile.go#L28-L33

paraglade: Thanks for bringing this to my attention. I have put this option back in and will review the build for this change.

Issue #1190, opened 2021-06-24
[epic] Centralize Kubernetes component building into rancher/image-build-kubernetes and drop kube-proxy helm chart


brandond:
This is a high-level tracking issue for work to refactor how we build and deploy Kubernetes components.

### Goals
* Move Kubernetes component building out of rancher/rke2:
    * #1078
    * #1104
* Switch kube-proxy over to a static pod managed by the supervisor/agent instead of Daemonset: 
    * #751

### Scope
The current plan is to backport these changes to **all active branches**.

### Tasks
1. [x] Bring https://github.com/rancher/image-build-kubernetes back to life and up-to-date with the current build scripts, mostly cribbed from image-build-kube-proxy; ensure that it also builds the kube-proxy binary.
    History: we stopped using it when we decided to build most of the Kubernetes stuff in RKE2, and kube-proxy in the image-build-kube-proxy repo.
2. [x] Pull all the Kubernetes build stuff out of the RKE2 dockerfile, replace it with FROM hardened-kubernetes from image-build-kubernetes
    * #1191
    * #1192
    * #1193
    * #1194
3. [x] Implement the missing bits of the kube-proxy static pod executor in RKE2 - have it use the same hardened-kubernetes image as the rest of the control plane, plumb through the options, restore the disabled CLI flag, etc.
    * #1195
    * #1196
    * #1197
    * #1198
4. [x] Disable the rke2-kube-proxy helm chart from rke2-charts; stop hardcoding `--disable-kube-proxy` and use the static pod executor to launch it.
    **TODO:*** Figure out how to handle this from an upgrade perspective - may want to keep `--disable-kube-proxy` set and not disable the rke2-kube-proxy helm chart if we detect a rke2-kube-proxy HelmChartConfig resource present in the cluster, similar to what we did for the K3s traefik upgrade?
    * #1199
    * #1200
    * #1201
    * #1202

brandond:
All the work for this has landed on master and release branches. Closing.

Issue #723, opened 2021-02-24
rke2 server -> changing cluster-cidr to i.e. "172.27.0.0/16" does NOT change podCidr: 10.42.0.0/16 for rke2-canal


Martin-Weiss:
We have deployed RKE2 v1.19.7+rke2r1 with these settings in config.yaml:

```
cluster-cidr: "172.27.0.0/16"
service-cidr: "172.28.0.0/16"
```

But had to realize that the setup for canal is NOT adjusted and still sets the "wrong" default podCidr: 10.42.0.0/16.

We can see this in the logs of 

`kubectl -n kube-system logs rke2-canal-8mrtw kube-flannel`:

```
I0224 15:04:51.738121       1 iptables.go:145] Some iptables rules are missing; deleting and recreating rules
I0224 15:04:51.738142       1 iptables.go:167] Deleting iptables rule: -s 10.42.0.0/16 -d 10.42.0.0/16 -j RETURN
I0224 15:04:51.738616       1 iptables.go:145] Some iptables rules are missing; deleting and recreating rules
I0224 15:04:51.738625       1 iptables.go:167] Deleting iptables rule: -s 10.42.0.0/16 -j ACCEPT
I0224 15:04:51.738771       1 iptables.go:167] Deleting iptables rule: -s 10.42.0.0/16 ! -d 224.0.0.0/4 -j MASQUERADE --random-fully
I0224 15:04:51.739207       1 iptables.go:167] Deleting iptables rule: -d 10.42.0.0/16 -j ACCEPT
I0224 15:04:51.739442       1 iptables.go:167] Deleting iptables rule: ! -s 10.42.0.0/16 -d 10.42.1.0/24 -j RETURN
I0224 15:04:51.740152       1 iptables.go:155] Adding iptables rule: -s 10.42.0.0/16 -j ACCEPT
I0224 15:04:51.740519       1 iptables.go:167] Deleting iptables rule: ! -s 10.42.0.0/16 -d 10.42.0.0/16 -j MASQUERADE --random-fully
I0224 15:04:51.741348       1 iptables.go:155] Adding iptables rule: -s 10.42.0.0/16 -d 10.42.0.0/16 -j RETURN
I0224 15:04:51.742006       1 iptables.go:155] Adding iptables rule: -d 10.42.0.0/16 -j ACCEPT
[...]

PR #1205, opened 2021-06-25
Adding configuration to run kube proxy as a static pod.

phillipsj:
This enables the flags, settings, images, and the static pod executor to allow configuring and running kube proxy as a static pod. 
Signed-off-by: Jamie Phillips <jamie.phillips@suse.com>

#### Proposed Changes ####

This enables running kube proxy as a static pod. It enables the CLI args and adds kube proxy to the list of images to be consumed.

This may require changes to existing documentation.

#### Types of Changes ####

This is a new feature and may be a breaking change without careful consideration to users of the existing RKE2 Chart.

#### Verification ####

Two items really need verified. The first is that a cluster can be created using the new static kube proxy. The other is that an existing cluster that is consuming the kube proxy chart can be upgraded to this new version without any issues.

#### Linked Issues ####

This is linked to the following epic: https://github.com/rancher/rke2/issues/1190
This specific issue: https://github.com/rancher/rke2/issues/1195
This work for Windows enablement: https://github.com/rancher/windows/issues/55

#### Further Comments ####

This is dependent on https://github.com/rancher/rke2/issues/1191


pkg/cli/defaults/defaults.go
```diff
@@ -25,7 +25,7 @@ func Set(clx *cli.Context, dataDir string) error {
 	cmds.ServerConfig.HTTPSPort = 6443
 	cmds.ServerConfig.APIServerPort = 6443
 	cmds.ServerConfig.APIServerBindAddress = "0.0.0.0"
-	cmds.ServerConfig.DisableKubeProxy = true
+	cmds.ServerConfig.DisableKubeProxy = false
```
brandond, line 28: just delete it
```suggestion
```

pkg/images/images.go
```diff
@@ -44,6 +45,7 @@ type Resolver struct {
 type ImageOverrideConfig struct {
 	SystemDefaultRegistry string
 	KubeAPIServer         string
+	KubeProxy             string
```
brandond, line 48: nit: this should go after KubeControllerManager (the Kube images are sorted alphabetically)

pkg/images/images.go
```diff
@@ -70,6 +72,7 @@ func NewResolver(c ImageOverrideConfig) (*Resolver, error) {
 	}{
 		{ETCD, c.ETCD},
 		{KubeAPIServer, c.KubeAPIServer},
+		{KubeProxy, c.KubeProxy},
```
brandond, line 75: nit: this should go after KubeControllerManager (the Kube images are sorted alphabetically)

brandond (changes requested):
* Need to add the kube-proxy image to a bunch of places:
https://github.com/rancher/rke2/blob/cb509d28471892f0135a998808654843a7a19905/pkg/images/images.go#L26
https://github.com/rancher/rke2/blob/cb509d28471892f0135a998808654843a7a19905/scripts/build-binary#L27-L30
https://github.com/rancher/rke2/blob/cb509d28471892f0135a998808654843a7a19905/pkg/images/images.go#L203
https://github.com/rancher/rke2/blob/cb509d28471892f0135a998808654843a7a19905/pkg/cli/cmds/root.go#L31-L36

Eventually it will just use the default kubernetes image, but until we get the build pipeline refactored it needs to be handled in its own case:
https://github.com/rancher/rke2/blob/cb509d28471892f0135a998808654843a7a19905/pkg/images/images.go#L212

pkg/cli/defaults/defaults.go
phillipsj: I am planning to wait until kube proxy is added to the kubernetes-images. If we wait on that, then I don't need to make the changes in the first two items, correct?

briandowns (approved):

pkg/cli/defaults/defaults.go
brandond: Yeah, if you get it in the kubernetes image then this should be fine

brandond (commented):
This looks good at the moment.

Still need a bit of code to handle upgrades - if there's a rke2-kube-proxy HelmChartConfig resource in the cluster we should set DisableKubeProxy=true; if there is not then we should add `rke2-kube-proxy` to the `disable` list.

phillipsj:
@brandond I will get that implemented this afternoon.

phillipsj:
@brandond @luthermonson I have added the startup hook that checks for the existence of the HelmChartConfig and if it exist, I disable kube proxy and add the chart to the disabled items.

pkg/rke2/kp.go
```diff
@@ -0,0 +1,43 @@
...
+				logrus.WithError(err).Info("Kube Proxy chart config wasn't found")
+			}
+			if helmChartConfig != nil {
+				cfg.DisableKubeProxy = true
+				if err := clx.Set("disable", kubeProxyChart); err != nil {
+					logrus.Fatalf("KubeProxyChartCheck: disabling the chart: %s", err.Error())
+				}
+
+			}
+			logrus.Info("Kube proxy chart check completed")
```
brandond, line 39: * Don't use server.NewContext, it sets up a lot more than we need. 
* Need to handle NotFound errors specifically, as that's not the only kind of issue we might get back.
* If the config IS found, we need to `--disable-kube-proxy`, if it IS NOT found we need to `--disable=rke2-kube-proxy`

Something like this perhaps:

```suggestion
			restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigAdmin)
			if err != nil {
				logrus.Fatalf("kp: new k8s client: %s", err.Error())
			}
			
			hc, err := helm.NewForConfig(restConfig)
			if err != nil {
				logrus.Fatalf("kp: new helm client: %s", err.Error())
			}

			if _, err := hc.HelmV1().HelmChartConfig().Get(metav1.NamespaceSystem, kubeProxyChart, metav1.GetOptions{}); err != nil {
				if apierrors.IsNotFound(err) {
					logrus.Infof("%[1]s HelmChartConfig not found, disabling %[1]s", kubeProxyChart)
					return clx.Set("disable", kubeProxyChart);
				} else {
					return errors.Wrap(err, )
				}
			}
			logrus.Infof("%s HelmChartConfig found, disabling embedded kube-proxy", kubeProxyChart)
			cfg.DisableKubeProxy = true
			return nil
			
```

brandond (changes requested):

pkg/rke2/kp.go
```diff
@@ -0,0 +1,52 @@
...
+// disables running kubeproxy as a static pod.
+func setKubeProxyDisabled(clx *cli.Context, cfg *cmds.Server) func(context.Context, <-chan struct{}, string) error {
+	return func(ctx context.Context, apiServerReady <-chan struct{}, kubeConfigAdmin string) error {
+		go func() {
+			<-apiServerReady
+			logrus.Info("Checking for kube proxy as a chart")
+
+			restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigAdmin)
+			if err != nil {
+				logrus.Fatalf("kp: new k8s cient: %s", err.Error())
```
brandond, line 31: this is probably my own typo
```suggestion
				logrus.Fatalf("kp: new k8s client: %s", err.Error())
```

pkg/rke2/kp.go
```diff
@@ -0,0 +1,52 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/k3s-io/helm-controller/pkg/generated/controllers/helm.cattle.io"
+
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
brandond, line 16: I feel like gofmt would tidy this a bit?

brandond (changes requested):

pkg/rke2/kp.go
```diff
@@ -0,0 +1,52 @@
...
+				if apierrors.IsNotFound(err) {
+					logrus.Infof("%[1]s HelmChartConfig not found, disabling %[1]s", kubeProxyChart)
+					clx.Set("disable", kubeProxyChart)
+					return
+				}
+				logrus.WithError(err).Fatalf("kp: failed to check for %s HelmChartConfig", kubeProxyChart)
+				return
+			}
+			logrus.Infof("%s HelmChartConfig found, disabling embedded kube-proxy", kubeProxyChart)
+			cfg.DisableKubeProxy = true
```
brandond, line 48: I am slightly concerned that kube-proxy might already be queued up to be started by the time we get here, but this should be easy enough to test.

We might also need to add some logic to remove the kube-proxy static chart if it was enabled and has been disabled on this startup. The static pod executor currently lacks this sort of logic, so there are bits of code scattered around to clean up manifests when etcd or the apiserver are disabled. This can probably be fixed later, and more holistically.

pkg/rke2/kp.go
phillipsj: Yeah, GoLand does that sometimes and I didn't notice.

pkg/rke2/kp.go
phillipsj: @brandond how do we want to proceed?

brandond: I would probably just verify that this is actually working - install RKE2 1.21 from tarball, add a HelmChartConfig for rke2-kube-proxy, then install a CI build and verify that it disables the kube-proxy static chart as desired.

phillipsj: @brandond I tested this and verified that the kube-proxy static pod doesn't start up.

brandond (approved):

galal-hussein (approved):
LGTM

mstrent:
How might this relate to #723? Would it resolve the issue, or do we need to continue to workaround with the new CLI arg?

PR #1218, opened 2021-06-29
Switch to using hardened k8s image

luthermonson:
Switches a few things in the build process, most importantly moving to the hardened-kubernetes image with a hardcoded v1.21.1 default tag which can be passed in to the build and corresponds to a matching tag in `rancher/image-build-kubernetes`. Other notable build changes...
* Moves the gzip/zstd of all images to the tag process to speed up PRs
* removes go mod vendor drone step as it's being done in validate now from a previous pr
* switch to pigz over gzip, credit to @paraglade for this one, can probably close #1203 in favor of this pr
* adds a dev-runtime-image which zstd and packages the runtime image for tests

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ platform:
   arch: amd64
 
 steps:
-- name: build-and-package
+- name: build
   image: rancher/dapper:v0.5.5
   environment:
     ENABLE_REGISTRY: 'true'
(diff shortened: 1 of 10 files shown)


brandond (approved):
LGTM.

One note for other reviewers: One notable side effect of this this PR is that the test image that we run our conformance in no longer gets a full airgap tarball - it instead only gets a limited archive containing the just-built runtime image. This will result in a more things getting pulled during tests, but since we do use a local pull-through mirror, the impact should be minimal.

Oats87 (approved):

briandowns (approved):

PR #1238, opened 2021-07-01
update golangci-lint settings to improve max mem usage and time

luthermonson:


.golangci.json
```diff
@@ -10,17 +10,37 @@
 			"gofmt"
 		]
 	},
+	"linters-settings": {
+		"govet": {
+			"check-shadowing": false
+		},
+		"gofmt": {
+			"simplify": false
```
brandond, line 18: I have found these useful in the past, how much time does dropping these save?

luthermonson: more than i care to admit. ill do a diff on the two but it's what weuse in rancher2, only reason i brought it forward. most everyone runs goimports locally any way i dont see why we need more than this but i could be talked out of it

luthermonson:
```
#with
level=info msg="Memory: 62 samples, avg is 325.6MB, max is 1366.2MB"
level=info msg="Execution took 6.0739442s"

#without
level=info msg="Memory: 62 samples, avg is 342.8MB, max is 1490.5MB"
level=info msg="Execution took 6.0010344s"
```

just changes max mem usage. execution time is basically the same. i dont care if we leave it but like i said, we disable it in rancher2 and dont seem to care over there

brandond: I'd vote for leaving it, its just a few milliseconds difference and somehow actually takes MORE memory to turn it off?

brandond (approved):

.golangci.json
```diff
@@ -10,17 +10,34 @@
...
+	"linters-settings": {
+		"govet": {
+			"check-shadowing": false
+		}
+	},
 	"run": {
 		"skip-files": [
 			"/zz_generated_"
 		],
+		"skip-dirs": [
```
briandowns, line 22: Does this have an option for "include" rather than "exclude"? Might end up being easier. No big deal either way.

luthermonson: weirdly... no

briandowns: Fair enough. :(

briandowns (approved):
LGTM

Issue #1252, opened 2021-07-07
INSTALL_RKE2_COMMIT installation method does not work


manuelbuil:
**Environmental Info:**
RKE2 Version: 

* master
* v1.21

**Describe the bug:**

When trying to install RKE2 pointing to a specific commit, it fails. e.g.

```
$> curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=06a4601e7f2cbf2589db4fc810519ed66dfaf0f3 sh -
[INFO]  using commit 06a4601e7f2cbf2589db4fc810519ed66dfaf0f3 as release
[INFO]  downloading airgap checksums at https://storage.googleapis.com/rke2-ci-builds/rke2-images.linux-amd64-06a4601e7f2cbf2589db4fc810519ed66dfaf0f3.tar.gz.sha256sum
curl: (22) The requested URL returned error: 404 
```
When looking at https://storage.googleapis.com/rke2-ci-builds/, I can see that the commit is there but the url is wrong. It should be https://storage.googleapis.com/rke2-ci-builds/rke2.linux-amd64-06a4601e7f2cbf2589db4fc810519ed66dfaf0f3.tar.gz.sha256sum

**Steps To Reproduce:**

```
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=06a4601e7f2cbf2589db4fc810519ed66dfaf0f3 sh -
```

manuelbuil:
Issue fixed in the script that uploads the tarball

brandond:
Not quite fixed yet, waiting on #1260

manuelbuil:
Removed the milestone while waiting for the PR to resolve this issue

manuelbuil:
PR has been merged. Testing the latest commit SHA works:
```
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=41d9fc0b3bf4673edd36de76735ab8d1956b93a9 sh -
[INFO]  using commit 41d9fc0b3bf4673edd36de76735ab8d1956b93a9 as release
[INFO]  downloading airgap checksums at https://storage.googleapis.com/rke2-ci-builds/rke2-images.linux-amd64-41d9fc0b3bf4673edd36de76735ab8d1956b93a9.tar.zst.sha256sum
[...]

PR #1253, opened 2021-07-07
Fix bug when installing by commit

manuelbuil:

#### Proposed Changes ####

Fixes the installation by INSTALL_RKE2_COMMIT which was pointing to a wrong url

#### Types of Changes ####

Bugfix

#### Verification ####

Install using INSTALL_RKE2_COMMIT, e.g.
`curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=06a4601e7f2cbf2589db4fc810519ed66dfaf0f3 sh -`

#### Linked Issues ####

https://github.com/rancher/rke2/issues/1252

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -344,9 +344,13 @@ download_airgap_checksums() {
         return
     fi
     AIRGAP_CHECKSUMS_URL=${STORAGE_URL}/rke2-images.${SUFFIX}-${INSTALL_RKE2_COMMIT}.tar.zst.sha256sum
-    # try for zst first; if that fails use gz for older release branches
+    # try zst first; if that fails try gz for other release branch
     if ! check_download "${AIRGAP_CHECKSUMS_URL}"; then
-        AIRGAP_CHECKSUMS_URL=${STORAGE_URL}/rke2-images.${SUFFIX}-${INSTALL_RKE2_COMMIT}.tar.gz.sha256sum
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -358,10 +358,10 @@ download_airgap_tarball() {
     if [ -z "${INSTALL_RKE2_COMMIT}" ]; then
         return
     fi
-    AIRGAP_TARBALL_URL=${STORAGE_URL}/rke2-images.${SUFFIX}-${INSTALL_RKE2_COMMIT}.tar.zst
+    AIRGAP_TARBALL_URL=${STORAGE_URL}/rke2.${SUFFIX}-${INSTALL_RKE2_COMMIT}.tar.zst
     # try for zst first; if that fails use gz for older release branches
```
brandond, line 362: Instead of changing this, another try for zst with the new name. With this change there will be three different variations that we might need to find depending on what release branch this commit was built from.

brandond (changes requested):

install.sh
manuelbuil: ok! For both tarball and sha256sum, right?

brandond: correct!

manuelbuil: The suffix I see in https://storage.googleapis.com is tar.gz though. Not tar.zst. I'll check rke2 without `-images` for that one

brandond:
This ended up being an issue with the upload script, fixed in #1258 - I don't think anything needs to be fixed on this side.

PR #1258, opened 2021-07-07
update crane to use .tar only, add pigz to windows image packaging, revert build-upload changes

rosskirkpat:


scripts/package-windows-images
```diff
@@ -8,22 +8,23 @@ source ./scripts/version.sh
 mkdir -p dist/artifacts
 
 # 1809
-crane --platform windows/amd64 pull ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64 rke2-windows-1809-images.tar.gz
-crane --platform windows/amd64 pull mcr.microsoft.com/oss/kubernetes/pause:3.5-windows-1809-amd64 rke2-windows-1809-images.tar.gz
+crane --platform windows/amd64 pull ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64 rke2-windows-1809-images.tar
+crane --platform windows/amd64 pull mcr.microsoft.com/oss/kubernetes/pause:3.5-windows-1809-amd64 rke2-windows-1809-images.tar
```
brandond, line 12: I just verified locally that the second command will overwrite the tar from the first command, so the resulting tarball only has the pause image in it. For this and the rest of the `crane pull` commands you need to put all of the images you want in the tarball in a single command, same as you would for `docker save`.

brandond (changes requested):

scripts/package-windows-images
brandond: I'm a little confused how we're building `${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64` as platform windows/amd64 since docker on Linux appears to refuse to interact with windows images entirely. Have you validated that this bit is working?

brandond: This needs a larger fix to how the `${PROG}-runtime:${DOCKERIZED_VERSION}` image list is built; going to punt on this for now so that we can get Linux working again.

brandond (approved):

luthermonson (approved):

luthermonson:
going to fix some other things on this after the fact, some of this isnt' 100% for windows

Issue #549, opened 2020-11-24
CIS 1.5 Control 5.1.5 Remediation


briandowns:
Currently, RKE2 can be secured manually by updating the default service account to conform to the CIS 1.5 control 5.1.5 however it'd be preferred to have RKE2 do this automatically if the `--profile=cis-1.5` argument is provided.

rancher-max:
Validated in a simple test on master branch with commit `525479888074426ea6019474c33e184a947c197c` using both cis-1.5 and cis-1.6 profiles that this control is now passing. As this has already been validated on previous released backport branches, I do not believe I need to do any more validations around this here.

However, during testing with an rc when available, I will run rancher scans on a hardened cluster to ensure they all pass in 1.22.

PR #1259, opened 2021-07-07
Added startup hook to restrict serviceAccounts

dereknola:

#### Proposed Changes ####
Automate CIS 5.1.5 so user intervention in not required to harden rke2. Changed documentation to reflect this automation.

#### Types of Changes ####
Enhancement

#### Verification ####
Startup rke2 with CIS profile:
`./rke2 server --profile=cis-1.5`
Once up, check service accounts
`kubectl get serviceaccount -A --field-selector metadata.name=default -o json`
All default accounts should have `"automountServiceAccountToken": false,`

#### Linked Issues ####
https://github.com/rancher/rke2/issues/549


pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func restrictServiceAccount(ctx context.Context, namespace string, cs *kubernetes.Clientset) error {
+	var sa *v1.ServiceAccount
```
briandowns, line 15: Generally we want to define this as a normal typed value and then take its address. You can remove the nil check on 29, simplifies line 32

pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,64 @@
...
+	var err error
```
briandowns, line 16: Can you scope lock the err value on 18?

pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,64 @@
...
+	return func(ctx context.Context, apiServerReady <-chan struct{}, kubeConfigAdmin string) error {
+		if cisMode {
+			logrus.Info("Restricting automount...")
+			go func() {
+				<-apiServerReady
+				cs, err := newClient(kubeConfigAdmin, nil)
+				if err != nil {
+					logrus.Fatalf("serviceAccount: new k8s client: %s", err.Error())
+				}
+				var namespaces = []string{
```
briandowns, line 48: We probably want to move this to package level since we're using a slice just like this in a number of other places in this package. In an individual function use case, we can append on additional namespaces as needed.

pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,64 @@
...
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func restrictServiceAccount(ctx context.Context, namespace string, cs *kubernetes.Clientset) error {
+	var sa *v1.ServiceAccount
+	var err error
+	for {
```
briandowns, line 17: There are examples of retry logic in this package in the ns.go and psp.go files that would probably be more appropriate or at least more consistent.

pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,64 @@
...
+		sa, err = cs.CoreV1().ServiceAccounts(namespace).Get(ctx, "default", metav1.GetOptions{})
```
briandowns, line 18:
```suggestion
		sa, err = cs.CoreV1().ServiceAccounts(namespace).Get(ctx, metav1.NamespaceDefault, metav1.GetOptions{})
```

briandowns:
Can you also include some tests for this? There are a few already in this package for reference.

pkg/rke2/serviceaccount.go
dereknola: This isn't a namespace, its the name of the service account resource. It just happens to be the "default" one.

pkg/rke2/serviceaccount.go
dereknola: Moved up a function.

pkg/rke2/serviceaccount.go
dereknola: Made a regular bool and assigned its pointer to the serviceaccount

pkg/rke2/serviceaccount.go
dereknola: I have scope locked the err and for loop.

pkg/rke2/serviceaccount.go
dereknola: Moved up a function

pkg/rke2/serviceaccount.go
brandond: Upstream uses:
```go
import (
  serviceaccountadmission "k8s.io/kubernetes/plugin/pkg/admission/serviceaccount"
)

sa, err = cs.CoreV1().ServiceAccounts(namespace).Get(ctx, serviceaccountadmission.DefaultServiceAccountName, metav1.GetOptions{})

File: pkg/rke2/serviceaccount.go [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+					metav1.NamespaceSystem,
+					metav1.NamespacePublic,
+					"kube-node-lease",
+				}
+				for _, namespace := range namespaces {
+					if err := restrictServiceAccount(ctx, namespace, cs); err != nil {
+						logrus.Fatalf("serviceAccount: namespace %s %s", namespace, err.Error())
+					}
+				}
+				logrus.Info("Restricting automount for serviceAccounts complete")
```
brandond, line 59:
```suggestion
				logrus.Info("Restricting automount for default serviceAccounts complete")
```

brandond (changes requested):
couple nits but looks good!

brandond (approved):

pkg/rke2/serviceaccount.go
dereknola: Fixed

pkg/rke2/serviceaccount.go
dereknola: Updated log message

pkg/rke2/serviceaccount_test.go
```diff
@@ -0,0 +1,131 @@
...
+	},
+}
+var testServiceAccountFilled = &v1.ServiceAccount{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "default",
+	},
+	AutomountServiceAccountToken: new(bool),
+}
+
+func addClientReactors(cs *fake.Clientset, verb string, pass bool) *fake.Clientset {
```
brandond, line 28: 💯

pkg/rke2/serviceaccount.go
```diff
@@ -0,0 +1,84 @@
...
+	*sa = *newSA
+	return nil
+}
+
+func restrictServiceAccount(ctx context.Context, namespace string, cs kubernetes.Interface) error {
+	var backoff = wait.Backoff{
+		Steps:    10,
+		Duration: 5 * time.Second,
+	}
+	// There are two race conditions this function avoids, a race on getting the initial sa because it does not yet exists,
```
briandowns, line 39:
```suggestion
	// There are two race conditions this function avoids, a race on getting the initial sa because it does not yet exist,
```

briandowns:
This looks great. What is the coverage look like for the added code?

dereknola:
> This looks great. What is the coverage look like for the added code?

We don't have code cov for rke2 (yet :), but the code coverage for the is as follows:
```
github.com/rancher/rke2/pkg/rke2/serviceaccount.go:17:	retryError				100.0%
github.com/rancher/rke2/pkg/rke2/serviceaccount.go:24:	updateServiceAccountRef			83.3%
github.com/rancher/rke2/pkg/rke2/serviceaccount.go:34:	restrictServiceAccount			92.9%
github.com/rancher/rke2/pkg/rke2/serviceaccount.go:63:	restrictServiceAccounts			0.0%
```
And `restrictServiceAccounts` just launched the thread for calling `restrictServiceAccount`

pkg/rke2/serviceaccount_test.go
```diff
@@ -0,0 +1,131 @@
...
+var testServiceAccountEmpty = &v1.ServiceAccount{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "default",
+	},
+}
+var testServiceAccountFilled = &v1.ServiceAccount{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "default",
+	},
+	AutomountServiceAccountToken: new(bool),
```
briandowns, line 25: the `new` keyword is soon to be removed or deprecated from Go and the Go authors suggest not using it.

brandond: vendor/k8s.io/utils/pointer `BoolPtr(bool)` perhaps?

briandowns: Dig it!

dereknola: Fixed!

briandowns (approved):
LGTM. Great job.

Oats87 (approved):

Issue #1279, opened 2021-07-12
Windows service Subcommand


luthermonson:
Create a service subcommand for windows to add/delete rke2 as a windows service using the service manager calls. This subcommand will be windows only and will not appear or work in linux.

e.g. 

```
rke2 agent service --add
rke2 agent service --delete
```

PR #1281, opened 2021-07-12
Adding agent service sub command for Windows

luthermonson:
adds the service subcommand to agent which can add/delete a windows service for rke2 via the windows service manager. the agent will only work with a config yaml so --config is a required param and get's passed/set on the service via args so ultimately the call to run the agent will default to `rke2 agent --config /etc/rancher/k3s/config.yaml`

https://github.com/rancher/rke2/issues/1279

### DIFF
--- a/pkg/cli/cmds/agent_service_windows.go
+++ b/pkg/cli/cmds/agent_service_windows.go
@@ -1,0 +1,132 @@
+// +build windows
+
+package cmds
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+	"unsafe"
+
+	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/rancher/k3s/pkg/version"
+	"github.com/urfave/cli"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+const (
+	defaultServiceDescription = "Rancher Kubernetes Engine v2 (agent) see docs https://github.com/rancher/rke2#readme"
+)
+
+var serviceSubcommand = cli.Command{
+	Name:   "service",
+	Usage:  "Manage RKE2 as a Windows Service",
+	Action: serviceAction,
+	Flags: []cli.Flag{
+		cmds.ConfigFlag,
+		&cli.BoolFlag{
+			Name:  "add",
+			Usage: "add RKE2 as a Windows Service",
+		},
+		&cli.BoolFlag{
+			Name:  "delete",
+			Usage: "stop and delete RKE2 as a Windows Service",
+		},
+		&cli.StringFlag{
+			Name:  "service-name",
+			Usage: "name for the RKE2 service",
+			Value: version.Program,
+		},
+	},
+}
+
+func serviceAction(ctx *cli.Context) error {
+	add := ctx.Bool("add")
+	del := ctx.Bool("delete")
+	if (!add && !del) || (add && del) {
+		return errors.New("service subcommand requires one of --add or --delete")
+	}
+
+	if del {
+		return deleteWindowsService(ctx.String("service-name"))
+	}
+
+	return addWindowService(ctx.String("service-name"), ctx.String("config"))
+}
+
+func getServicePath() (string, error) {
+	p, err := exec.LookPath(os.Args[0])
+	if err != nil {
+		return "", err
+	}
+	return filepath.Abs(p)
+}
+
+// inspiration https://github.com/containerd/containerd/blob/main/cmd/containerd/command/service_windows.go#L127
+func addWindowService(serviceName, config string) error {
+	p, err := getServicePath()
+	if err != nil {
+		return err
+	}
+	m, err := mgr.Connect()
+	if err != nil {
+		return err
+	}
+	defer m.Disconnect()
+
+	s, err := m.CreateService(serviceName, p, mgr.Config{
+		ServiceType:  windows.SERVICE_WIN32_OWN_PROCESS,
+		StartType:    mgr.StartAutomatic,
+		ErrorControl: mgr.ErrorNormal,
+		DisplayName:  version.Program,
+		Description:  defaultServiceDescription,
+	}, "agent", "--config", config)
+	if err != nil {
+		return err
+	}
+	defer s.Close()
+
+	const (
+		scActionRestart             = 1
+		serviceConfigFailureActions = 2
+	)
+
+	type scAction struct {
+		Type  uint32
+		Delay uint32
+	}
+
+	t := []scAction{
+		{Type: scActionRestart, Delay: uint32(30 * time.Second / time.Millisecond)}, // on failure restart every 30s
+	}
+
+	type serviceFailureActions struct {
+		ResetPeriod  uint32
+		RebootMsg    *uint16
+		Command      *uint16
+		ActionsCount uint32
+		Actions      uintptr
+	}
+
+	lpInfo := serviceFailureActions{ResetPeriod: uint32(30), ActionsCount: uint32(1), Actions: uintptr(unsafe.Pointer(&t[0]))}
+	return windows.ChangeServiceConfig2(s.Handle, serviceConfigFailureActions, (*byte)(unsafe.Pointer(&lpInfo)))
+}
+
+func deleteWindowsService(serviceName string) error {
+	m, err := mgr.Connect()
+	if err != nil {
+		return err
+	}
+	defer m.Disconnect()
+
+	s, err := m.OpenService(serviceName)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+	"unsafe"
+
+	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/urfave/cli"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+const (
+	defaultServiceDisplayName = "rke2"
```
briandowns, line 20: I think this is already defined in the version package in k3s and set at build time.

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+func serviceSubcommand() cli.Command {
+	return cli.Command{
+		Name:   "service",
+		Usage:  "Manage rke2 as a Windows Service",
+		Action: serviceAction,
+		Flags: []cli.Flag{
+			cmds.ConfigFlag,
+			&cli.BoolFlag{
+				Name:  "add",
+				Usage: "add rke2 as a Windows Service",
```
phillipsj, line 33: RKE should be capitalized.

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+			},
+			&cli.BoolFlag{
+				Name:  "delete",
+				Usage: "stop and delete rke2 as a Windows Service",
+			},
+			&cli.StringFlag{
+				Name:  "service-name",
+				Usage: "name for the rke2 service",
```
phillipsj, line 41: RKE should be capitalized.

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+		Action: serviceAction,
+		Flags: []cli.Flag{
+			cmds.ConfigFlag,
+			&cli.BoolFlag{
+				Name:  "add",
+				Usage: "add rke2 as a Windows Service",
+			},
+			&cli.BoolFlag{
+				Name:  "delete",
+				Usage: "stop and delete rke2 as a Windows Service",
```
phillipsj, line 37: RKE should be capitalized.

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+
+const (
+	defaultServiceDisplayName = "rke2"
+	defaultServiceDescription = "Rancher Kubernetes Engine v2 (agent) see docs https://github.com/rancher/rke2#readme"
+)
+
+func serviceSubcommand() cli.Command {
+	return cli.Command{
+		Name:   "service",
+		Usage:  "Manage rke2 as a Windows Service",
```
phillipsj, line 27: RKE should be capitalized.

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+const (
+	defaultServiceDisplayName = "rke2"
+	defaultServiceDescription = "Rancher Kubernetes Engine v2 (agent) see docs https://github.com/rancher/rke2#readme"
+)
+
+func serviceSubcommand() cli.Command {
```
briandowns, line 24: Does this need to be returned from a function? Can it be an unexported package var?

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,138 @@
...
+
+func serviceAction(ctx *cli.Context) error {
+	add := ctx.Bool("add")
+	del := ctx.Bool("delete")
+	if !add && !del {
+		return errors.New("service subcommand requires --add or --delete")
+	}
+
+	if _, err := os.Stat(ctx.String("config")); os.IsNotExist(err) {
+		return errors.New("config must be a valid rke2 yaml config file")
```
phillipsj, line 56: RKE should be capitalized.

briandowns:
Should this functionality be added to k3s and carried into RKE2? That's what we did for the etcd subcommands. Also, take a look at how the etcd subcommands are and try to mirror that approach for consistency's sake.

luthermonson:
> Should this functionality be added to k3s and carried into RKE2? That's what we did for the etcd subcommands. Also, take a look at how the etcd subcommands are and try to mirror that approach for consistency's sake.

ya we already had this convo, no one committing to windows in k3s so no point in the commands there. i agree it SHOULD be there but that requires us to get to a point where k3s works in windows too

pkg/cli/cmds/agent_service_windows.go
luthermonson: yes it could, not a bad idea

briandowns:
> > Should this functionality be added to k3s and carried into RKE2? That's what we did for the etcd subcommands. Also, take a look at how the etcd subcommands are and try to mirror that approach for consistency's sake.
> 
> ya we already had this convo, no one committing to windows in k3s so no point in the commands there. i agree it SHOULD be there but that requires us to get to a point where k3s works in windows too

Fair enough. I had forgotten about that convo.

pkg/cli/cmds/agent_service_windows.go
luthermonson: it is indeed, version.Program, i've moved to this now

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func serviceAction(ctx *cli.Context) error {
+	add := ctx.Bool("add")
+	del := ctx.Bool("delete")
+	if !add && !del {
+		return errors.New("service subcommand requires --add or --delete")
+	}
+
+	if _, err := os.Stat(ctx.String("config")); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
luthermonson: it will be set as the default, im just confirming it actually exists. it's not required to pass, the default value is already there from urfave populating it from flag Value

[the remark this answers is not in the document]
brandond: OK - get the intent, but with currently this file doesn't actually have to exist. For example, due to directory loading support, the user may choose to put files at c:\etc\rancher\rke2\config.yaml.d\01-my-config.yaml. I would probably just skip this check entirely.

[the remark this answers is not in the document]
luthermonson: should we not allow/care about passing in a config at all? should we just call `rke2 agent` in the service and let everything else sort out? or you're saying if it's set then set it else leave blank?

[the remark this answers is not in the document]
brandond: For this subcommand, I would:
* Leave it blank by default
* If the value is empty, don't pass through a `--config` flag to the service
* If the value is non-empty, pass it through to rke2 as `--config=X` without validating it.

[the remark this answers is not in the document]
brandond: or since you're just reusing the flag including the default, just pass it through without `stat`ing I guess

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,136 @@
...
+			Usage: "name for the RKE2 service",
+			Value: version.Program,
+		},
+	},
+}
+
+func serviceAction(ctx *cli.Context) error {
+	add := ctx.Bool("add")
+	del := ctx.Bool("delete")
+	if !add && !del {
```
brandond, line 49: Similar check for `if add && del {`?

phillipsj (approved):

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	defer m.Disconnect()
+
+	s, err := m.CreateService(serviceName, p, mgr.Config{
+		ServiceType:  windows.SERVICE_WIN32_OWN_PROCESS,
+		StartType:    mgr.StartAutomatic,
+		ErrorControl: mgr.ErrorNormal,
+		DisplayName:  version.Program,
+		Description:  defaultServiceDescription,
+	}, "agent", "--config "+config)
```
brandond, line 86: should be `"--config=config"` or `"--config", "config"` but not `--config config`
```suggestion
	}, "agent", "--config="+config)
```

brandond (changes requested):

brandond (approved):

briandowns (approved):

PR #1311, opened 2021-07-14
fix crane cli for windows, add manifest-tool

rosskirkpat:
fix how crane builds the windows airgap images
add manifest-tool for runtime images


scripts/package-windows-images
```diff
@@ -5,26 +5,31 @@ cd $(dirname $0)/..
...
 
-#mkdir -p dist/artifacts
+mkdir -p dist/artifacts
 
 # 1809
-#crane --platform windows/amd64 pull ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64 rke2-windows-1809-images.tar
-#crane --platform windows/amd64 pull mcr.microsoft.com/oss/kubernetes/pause:3.5-windows-1809-amd64 rke2-windows-1809-images.tar
+crane --platform windows/amd64 pull \
+  ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64 \
+  mcr.microsoft.com/oss/kubernetes/pause:3.5-windows-1809-amd64 \
```
brandond, line 13: as per @luthermonson we are going to use the gcr.k8s.io/pause:3.5 images as they have all the right Windows versions in them already. Long term these need to be mirrored into docker.io/rancher/mirrored-pause.

brandond (changes requested):

manifest-runtime.tmpl
```diff
@@ -0,0 +1,12 @@
+image: {{repo.name}}/rke2-runtime:{{replace "+" "-" build.tag}}
+manifests:
+  - 
+    image: {{repo.name}}/rke2-runtime:{{replace "+" "-" build.tag}}-linux-amd64
```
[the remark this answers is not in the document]
rosskirkpat: I was going off of how it is implemented in [rancher/rancher](https://github.com/rancher/rancher/blob/master/manifest-runtime.tmpl). I can't seem to find any documentation on what the proper way to do this is yet

scripts/package-windows-images
rosskirkpat: Looks like we already have 3.5 pause images for the three versions of Windows that we are supporting (1809, 2004, 20H2) in https://hub.docker.com/r/rancher/mirrored-pause/

Is there a reason why we can't point to those pause images now?

manifest-runtime.tmpl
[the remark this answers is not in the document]
brandond: It's just weird to have a newline immediately after starting the list, before the first key in the dict

scripts/package-windows-images
brandond: I don't personally care, we just need to use the same one in the airgap tarballs as the build. I'll let the two of you work it out.

rosskirkpat: Switching back to `gcr.k8s.io/pause:3.5` after talking with Luther

scripts/package-windows-images
```diff
@@ -5,26 +5,30 @@ cd $(dirname $0)/..
...
-#for TARFILE in "${WINDOWS_TARFILES[@]}"; do
-#    cp -f ${TARFILE} dist/artifacts/${PROG}-${BASE}.windows-amd64.txt
-#    zstd -T0 -16 -f --long=25 --no-progress ${TARFILE} -o dist/artifacts/${PROG}-${BASE}.windows-amd64.tar.zst
-#    pigz -v -c ${TARFILE} > dist/artifacts/${PROG}-${BASE}.windows-amd64.tar.gz
-#    cp -f ${TARFILE} dist/artifacts/
-#
-#done
-#cat build/windows-images.txt | sort -V | uniq > dist/artifacts/${PROG}-images.windows-amd64.txt
+crane --platform windows/amd64 pull \
+  ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64 \
```
brandond, line 24: The tags that you're pulling here don't match the tags that you're specifying for DefaultPauseImage or DefaultKubernetesImage. They need to be exactly the same or airgap will not work.

manifest-runtime.tmpl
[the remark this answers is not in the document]
brandond: I'd still like this to be formatted normally, in line with other yaml documents in this repo.

brandond (changes requested):

scripts/package-windows-images
rosskirkpat: This will be documented as a limitation for airgap of the 2.6.0 Tech Preview for Windows custom cluster provisioning. We will provide documentation on how the customers should re-tag the pause image. Windows custom cluster provisioning will GA in 2.6.x 
We have opened an issue to track this #1336

.drone.yml
```diff
@@ -120,28 +120,6 @@ platform:
...
-  when:
-    event:
-    - tag
-    instance:
-    - drone-publish.rancher.io
-    ref:
-    - refs/head/master
-    - refs/tags/*
-
 - name: manifest-kubernetes
```
galal-hussein, line 123: we should also remove the manifest-kubernetes since its already being made in https://github.com/rancher/image-build-kubernetes/blob/master/.drone.yml#L61-L76

briandowns (approved):

.drone.yml
rosskirkpat: Done

galal-hussein (approved):

PR #1345, opened 2021-07-16
[Release 1.21] Adding 9153 in networkpolicy for core DNS metrics

sebastienmusso:
Adding core dns metrics tcp port in network policy

### DIFF
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -81,6 +81,12 @@ var networkDNSPolicy = v1.NetworkPolicy{
 							IntVal: int32(53),
 						},
 					},
+					{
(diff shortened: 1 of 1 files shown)


brandond:
Can you open an issue describing the reason you'd like this change?

sebastienmusso:
> Can you open an issue describing the reason you'd like this change?

Hi,

Issue have bien créateur

https://github.com/rancher/rke2/issues/1383

QuinnBast:
Why was this closed? Without this, Prometheus cannot scrape the coredns pods for metrics.

Is there a workaround to this that would allow Prometheus to scrape coredns?

Issue #1301, opened 2021-07-14
kube-proxy deployed twice undetected by CI


manuelbuil:
**Environmental Info:**
RKE2 Version: master

manuel@master:~$ /usr/local/bin/rke2 -v
rke2 version v1.21.1-dev+87acb68b (87acb68b3e2ab4788c1915628535766a5c40a653)
go version go1.16.4b7

Node(s) CPU architecture, OS, and Version: 

Linux master 5.4.0-74-generic #83-Ubuntu SMP Sat May 8 02:35:39 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

1 master

**Describe the bug:**

3 bugs:
- [x] kube-proxy is deployed twice, one as a static pod and one as a daemonset via helmChart install

This is the result:
```
kube-system   kube-proxy-master                            1/1     Running     0          34m   10.17.3.1   master   <none>           <none>
kube-system   kube-proxy-p2vrt                             1/1     Running     0          34m   10.17.3.1   master   <none>           <none>
```
- [x] the static kube-proxy is not deployed with privileged: true and thus, iptables does not have the permissions to run
Logs:
```
E0714 14:54:03.159397       1 proxier.go:867] "Failed to ensure chain exists" err="error creating chain \"KUBE-EXTERNAL-SERVICES\": exit status 3: iptables v1.8.5 (legacy): can't initialize iptables table `filter': Permission denied (you must be root)\nPerhaps iptables or your kernel needs to be upgraded.\n" table=filter chain=KUBE-EXTERNAL-SERVICES
```

- [x] even though the static kube-proxy is failing because of the iptables permissions and the "daemonset" kube-proxy fails because of colliding with the static kube-proxy:
```
E0714 14:27:41.275235       1 server.go:627] starting metrics server failed: listen tcp 127.0.0.1:10249: bind: address already in use
```
both pods are running correctly when listed, which made the bug undetectable by CI

**Steps To Reproduce:**

- Installed RKE2: 
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=87acb68b3e2ab4788c1915628535766a5c40a653 sh -

brandond:
The fix for this will need to be backported to all branches that the static kube-proxy changes were backported to, as a release blocker.

rancher-max:
[...]

PR #1381, opened 2021-07-20
Switching to using StartupHookArgs and adding Disables/Skips

luthermonson:
#### Linked Issues ####
https://github.com/rancher/rke2/issues/1301

### DIFF
--- a/pkg/rke2/kp.go
+++ b/pkg/rke2/kp.go
@@ -2,7 +2,6 @@ package rke2
 
 import (
 	"context"
-	"sync"
 
 	"github.com/k3s-io/helm-controller/pkg/generated/controllers/helm.cattle.io"
 	"github.com/rancher/k3s/pkg/cli/cmds"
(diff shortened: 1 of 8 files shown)


pkg/rke2/kp.go
```diff
@@ -36,7 +35,8 @@ func setKubeProxyDisabled(clx *cli.Context, cfg *cmds.Server) func(context.Conte
 			if _, err := hc.Helm().V1().HelmChartConfig().Get(metav1.NamespaceSystem, kubeProxyChart, metav1.GetOptions{}); err != nil {
 				if apierrors.IsNotFound(err) {
 					logrus.Infof("%[1]s HelmChartConfig not found, disabling %[1]s", kubeProxyChart)
-					clx.Set("disable", kubeProxyChart)
+					args.Skips[kubeProxyChart] = false
```
brandond, line 38: We **do** want to skip and disable these.

brandond (changes requested):

pkg/rke2/kp.go
luthermonson: ah yes

brandond (approved):

brandond (approved):

manuelbuil (approved):

PR #1437, opened 2021-07-22
moving all steps into one pipeline for dist dir to be available

luthermonson:


rosskirkpat (approved):
LGTM. this condenses multiple pipelines into one

PR #1461, opened 2021-07-26
Add Windows Installation for Rancher, Update existing Installer

luthermonson:
#### Proposed Changes ####

We are adding support for installing the Windows Agent via Rancher with the rke2-installer.ps1. We have also resolved bugs related to the current install.ps1 in the root of the repo.

This will improve the experience for all users and allow better integration with Rancher provisioning.

#### Types of Changes ####
A new feature which is the RKE2-Installer.ps1 and bug fixes for installer.ps1.

#### Verification ####

Install via the installer.ps1 produces a Windows node that joins a cluster. The rancher installer will need have a build at the head of the repository, then create a custom cluster with Windows.

#### Linked Issues ####
https://github.com/rancher/rancher/issues/33625

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -95,6 +95,24 @@ function Write-FatalLog() {
     exit 1
 }
 
+function Confirm-WindowsFeatures {
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory = $true)]
+        [String[]]
+        $RequiredFeatures
+    )
+    foreach ($feature in $RequiredFeatures) {
+        $f = Get-WindowsFeature -Name $feature
+        if (-not $f.Installed) {
+            Write-FatalLog "Windows feature: '$feature' is not installed. Please run: Install-WindowsFeature -Name $feature"
+        }
+        else {
+            Write-InfoLog "Windows feature: '$feature' is installed. Installation will proceed."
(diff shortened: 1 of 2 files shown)


install.ps1
```diff
@@ -509,7 +542,9 @@ function Install-AirgapTarball() {
 $STORAGE_URL = "https://storage.googleapis.com/rke2-ci-builds"
 $INSTALL_RKE2_GITHUB_URL = "https://github.com/rancher/rke2"
 $DEFAULT_TAR_PREFIX = "C:\usr\local"
+$INSTALL_RKE2_TAR_PREFIX = "C:\usr\local"
```
dereknola, line 545: Does this need to exist, as $DEFAULT_TAR_PREFIX is the same value above?

briandowns (approved):
LGTM overall but I'd like someone who knows powershell to give it a look.

install.ps1
luthermonson: its cause they both exist in the install.sh... dont disagree with you but im not sure it's my call

dereknola: Okay if they already are there, matching install.sh is fine.

dereknola (approved):

Issue #1046, opened 2021-05-24
Allow configuring resource requests and limits for control-plane static pods


brandond:
**Is your feature request related to a problem? Please describe.**
Allow configuring resource requests and limits for control-plane static pods

**Describe the solution you'd like**

Allow setting requests and limits via two new flags: `control-plane-resource-requests` and `control-plane-resource-limits`. 

These would be structured much like the [kubelet system-reserved flag](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#system-reserved). The bare `cpu`, `memory`, and `ephemeral-storage` keys are honored as a default for all components. Component-specific overrides can also be specified, in the form of `<NAME>-<RESOURCE>` for example `etcd-cpu`, `kube-apiserver-memory`, `kube-controller-manager-ephemeral-storage` et cetera.

We **should** include the `kube-` prefix on these, as this is the convention for the CLI pasthrough flags (`kube-apiserver-arg`, et al).

Example:  
* `--control-plane-resource-requests=[cpu=100m][,][memory=100Mi][,][ephemeral-storage=1Gi][,][etc-cpu=100m][,][etcd-memory=100Mi][,][etcd-ephemeral-storage=1Gi]`
* `--control-plane-resource-limits=[cpu=100m][,][memory=100Mi][,][ephemeral-storage=1Gi][,][kube-apiserver-cpu=100m][,][kube-apiserver-memory=100Mi][,][kube-apiserver-ephemeral-storage=1Gi]`

**Describe alternatives you've considered**
* Do nothing, continue to only use the embedded default CPU requests
* Allow the user to provide a 'template' static pod manifest for each component that we merge our desired configuration into, and document which fields in the manifest we overwrite or merge.    
    * This was decided against as being documentation-heavy, in addition to the fact that we currently blindly overwrite and/or delete the static pod manifests during cluster startup or reconfiguration.

**Additional context**
cc @nikkelma @Oats87

Oats87:
Consolidating this issue to https://github.com/rancher/rke2/issues/1233

Issue #1233, opened 2021-06-30
Allow customizing of the static pod manifests for CPU, files, and directories


Oats87:
In order to keep feature parity with RKE1, we should support setting:

* Custom Environment Variables
* Custom bind-mounted directories from the host

More information on this capability can be found here: https://rancher.com/docs/rke/latest/en/config-options/services/services-extras/

We should also allow customization of the CPU request, and, optionally, allow the user to set CPU limits on the static pod run components even though this is not ideal from a usability perspective, there could still be merit/reason to do so.

brandond:
We have existing backlog issues for these under:
* #930
* #1046

Oats87:
Thanks @brandond . Those are quite a bit more "nuanced" and specific. Wondering if we should independently track the functionality or just merge into one "super" issue, considering the code change to implement all of this is going to be in the same part of code.

Oats87:
Consolidated https://github.com/rancher/rke2/issues/1046 to this issue

rancher-max:
Confirmed these extra args are working on master branch commit `525479888074426ea6019474c33e184a947c197c` using the same steps as mentioned in https://github.com/rancher/rke2/issues/1525

PR #1463, opened 2021-07-27
Custom resource requests/limits, extra binds, and extra env for control plane static pod components

Oats87:
#### Proposed Changes ####
Allow for custom resource requests/limits, extra binds, and extra environment variables for control plane static pod components

#### Types of Changes ####
CLI, static pod executor

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1046


pkg/cli/cmds/root.go
```diff
@@ -93,10 +93,96 @@ var (
...
+			Usage:  "(components) " + rke2.CloudControllerManager + " extra environment variables",
+			EnvVar: "RKE2_" + strings.ToUpper(strings.ReplaceAll(rke2.CloudControllerManager, "-", "_")) + "_EXTRA_ENV",
+			Value:  &config.ExtraEnv.CloudControllerManager,
+		},
 	}
 )
 
-const pkdFlagName = "protect-kernel-defaults"
+const (
+	pkdFlagName = "protect-kernel-defaults"
```
briandowns, line 184:
```suggestion
	protectKernelDefaultsFlag = "protect-kernel-defaults"
```

pkg/podexecutor/staticpod.go
```diff
@@ -41,17 +41,65 @@ var (
 	defaultAuditPolicyFile = "/etc/rancher/rke2/audit-policy.yaml"
 )
 
+type ControlPlaneResources struct {
```
briandowns, line 44: We can probably drop all of the prefixed "kube" since that's implied and will make visually scanning this struct a lot easier.

pkg/podexecutor/staticpod.go
```diff
@@ -41,17 +41,65 @@ var (
...
-	DataDir         string
-	AuditPolicyFile string
-	KubeletPath     string
-	DisableETCD     bool
-	IsServer        bool
+	ManifestsDir          string
+	ImagesDir             string
+	Resolver              *images.Resolver
+	CloudProvider         *CloudProviderConfig
+	CISMode               bool
```
briandowns, line 94: Can you group the bools together and put them at the bottom?

briandowns:
Can you provide a few examples of what the data passed in would look like to the newly added arguments?

pkg/podexecutor/staticpod.go
brandond: I am +1 on keeping them as-is for parity with the Image stuff at https://github.com/rancher/rke2/blob/6c1df6c9f38c89f4c9d629a2f775804f589795b7/pkg/images/images.go#L47-L52

pkg/rke2/rke2_linux.go
```diff
@@ -56,16 +58,187 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
...
+
+	extraEnv := podexecutor.ControlPlaneEnv{
+		KubeAPIServer:          cfg.ExtraEnv.KubeAPIServer.Value(),
+		KubeScheduler:          cfg.ExtraEnv.KubeScheduler.Value(),
+		KubeControllerManager:  cfg.ExtraEnv.KubeControllerManager.Value(),
+		Etcd:                   cfg.ExtraEnv.Etcd.Value(),
+		CloudControllerManager: cfg.ExtraEnv.CloudControllerManager.Value(),
+	}
+
+	extraMounts := podexecutor.ControlPlaneMounts{
```
brandond, line 221: Should we do input validation on these? Later on we do `mount := strings.Split(rawMount, ":")` and reference `mount[1]` but we don't actually validate that they contain at least one semicolon, so that could cause a runtime panic.

brandond (changes requested):

pkg/rke2/rke2_linux.go
Oats87: Yeah, I've added some basic input validation to check that the length of the resultant slice is 2, but this doesn't actually validate that the paths are valid or anything. I'm not sure how far we want to go with the validations, as the kubelet will block the static pods from running itself as well if the volume mounts/source don't make sense

pkg/rke2/rke2_linux.go
```diff
@@ -56,16 +58,188 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 		cfg.KubeletPath = "kubelet"
 	}
 
+	var controlPlaneResources podexecutor.ControlPlaneResources
+
+	if cfg.ControlPlaneResourceRequests != "" {
```
brandond, line 63: There's a ton of copy-pasted boilerplate code for extracting the requests and limits. I feel like we could do something with a multidimensional map of component-resource-limit and iterate across those, instead of having it all unrolled here.

What do you think about something like this:
https://play.golang.org/p/A79up20daNq

brandond (changes requested):

pkg/rke2/rke2_linux.go
Oats87: This is super efficient compared to what I built. I'll integrate and test this

Oats87: Integrated your suggestion and optimized it a little bit to be more universal.

brandond (approved):

pkg/staticpod/staticpod.go
```diff
@@ -231,6 +269,50 @@ func addVolumes(p *v1.Pod, src []string, dir bool) {
...
+	}
+}
+
+func addExtraEnv(p *v1.Pod, extraEnv []string) {
+	for _, rawEnv := range extraEnv {
+		env := strings.Split(rawEnv, "=")
+		if len(env) != 2 {
+			logrus.Errorf("environment variable for pod %s %s was not valid", p.Name, rawEnv)
+			continue
+		}
```
brandond, line 308: I think this would preclude setting environment variables with an `=` in them; we should probably just split it by the first = sign and pass it in as-is.

```suggestion
		env := strings.SplitN(rawEnv, "=", 2)
```

brandond (changes requested):

pkg/staticpod/staticpod.go
```diff
@@ -231,6 +269,50 @@ func addVolumes(p *v1.Pod, src []string, dir bool) {
...
+			Name:      name,
+			ReadOnly:  false,
+			MountPath: mount[1],
+		})
+	}
+}
+
+func addExtraEnv(p *v1.Pod, extraEnv []string) {
+	for _, rawEnv := range extraEnv {
+		env := strings.Split(rawEnv, "=")
```
brandond, line 304: I think this would preclude setting environment variables with an = in them as you'll end up with too many parts.
```suggestion
		env := strings.Split(rawEnv, "=", 2)
```

pkg/staticpod/staticpod.go
```diff
@@ -231,6 +269,61 @@ func addVolumes(p *v1.Pod, src []string, dir bool) {
...
+		if len(mount) == 3 {
+			switch strings.ToLower(mount[2]) {
+			case "ro":
+				ro = true
+			case "rw":
+				ro = false
+			default:
+				logrus.Errorf("unknown mount option: %s encountered in extra mount %s for pod %s", mount[2], rawMount, p.Name)
+			}
+		}
```
brandond, line 294: I think this would be a lot tidier as a `switch len(mount)`. case 2 is a noop, case 3 checks for ro/rw, default is an error.

brandond (changes requested):

pkg/staticpod/staticpod.go
Oats87: I changed to using a switch, but am now doing a switch within a switch. I think this is acceptable.

pkg/staticpod/staticpod.go
```diff
@@ -231,6 +268,64 @@ func addVolumes(p *v1.Pod, src []string, dir bool) {
...
+	var (
+		prefix     = "extra-mount"
+		sourceType = v1.HostPathDirectoryOrCreate
+	)
+
+	for i, rawMount := range extraMounts {
+		mount := strings.Split(rawMount, ":")
+		var ro bool
+		switch len(mount) {
+		case 2:
```
briandowns, line 281: If we intend to keep this, can you add a comment as to why it's empty? Otherwise, remove it.

pkg/staticpod/staticpod.go
```diff
@@ -231,6 +268,64 @@ func addVolumes(p *v1.Pod, src []string, dir bool) {
 	}
 }
 
+func addExtraMounts(p *v1.Pod, extraMounts []string) {
+	var (
+		prefix     = "extra-mount"
```
briandowns, line 273: It looks like this can be a `const`.

pkg/staticpod/staticpod.go
```diff
@@ -143,11 +149,40 @@ func pod(args Args) (*v1.Pod, error) {
...
-			Requests: v1.ResourceList{
-				v1.ResourceCPU: *resource.NewMilliQuantity(args.CPUMillis, resource.DecimalSI),
-			},
+	p.Spec.Containers[0].Resources = v1.ResourceRequirements{}
+
+	p.Spec.Containers[0].Resources.Requests = v1.ResourceList{}
+	p.Spec.Containers[0].Resources.Limits = v1.ResourceList{}
+
+	if args.CPURequest != "" {
+		if cpuRequest, err := resource.ParseQuantity(args.CPURequest); err != nil {
```
briandowns, line 158: This isn't a blocker to merging but the `if` `else` can be simplified to a single `if`.

pkg/staticpod/staticpod.go
```diff
@@ -143,11 +149,40 @@ func pod(args Args) (*v1.Pod, error) {
 		},
 	}
 
-	if args.CPUMillis > 0 {
-		p.Spec.Containers[0].Resources = v1.ResourceRequirements{
-			Requests: v1.ResourceList{
-				v1.ResourceCPU: *resource.NewMilliQuantity(args.CPUMillis, resource.DecimalSI),
-			},
+	p.Spec.Containers[0].Resources = v1.ResourceRequirements{}
```
briandowns, line 152: Are these not already initialized to their 0 value for that type?

briandowns (approved):

brandond (approved):

pkg/staticpod/staticpod.go
brandond: I put a switch in your switch, so you can switch while you switch /xzibit

Oats87:
https://github.com/rancher/rke2/issues/1233

Issue #1467, opened 2021-07-27
Windows Service fails due to timeout issue.


phillipsj:
**Environmental Info:**
RKE2 Version: 

rke2.exe version v1.21.3+rke2r1 (2ed0b0d1b6924af4414393cd1796c174a1ff5352)
go version go1.16.6b7

Node(s) CPU architecture, OS, and Version: 

```
WindowsBuildLabEx                                       : 19041.1.amd64fre.vb_release.191206-1406
WindowsCurrentVersion                                   : 6.3
WindowsEditionId                                        : ServerStandardACor
WindowsInstallationType                                 : Server Core
WindowsInstallDateFromRegistry                          : 7/26/2021 2:10:32 PM
WindowsProductId                                        : 00431-30000-00000-AA117
WindowsProductName                                      : Windows Server Standard
```

Cluster Configuration:
1 server
1 windows node

**Describe the bug:**
Running the agent as a windows service errors with a timeout issue, why executing the agent not as a service works.

```
rke2.exe agent service --add
```

**Steps To Reproduce:**
Install RKE2 for Windows and try to run the agent as a windows service.

**Expected behavior:**
rke2 starts up as a windows service and joins a cluster.

**Actual behavior:**
rke2 starts then fails.

**Additional context / logs:**

rancher-max:
I'm still seeing this fail on v1.21.3-rc1+rke2r2, but it's no longer a timeout. Details below:

**Windows Version:**
```
WindowsBuildLabEx                                       : 17763.1.amd64fre.rs5_release.180914-1434
WindowsCurrentVersion                                   : 6.3
WindowsEditionId                                        : ServerDatacenter
WindowsInstallationType                                 : Server
WindowsInstallDateFromRegistry                          : 7/24/2021 6:08:30 AM
WindowsProductId                                        : 00430-00000-00000-AA230
WindowsProductName                                      : Windows Server 2019 Datacenter
```

**Cluster Info:**
1 server (ubuntu) running calico cni with strictaffinity

**Windows Node Steps:**
```
$ProgressPreference = 'SilentlyContinue'
New-Item -Type Directory c:\usr\local\bin -Force
Invoke-WebRequest -UseBasicParsing https://raw.githubusercontent.com/rancher/rke2/master/install.ps1 -OutFile c:\usr\local\bin\install-rke2.ps1
c:\usr\local\bin\install-rke2.ps1 -Version v1.21.3-rc1+rke2r2
[...]

PR #1475, opened 2021-07-28
Update service watching logic for windows service manager

luthermonson:
#### Proposed Changes ####
Adding windows service manager and eventlog hooks for running as a Windows Service.

#### Types of Changes ####
Bugfix

#### Verification ####
Start-Service rke2 should boot, check eventlog for errors

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1467

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -3,6 +3,7 @@ package cmds
 import (
 	"github.com/rancher/k3s/pkg/cli/cmds"
 	"github.com/rancher/rke2/pkg/rke2"
+	"github.com/rancher/rke2/pkg/windows"
 	"github.com/urfave/cli"
 )
 
@@ -80,5 +81,8 @@ func agentSubcommands() cli.Commands {
 func AgentRun(clx *cli.Context) error {
 	validateCloudProviderName(clx)
 	validateProfile(clx, "agent")
+	if err := windows.StartService(); err != nil {
+		return err
+	}
 	return rke2.Agent(clx, config)
 }
--- a/pkg/cli/cmds/agent_service_windows.go
+++ b/pkg/cli/cmds/agent_service_windows.go
@@ -13,7 +13,7 @@ import (
 	"github.com/rancher/k3s/pkg/cli/cmds"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/urfave/cli"
-	"golang.org/x/sys/windows"
+	syswin "golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc/mgr"
 )
 
@@ -78,7 +78,7 @@ func addWindowService(serviceName, config string) error {
 	defer m.Disconnect()
 
 	s, err := m.CreateService(serviceName, p, mgr.Config{
-		ServiceType:  windows.SERVICE_WIN32_OWN_PROCESS,
+		ServiceType:  syswin.SERVICE_WIN32_OWN_PROCESS,
 		StartType:    mgr.StartAutomatic,
 		ErrorControl: mgr.ErrorNormal,
 		DisplayName:  version.Program,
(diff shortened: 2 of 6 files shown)


pkg/cli/cmds/agent_service_windows.go
```diff
@@ -43,6 +47,36 @@ var serviceSubcommand = cli.Command{
...
+	}()
+
+	return nil
+}
+
+func watchService(stop chan struct{}) {
+	<-stop // pause for service to be stopped
+	ok, err := svc.IsWindowsService()
+	if err != nil {
+		logrus.Warnf("Error trying to determine if running as a Windows Service: %s", err)
```
briandowns, line 71: Should we stop processing here?

pkg/cli/cmds/agent_service_windows.go
```diff
@@ -10,10 +10,14 @@ import (
...
 	"unsafe"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/rancher/k3s/pkg/cli/cmds"
 	"github.com/rancher/k3s/pkg/version"
+	"github.com/rancher/rke2/pkg/windows"
 	"github.com/urfave/cli"
-	"golang.org/x/sys/windows"
+	syswin "golang.org/x/sys/windows"
```
briandowns, line 19: Why are we import aliasing?

brandond: because we already have `"github.com/rancher/rke2/pkg/windows"`

briandowns: Ugh, I just noticed. Right.

pkg/cli/cmds/agent_service_windows.go
luthermonson: debated that... ok will also be false and it will fall through, felt you had both things any way might as well just leave it. i can add a return here if you think it'll be more clear

brandond: Seems like the sort of thing that should either succeed or fail; if we can't figure it out I say we just exit.

luthermonson: ya but if it's not actually running in a service and this errored we'd kill the run? seems unlikely, thinking out loud

briandowns (approved):

pkg/cli/cmds/agent.go
```diff
@@ -80,5 +80,10 @@ func agentSubcommands() cli.Commands {
 func AgentRun(clx *cli.Context) error {
 	validateCloudProviderName(clx)
 	validateProfile(clx, "agent")
+	stop := make(chan struct{})
+	if err := startService(stop); err != nil {
+		return err
+	}
+	go watchService(stop)
```
brandond, line 87: Can you move all these lines into a function that's a noop on Linux, rather than unconditionally creating a channel that Linux doesn't even use?

pkg/cli/cmds/agent_service_windows.go
brandond: Well we're stopping here anyway so yeah, kill the run?

brandond (changes requested):

pkg/cli/cmds/agent_service_windows.go
luthermonson: not an exit tho? if its not a service but on the cli and this errors we can ignore and move along

pkg/cli/cmds/agent.go
luthermonson: kept them out here so it would make more sense that the cmd level is where a stop channel is created and acknowledged instead of deep in a library called "windows". i can tho, it's not hard to move

brandond: but Linux doesn't need the stop channel - it only exists to be passed into a goroutine that is a noop on Linux. I say move all of it into the windows package.

pkg/cli/cmds/agent_service_windows.go
brandond: How about only creating the stop channel and calling `go watchService(stop)` if `svc.IsWindowsService()` in `startService` returns `ok && err == nil`?

pkg/cli/cmds/agent.go
luthermonson: sure, can do

pkg/cli/cmds/agent_service_windows.go
luthermonson: sure lets

luthermonson:
@brandond @briandowns rewrote, added event logger

briandowns (approved):

brandond (approved):

Issue #1473, opened 2021-07-28
Harden autoscaler and clusterDNS images


manuelbuil:
**Is your feature request related to a problem? Please describe.**

rke2 should use harden all the images it uses

**Describe the solution you'd like**

Have the new images related to coredns hardened

manuelbuil:
We need https://github.com/rancherlabs/drone-publishing-setup/pull/102, so that drone can push the image into dockerhub

manuelbuil:
Important for testing! It takes ~2m for the `node-local-dns` pod to create the iptables. Wait for that to check the logs, otherwise it could be concluded that things work but they may not (talking from experience ;) )

galal-hussein:
Verified Against commit id 09bb5c27f29cf3dd831653543d274c948a225385

Steps to verify:

- Install rke2 server with this commit id 09bb5c27f29cf3dd831653543d274c948a225385
- add helm chart config for coredns to enable nodelocal cache:
```
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rke2-coredns
  namespace: kube-system
spec:
  valuesContent: |-
    nodelocal:
      enabled: true
```
- start rke2 server
- check images for coredns, autoscaler, and node-local-dns pods:
```
kubectl describe pod rke2-coredns-rke2-coredns-7bb4f446c-k8g8l -n kube-system | grep Image
    Image:         rancher/hardened-coredns:v1.8.3-build20210720

kubectl describe pod node-local-dns-9479c -n kube-system | grep Image
    Image:         rancher/hardened-dns-node-cache:1.20.0-build20210803

kubectl describe pod rke2-coredns-rke2-coredns-autoscaler-7c58bd5b6c-j95ml -n kube-system | grep -i image
    Image:         rancher/hardened-cluster-autoscaler:v1.8.3-build20210729
```
These versions are mentioned here https://github.com/rancher/rke2/pull/1513/files#diff-5860f3d4051a1dbb1a2a15c593338edfd3682123f4b21cda1699a1105a3f236dR16-R18

- Validated that DNS is working fine

PR #1513, opened 2021-08-03
Start using hardened dns images

manuelbuil:

#### Proposed Changes ####

This PR replaces the nodeCache and autoscaler default images with the hardened ones. To do so:

* Replaces the helm chart with one that uses those images
* Replaces the images in the airgap scripts
* Removes busybox image because we use the nodeCache image to check if the dns service is up (netcat was installed there)

#### Types of Changes ####

Image replacement

#### Verification ####

Deploy enabling autoscaler (by default) and nodeCache and verify that the images are the new ones

#### Linked Issues ####

https://github.com/rancher/rke2/issues/1473


manuelbuil:
node_local dns Cache has a bug in the image build

brandond (approved):

briandowns (approved):

Issue #1577, opened 2021-08-09
Windows Installation command isn't installing the correct version of RKE2, it installs the latest stable channel.


phillipsj:
**Describe the bug:**

The windows RKE2 installer for Rancher isn't receiving the version of RKE2 to install. By default it will install the latest stable channel release if no version is provided.

**Steps To Reproduce:**

Create a new RKE2 custom cluster with a 1.21.4 RC version and add Windows node.

**Expected behavior:**

Running rke2.exe --version should report 1.21.3-rc4+rke2r2

**Actual behavior:**

The version report 1.21.3-rke2r1

**Additional context / logs:**

* https://github.com/rancher/rancher/issues/34094

brandond:
The RKE2 PR for this issue has been merged and is ready to test; if there is another issue on the rancher side that is still in peer review that issue should be moved into Peer Review instead of this one.

sirredbeard:
Backing this out in order to perform some additional ETE testing prior to QA handoff.

rancher-max:
This was validated as part of v1.21. Closing as already done

PR #1602, opened 2021-08-10
Allows getting the version from Rancher and passes it to the installer.

phillipsj:
The Rancher installer now gets the version from Rancher and passes it to the regular installer.

Signed-off-by: Jamie Phillips <jamie.phillips@suse.com>

#### Types of Changes ####

Bugfix

#### Verification ####

That an install from Rancher installs the correct version on Windows.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/1577

### DIFF
--- a/windows/rke2-install.ps1
+++ b/windows/rke2-install.ps1
@@ -178,6 +178,7 @@ function Rke2-Installer
 
     function Set-Path
     {
+        $env:PATH += ";C:\var\lib\rancher\rke2\bin;C:\usr\local\bin"
         $environment = [System.Environment]::GetEnvironmentVariable("Path", "Machine")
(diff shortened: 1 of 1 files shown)


windows/rke2-install.ps1
```diff
@@ -365,6 +366,33 @@ function Rke2-Installer
...
+    {
+        $Uri = "$($env:CATTLE_SERVER)/v3/connect/cluster-info"
+        $path = "C:\etc\rancher\rke2"
+        $file = "info.json"
+        if (-Not(Test-Path $path)) {
+            New-Item -Path $path -ItemType Directory
+        }
+
+        $infoFile = Join-Path -Path $path -ChildPath $file
+        if (-Not $env:CATTLE_CA_CHECKSUM)
```
luthermonson, line 379: pretty sure you need to drop the -Not

brandond (approved):

Issue #1470, opened 2021-07-27
Windows agent doesn't always cleanup child processes like containerd.


phillipsj:
**Environmental Info:**
RKE2 Version: 

rke2.exe version v1.21.3+rke2r1 (2ed0b0d1b6924af4414393cd1796c174a1ff5352)
go version go1.16.6b7

Node(s) CPU architecture, OS, and Version: 

```
WindowsBuildLabEx                                       : 19041.1.amd64fre.vb_release.191206-1406
WindowsCurrentVersion                                   : 6.3
WindowsEditionId                                        : ServerStandardACor
WindowsInstallationType                                 : Server Core
WindowsInstallDateFromRegistry                          : 7/26/2021 2:10:32 PM
WindowsProductId                                        : 00431-30000-00000-AA117
WindowsProductName                                      : Windows Server Standard
```

Cluster Configuration:
1 server
1 windows node

**Describe the bug:**
Occasionally when the agent is stopped, the containerd process continues to run.

**Steps To Reproduce:**
Install RKE2 and start the windows agent. Stop the windows agent and check to see if the containerd process is still running.

**Expected behavior:**
All processes are stopped with agent.

**Actual behavior:**
Inconsistently containerd doesn't stop when the agent is stopped.

**Additional context / logs:**

rgl:
Maybe this is related to https://github.com/rancher/rke2/issues/1755?

brandond:
This may be resolved in 1.22:

https://github.com/rancher/rke2/pull/1827/commits/ad2bf416ee85aec8b3669156fd1a258809bb385c#diff-cc5276517ae75e00318f83d1a83000cb1836748a5a2dd32559b2fc9dc3483f10

PR #1619, opened 2021-08-12
Refactored how we start calico with kubelet context to cancel

luthermonson:
#### Proposed Changes ####

We needed to rework how we were handling the startup for Calico node and felix. We also discovered a few additional issues related to some tweaks that needed to be corrected.

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy a windows node with no error in the log files.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/1638
* https://github.com/rancher/rke2/issues/1470

### DIFF
--- a/pkg/pebinaryexecutor/pebinary.go
+++ b/pkg/pebinaryexecutor/pebinary.go
@@ -145,15 +145,24 @@ func (p *PEBinaryConfig) Kubelet(args []string) error {
 	logrus.Infof("Running RKE2 kubelet %v", cleanArgs)
 	go func() {
 		for {
+			ctx, cancel := context.WithCancel(context.Background())
+			go func() {
+				if err := p.cni.Start(ctx, p.cniConig); err != nil {
+					logrus.Errorf("error in cni start: %s", err)
+				}
+			}()
+
 			cmd := exec.Command(p.KubeletPath, cleanArgs...)
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
-			err := cmd.Run()
-			logrus.Errorf("Kubelet exited: %v", err)
+			if err := cmd.Run(); err != nil {
+				logrus.Errorf("Kubelet exited: %v", err)
+			}
+			cancel()
 			time.Sleep(5 * time.Second)
 		}
 	}()
-	return p.cni.Start(p.cniConig)
+	return nil
 }
 
 // KubeProxy starts the kubeproxy in a subprocess with watching goroutine.
@@ -173,13 +182,14 @@ func (p *PEBinaryConfig) KubeProxy(args []string) error {
 		extraArgs["enable-dsr"] = "true"
 	}
 
+	var vip string
 	for range time.Tick(time.Second * 5) {
 		endpoint, err := hcsshim.GetHNSEndpointByName("Calico_ep")
 		if err != nil {
 			logrus.WithError(err).Warningf("can't find %s, retrying", "Calico_ep")
 			continue
 		}
-		extraArgs["source-vip"] = endpoint.IPAddress.String()
+		vip = endpoint.IPAddress.String()
 		break
 	}
 
@@ -190,6 +200,13 @@ func (p *PEBinaryConfig) KubeProxy(args []string) error {
 	}
 
 	args = append(getArgs(extraArgs), args...)
+
+	for i, arg := range args {
+		if strings.Contains(arg, "source-vip") {
+			args[i] = "--source-vip=" + vip
+		}
+	}
+
 	logrus.Infof("Running RKE2 kube-proxy %s", args)
 	go func() {
 		for {
--- a/pkg/windows/calico.go
+++ b/pkg/windows/calico.go
@@ -106,24 +106,25 @@ func (c *Calico) Setup(ctx context.Context, dataDir string, nodeConfig *daemonco
 	if err := createCNIConfig(cfg); err != nil {
 		return nil, err
 	}
+
+	logrus.Info("Generating HNS networks, please wait")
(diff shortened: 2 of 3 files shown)


pkg/pebinaryexecutor/pebinary.go
```diff
@@ -190,6 +200,19 @@ func (p *PEBinaryConfig) KubeProxy(args []string) error {
...
+		if strings.Contains(arg, "source-vip") {
+			si = i
+		}
+	}
+
+	si, err := argIndex(args, "source-vip")
+	if err != nil {
+		return err
+	}
+	args[si] = fmt.Sprintf("--source-vip=%s", vip)
```
briandowns, line 214: Nit:

```suggestion
	args[si] = "--source-vip="+vip
```

This is more direct and removes a significant amount of unnecessary processing from "under the hood".

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -254,6 +277,15 @@ func getArgs(argsMap map[string]string) []string {
 	return args
 }
 
+func argIndex(args []string, s string) (int, error) {
+	for i, arg := range args {
+		if strings.Contains(arg, s) {
+			return i, nil
+		}
+	}
+	return -1, errors.New("index not found")
```
briandowns, line 286: Can we provide a little more information in this message?

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -254,6 +277,15 @@ func getArgs(argsMap map[string]string) []string {
 	return args
 }
 
+func argIndex(args []string, s string) (int, error) {
+	for i, arg := range args {
+		if strings.Contains(arg, s) {
+			return i, nil
+		}
+	}
+	return -1, errors.New("no matching argument found")
```
briandowns, line 286: What significance does `-1` have here? Can you create a constant with a descriptive name for this value?

phillipsj: What should I return as a default if the value isn't found? What is norm for that?

briandowns: I would return `0`. Anything else implies meaning and that it'd be checked. In this case, I'd assume the `err` will be checked and in that case, the calling code won't be checking the int value.

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -190,6 +200,19 @@ func (p *PEBinaryConfig) KubeProxy(args []string) error {
 	}
 
 	args = append(getArgs(extraArgs), args...)
+	var si int
+	for i, arg := range args {
+		if strings.Contains(arg, "source-vip") {
+			si = i
+		}
+	}
```
brandond, line 208: This appears to do the same thing as `argIndex`, should this have been removed?

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -254,6 +277,15 @@ func getArgs(argsMap map[string]string) []string {
 	return args
 }
 
+func argIndex(args []string, s string) (int, error) {
+	for i, arg := range args {
+		if strings.Contains(arg, s) {
+			return i, nil
+		}
+	}
```
brandond, line 285: If you're going to make a function for this, it should be safe to use for any arg - not just ones that you know aren't substrings of other args or possible arg values.
```suggestion
	s = "--" + s
	for i, arg := range args {
		parts := strings.SplitN(arg, "=", 2)
		if strings.HasPrefix(parts[0], s) {
			return i, nil
		}
	}
```

brandond (changes requested):

pkg/pebinaryexecutor/pebinary.go
phillipsj: Nice catch, fixed and retested.

pkg/pebinaryexecutor/pebinary.go
phillipsj: This adds functionality beyond what is required at the moment. I am happy to add an issue to track it, I just feel that it adds too much to the scope to address the current need and may cause potential issues if extra logic is added. This is only be used in the context of the Windows agent and the binary executor.

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -254,6 +273,15 @@ func getArgs(argsMap map[string]string) []string {
...
+	return indexNotFound, errors.New("no matching argument found")
```
brandond, line 282: I don't think this needs a constant. The convention is to return `nil` or the type default when err is non-nil. Just do `return 0, errors.New(...)` and drop the constant.

pkg/pebinaryexecutor/pebinary.go
brandond: I saw you had an earlier version of this that didn't use a function. If you have a function, make it work right and be reusable. If it's not reusable, or is implemented such that it only works for your specific use case, don't make it a function.

brandond (changes requested):

pkg/pebinaryexecutor/pebinary.go
phillipsj: @briandowns suggest using a constant. What should I do?

briandowns: The const was for the `-1` since it would generally imply some meaning. If it's just 0, that's fine.

phillipsj: @brandond Method removed.

pkg/pebinaryexecutor/pebinary.go
phillipsj: @brandond function removed.

briandowns (approved):

brandond (approved):

Issue #1525, opened 2021-08-03
[release-1.21] Allow customizing of the static pod manifests for CPU, files, and directories


Oats87:
Fix for **Allow customizing of the static pod manifests for CPU, files, and directories** on branch **release-1.21**
* https://github.com/rancher/rke2/issues/1233

rancher-max:
#### Validated the flags are present on v1.21.3-rc5+rke2r2

Made the following confirmations with set flags, using config.yaml shown below:
```yaml
write-kubeconfig-mode: 644
control-plane-resource-requests: kube-apiserver-cpu=250m,kube-apiserver-memory=256M,kube-scheduler-cpu=125m,kube-scheduler-memory=256M,etcd-cpu=500m
control-plane-resource-limits: kube-apiserver-cpu=500m,kube-apiserver-memory=512M,kube-scheduler-cpu=250m,kube-scheduler-memory=512M,etcd-cpu=1000m
kube-apiserver-extra-mount: /root/test/ka:/max/is/testing
kube-scheduler-extra-mount: /root/test/ks:/max/is/testing
kube-controller-manager-extra-mount: /root/test/kc:/max/is/testing
kube-proxy-extra-mount: /root/test/kp:/max/is/testing
etcd-extra-mount: /root/test/e:/max/is/testing:ro
cloud-controller-manager-extra-mount: 
  - /root/test/ccm:/max/is/testing
  - /root/test/ccmtwo:/max/is/boop:ro
kube-apiserver-extra-env: 
  - FOO=BAR
  - ANOTHER=OTHER
kube-scheduler-extra-env: BAR=BAZ
kube-controller-manager-extra-env: BAZ=FOO
kube-proxy-extra-env: HI=BYE
etcd-extra-env: GIMME=ALLTHETHINGS
cloud-controller-manager-extra-env: BOOP=BEEP
```

```
$ k describe -n kube-system pod/cloud-controller-manager-maxt2
...
    Environment:
      FILE_HASH:  330f44e7d71e959463fcbb6a7ca875b282e1f5d2c614f9c6b4b835642947a590
      NO_PROXY:   .svc,.cluster.local,10.42.0.0/16,10.43.0.0/16
      BOOP:       BEEP
    Mounts:
      /etc/ca-certificates from dir1 (rw)
      /etc/ssl/certs from dir0 (rw)
      /max/is/boop from extra-mount-1 (ro)
      /max/is/testing from extra-mount-0 (rw)
...

[...]

PR #1620, opened 2021-08-12
Make docs for new flags more clear

rancher-max:
#### Types of Changes ####

docs

#### Verification ####

N/A

#### Linked Issues ####

https://github.com/rancher/rke2/issues/1525

#### Further Comments ####

More clear since config.yaml is our recommended approach of installing and specifying the flag multiple times there only uses the last value specified.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -179,7 +179,7 @@ cloud-controller-manager
 Thus, an example `--control-plane-resource-requests` or `--control-plane-resource-limits` value may look like:
 
 ```
-kube-apiserver-cpu=500m,kube-apiserver-memory=512MiB,kube-scheduler-cpu=250m,kube-scheduler-memory=512MiB,etcd-cpu=1000m
+kube-apiserver-cpu=500m,kube-apiserver-memory=512M,kube-scheduler-cpu=250m,kube-scheduler-memory=512M,etcd-cpu=1000m
 ```
 
 The unit values for CPU/memory are identical to Kubernetes resource units (See: [Resource Limits in Kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes))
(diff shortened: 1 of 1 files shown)


brandond (approved):

rancher-max:
Had to push an update to the recommended limits, otherwise we hit this error:
```
Aug 12 18:24:07 maxt2 rke2[43738]: time="2021-08-12T18:24:07Z" level=error msg="error parsing memory request for static pod kube-apiserver: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'"
Aug 12 18:24:07 maxt2 rke2[43738]: time="2021-08-12T18:24:07Z" level=error msg="error parsing memory limit for static pod kube-apiserver: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'"
```

docs/advanced.md
```diff
@@ -179,7 +179,7 @@ cloud-controller-manager
 Thus, an example `--control-plane-resource-requests` or `--control-plane-resource-limits` value may look like:
 
 ```
-kube-apiserver-cpu=500m,kube-apiserver-memory=512MiB,kube-scheduler-cpu=250m,kube-scheduler-memory=512MiB,etcd-cpu=1000m
+kube-apiserver-cpu=500m,kube-apiserver-memory=512M,kube-scheduler-cpu=250m,kube-scheduler-memory=512M,etcd-cpu=1000m
```
brandond, line 182: Either one is fine: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

brandond:
That's interesting - either should be fine: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

rancher-max:
Yeah it was really the 'B' that was breaking it. Setting it with Mi works fine as well as it can be parsed by the given regex. The upstream docs _can_ be a little confusing due to statements like this: 
> ... Each Container has a limit of 0.5 cpu and 128MiB of memory ...

But overall it's pretty clear there what to use, just making this update for users who don't want to go look upstream.

brandond:
You're right, that's probably worth opening an issue in the upstream docs. They shouldn't use units in the text that the thing they're documenting doesn't support.

brandond (approved):

vadorovsky (approved):

Issue #1618, opened 2021-08-12
[release-1.21] Cluster fail to provision when using AWS cloud provider 


galal-hussein:
**Environmental Info:**
RKE2 Version: v1.21.3+rke2r1

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04

Cluster Configuration:
1 Server

**Describe the bug:**
When not passing `--node-name` to rke2 server and using `--cloud-provider-name=aws` , kubelet fail to register itself with kubernetes api because of the NodeRestriction Admission plugin https://github.com/rancher/rancher/issues/34105#issuecomment-897258413
  
**Steps To Reproduce:**
- Create ec2 instance on AWS
- Tag instance, subnet, vpc, and security group with kubernetes tag
- Start rke2 server with option --cloud-provider-name=aws

**Expected behavior:**
The cluster should start normally

**Actual behavior:**
Kubelet fails to register with the following error:

```
Attempting to register node ip-172-28-68-60.eu-west-1.compute.internal
Unable to register node "ip-172-28-68-60.eu-west-1.compute.internal" with API server: nodes "ip-172-28-68-60.eu-west-1.compute.internal" is forbidden: node "ip-172-28-68-60" cannot modify node "ip-172-28-68-60.eu-west-1.compute.internal"
```

cjellick:
I think we want the 1.21.3+rke2r2 milestone

brandond:
@cjellick I thought we were not actually planning on doing a 1.21.3+rke2r2 release - the current RCs are just for QA. The next actual release will be 1.21.4+rke2r1.

galal-hussein:
> @cjellick I thought we were not actually planning on doing a 1.21.3+rke2r2 release - the current RCs are just for QA. The next actual release will be 1.21.4+rke2r1.

Thats what I thought too

brandond:
This has been merged to master; I have updated the milestone to 1.22 and created a backport issue for 1.21.

brandond:
Edit: was merged to release-1.21 first, will convert the other issue into a forwardport

rancher-max:
#### Validated using v1.21.3-rc6+rke2r2

No longer need to supply node-name as hostname when using aws cloud provider. Can still supply it if desired. It will automatically use an equivalent to `hostname -f` when starting rke2 with aws cloud provider now though.

PR #1622, opened 2021-08-12
Use FQDN for node name if cloud provider is set to AWS

galal-hussein:
#### Proposed Changes ####

Use FQDN hostname for node name if `--node-name` is not set and `--cloud-provider-name` is set to AWS
 

Yes we should mention that if the user didn't enter a specific node name and cloud provider is set to aws then node name will default to `hostname -f`

#### Types of Changes ####
Bug Fix

#### Verification ####

- Start k3s server with cloud-provider-name=aws
 
#### Linked Issues ####
https://github.com/rancher/rke2/issues/1618

### DIFF
--- a/pkg/rke2/rke2_linux.go
+++ b/pkg/rke2/rke2_linux.go
@@ -66,9 +66,8 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 			Name: cfg.CloudProviderName,
 			Path: cfg.CloudProviderConfig,
 		}
-		nodeName := clx.String("node-name")
-		if nodeName == "" && cfg.CloudProviderName == "aws" {
-			fqdn, err := getFQDNHostname()
+		if clx.String("node-name") == "" && cfg.CloudProviderName == "aws" {
+			fqdn, err := hostnameFQDN()
 			if err != nil {
 				return nil, err
 			}
@@ -191,7 +190,7 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 	}, nil
 }
 
-func getFQDNHostname() (string, error) {
+func hostnameFQDN() (string, error) {
 	cmd := exec.Command("hostname", "-f")
 
 	var b bytes.Buffer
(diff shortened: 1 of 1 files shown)


briandowns (approved):

pkg/rke2/rke2_linux.go
```diff
@@ -178,3 +190,19 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
...
+	cmd.Stdout = &b
+
+	if err := cmd.Run(); err != nil {
+		return "", err
+	}
+
+	fqdn := b.String()
+	fqdn = fqdn[:len(fqdn)-1]
+
+	return fqdn, nil
```
brandond, line 207:
```suggestion
	return strings.TrimSpace(b.String()), nil
```

brandond (changes requested):
minor nit on string trimming, but lgtm!

pkg/rke2/rke2_linux.go
briandowns: If we do this, then we don't need line 205.

pkg/rke2/rke2_linux.go
```diff
@@ -64,6 +66,16 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 			Name: cfg.CloudProviderName,
 			Path: cfg.CloudProviderConfig,
 		}
+		nodeName := clx.String("node-name")
+		if nodeName == "" && cfg.CloudProviderName == "aws" {
```
brandond, line 70: We only appear to use the nodeName in the test; don't bother creating a variable for it?
```suggestion
		if clx.String("node-name") == "" && cfg.CloudProviderName == "aws" {
```

brandond (changes requested):

pkg/rke2/rke2_linux.go
galal-hussein: fixed

pkg/rke2/rke2_linux.go
galal-hussein: fixed

brandond (approved):
LGTM. I think at some point we might want this function in a util package somewhere, but that doesn't have to happen now.

PR #1723, opened 2021-08-29
Bump vsphere chart versions

cjellick:
#### Proposed Changes ####

Bump vsphere chart versions

The new versions of these charts have a control plane selector that can
handle both the "controlplane" and "control-plane" label. Previously the
chart was only looking for "controlplane" and rke2 only has the
"control-plane" label.

We need to do the raw chart versions because these charts are not
officially published yet in charts.rancher.io. This problem has
highlighted a bigger process problem we have around these charts, which
we'll need to fix later. Pulling in directly from the commit has is a
bit of a band aid.

Signed-off-by: Craig Jellick <craig@rancher.com>

#### Types of Changes ####
Chart version change

#### Verification ####
Deploy a multi-node rke2 cluster with vsphere turned on and verify vsphere's cloud controller manager properly deploys

#### Linked Issues ####
https://github.com/rancher/rancher/issues/33875

#### Further Comments ####
QA has observed this not working when rke2 is deployed from rancher. What is unclear to me is how it was ever working in stand alone rke2 and if we introduce a regression and if so does that regression affect 1.20 as well


briandowns (approved):

cjellick:
@ibuildthecloud lgtmed offline

cjellick:
RKE2 issue: https://github.com/rancher/rke2/issues/1724

PR #1729, opened 2021-08-30
update airgap images

cjellick:
need to bump the vsphere images we bundle for airgap


rancher-max (approved):
confirmed these are the expected tags -- LGTM

Issue #1611, opened 2021-08-11
Port K3s testing framework into example tests in Rke2


dereknola:
**Is your feature request related to a problem? Please describe.**

Rke2 lack testing
**Describe the solution you'd like**

Three levels of test: unit, integration, and e2e.
Example tests implemented for each.

dereknola:
Closing as this is a test only change.

PR #1782, opened 2021-09-09
Add integration test framework and documentation into RKE2

dereknola:
#### Proposed Changes ####
- Example integration test into RKE2 (etcd snapshot test)

#### Types of Changes ####
New Feature

#### Verification ####
See `testing.md` in `developer docs` to run tests.
Use `make dev-shell` and launch and rke2 server
Once server has started, launch tests via `go tests ./tests/`

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1611

#### Further Comments ####
As RKE2 does not launch as fast as K3s, it was decided that having each test launch and kill a RKE2 server would be time prohibited. It is left to the user to deploy the appropriate server for CLI testing.

### DIFF
--- a/tests/util/cmd.go
+++ b/tests/util/cmd.go
@@ -55,6 +55,7 @@ func Rke2Ready() bool {
 		reg := pod + ".+Running"
 		match, err := regexp.MatchString(reg, pods)
 		if !match || err != nil {
+			logrus.Error(err)
 			return false
 		}
 	}
@@ -72,7 +73,11 @@ func contains(source []string, target string) bool {
 
 // ServerArgsPresent checks if the given arguments are found in the running k3s server
 func ServerArgsPresent(neededArgs []string) bool {
-	currentArgs := Rke2ServerArgs()
+	currentArgs, err := Rke2ServerArgs()
+	if err != nil {
+		logrus.Error(err)
+		return false
+	}
 	for _, arg := range neededArgs {
 		if !contains(currentArgs, arg) {
 			return false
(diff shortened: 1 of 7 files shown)


tests/util/cmd.go
```diff
@@ -0,0 +1,97 @@
...
+func Rke2ServerArgs() []string {
+	results, err := Rke2Cmd("kubectl", "get", "nodes", "-o", `jsonpath='{.items[0].metadata.annotations.rke2\.io/node-args}'`)
+	if err != nil {
+		return nil
+	}
+	res := strings.ReplaceAll(results, "'", "")
+	var args []string
+	if err := json.Unmarshal([]byte(res), &args); err != nil {
+		logrus.Error(err)
+		return nil
```
briandowns, line 94: Should we return the error here rather than return an empty slice?

tests/util/cmd.go
```diff
@@ -0,0 +1,97 @@
...
+		"kube-scheduler-rke2-server",
+		"rke2-coredns-rke2-coredns",
+	}
+	pods, err := Rke2Cmd("kubectl", "get", "pods", "-A")
+	if err != nil {
+		return false
+	}
+	for _, pod := range podsToCheck {
+		reg := pod + ".+Running"
+		match, err := regexp.MatchString(reg, pods)
```
briandowns, line 56: What happens to this err value?

dereknola: Err now gets logged

tests/util/cmd.go
dereknola: function now returns err and caller deals with result properly.

brandond (approved):

briandowns (approved):
LGTM

PR #1827, opened 2021-09-16
Update to Kubernetes / K3s 1.22.1

brandond:
#### Proposed Changes ####

Update to Kubernetes / K3s 1.22.1

#### Types of Changes ####

Minor version update

#### Verification ####

`kubectl get nodes -o wide`

#### Linked Issues ####

* #1549 
*

PR #1838, opened 2021-09-17
1.22 version bumps and use stable cni-plugins release

rosskirkpat:
#### Proposed Changes ####

version bumps for the first rke2 release of 1.22

no docs changed required

#### Types of Changes ####

version bumps only

#### Verification ####

dockerfile builds with new versions

#### Further Comments ####


rosskirkpat:
Reasoning:
- Without this version bump, we would be shipping 1.22 with a 1.21 crictl for Windows
- The Windows dockerfile is still using the 1.16.6b7 build base instead of 1.16.7b7. I believe there are CVEs in 6b7

Dockerfile.windows
```diff
@@ -42,12 +42,12 @@ FROM build as windows-runtime-collect
...
 # windows runtime image
-ENV CRICTL_VERSION="v1.21.0"
-ENV CONTAINERD_VERSION="1.5.4"
+ENV CRICTL_VERSION="v1.22.0"
+ENV CONTAINERD_VERSION="1.5.5"
 ENV WINS_VERSION="0.1.1"
 ENV FLANNEL_VERSION="v0.14.0"
 ENV CALICO_VERSION="v3.19.2"
-ENV CNI_PLUGIN_VERSION="v0.9.1"
+ENV CNI_PLUGIN_VERSION="v1.0.1"
```
manuelbuil, line 50: Two things:
1 - After v0.9.1, cni-plugins repo does not include the flannel binary. IIRC we need the binary in case we deploy cni: canal
2 - the windows version of rke2 does not need to be hardcoded? In linux-rke2 we are using a hardened-cni-plugins version

rosskirkpat: Canal does not support Windows and we are also not currently supporting flannel either. 

We are using upstream CNI plugins as we have no requirements for hardening or security at this time.

These binaries/versions will only ever be deployed on Windows nodes in RKE2 clusters that meets the requirement of using Calico as the CNI.

manuelbuil: Ok! Then no worries :)

phillipsj (approved):
LGTM

brandond:
Please hold off on merging this until https://github.com/rancher/rke2/pull/1827 lands; I'd prefer not to have Windows go to 1.22 before the rest of the project does.

brandond (changes requested):
Please hold off on merging this until https://github.com/rancher/rke2/pull/1827 lands; I'd prefer not to have Windows go to 1.22 before the rest of the project does.

nickgerace:
@brandond is this is a technical limitation, supportability one, or other? Could this PR being merged first cause a breakge?

brandond:
It just feels backwards; the servers should be upgraded before agents. I can merge that PR as it currently stands if that would help. At this point it's all just stuff in rke2-charts that needs to be fixed.

nickgerace:
@brandond that makes sense to me, I was just curious if there was a technical component I was missing here. Thanks for the explanation.

Issue #1074, opened 2021-05-27
kube-controller-manager pod exits on leader node


tiberiu89:
After starting 3 server nodes(via terraform),  I was inspecting system pods and this caught my attention.

![image](https://user-images.githubusercontent.com/42996791/119865646-8f5e9380-bf24-11eb-9779-09396005702a.png)

Seems pod cannot be started due to port 10252 being already in use. This seem to happen always on the node elected as leader.
To be mentioned I tested a lot of creating/destroying clusters via terraform and this particular problem does not always reproduce.

Let me know what other logs I can provide to be of use.

brandond:
This is not something I've seen before, or something that's come up in our QA testing. What is using this port on your nodes?

tiberiu89:
Hmm, weird. I started a new cluster to reproduce this. Not a linux guru here, just starting out with k8s, but I've tried a few commands on the server node
![image](https://user-images.githubusercontent.com/42996791/119874229-0fd5c200-bf2e-11eb-90b2-43e5228762c1.png)

I can see a kube-scheduler process is already started and using the port. But still...

![image](https://user-images.githubusercontent.com/42996791/119874433-44e21480-bf2e-11eb-91da-5068d0b92f8a.png)

As you can see now it happens on 2 server nodes

brandond:
The kube-scheduler process listening on that port - is that the crashlooping kube-scheduler process? Or do you have some other kube-scheduler deployed on this node?

tiberiu89:
Sry, I'm fairly new to this, just starting out. The RKE2 installation is fresh using AWS base ubuntu 20.04 machines. Nothing additionally deployed, everything is pristine. Is kube-scheduler supposed to run as a system process or pod? Anything I can run to check further logs or something? Does it tries to run kube-scheduler twice?

brandond:
Can you get logs from the previous container?  Use `kubectl logs --previous`

tiberiu89:
I just saw this in the docs https://docs.rke2.io/known_issues/#control-groups-v2...Is it still applicable for current version? I just created a custom ubuntu AMI with those kernel params configured in GRUB and at least after creating clusters twice it seems the problem went away.

brandond:
No, RKE2 has been shipping containerd 1.4 with cgroupv2 support for a while, so newer releases are not affected by that.

[...]

PR #1841, opened 2021-09-17
Add Permit Port Sharing Flag to Scheduler and Controller Manager

briandowns:
#### Verification ####

```sh
kubectl describe pod <kube-scheduler-name> | grep permit
```

#### Linked Issues ####

* #1074

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -276,7 +276,7 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
 	return auth, http.NotFoundHandler(), err
 }
 
-const permitPortSharingFlag = "--permit-port-sharing=true"
+var permitPortSharingFlag = []string{"--permit-port-sharing=true"}
 
 // Scheduler starts the kube-scheduler static pod, once the apiserver is available.
 func (s *StaticPodConfig) Scheduler(apiReady <-chan struct{}, args []string) error {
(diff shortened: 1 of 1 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -276,6 +276,8 @@ func (s *StaticPodConfig) APIServer(ctx context.Context, etcdReady <-chan struct
 	return auth, http.NotFoundHandler(), err
 }
 
+const permitPortSharingFlag = "--permit-port-sharing=true"
```
brandond, line 279:
```suggestion
const permitPortSharingFlag = []string{"--permit-port-sharing=true"}
```

pkg/podexecutor/staticpod.go
```diff
@@ -348,6 +351,8 @@ func (s *StaticPodConfig) ControllerManager(apiReady <-chan struct{}, args []str
 		}
 		args = append(extraArgs, args...)
 	}
+	args = append(args, permitPortSharingFlag)
```
brandond, line 354: nit: This needs to be done in the reverse order (appending args to the args added here) to ensure that args added by the user come last and supersede any of our defaults.
```suggestion
	args = append(permitPortSharingFlag, args...)
```

pkg/podexecutor/staticpod.go
```diff
@@ -292,6 +294,7 @@ func (s *StaticPodConfig) Scheduler(apiReady <-chan struct{}, args []string) err
 	if s.ControlPlaneResources.KubeSchedulerCPURequest == "" {
 		s.ControlPlaneResources.KubeSchedulerCPURequest = defaultKubeSchedulerCPURequest
 	}
+	args = append(args, permitPortSharingFlag)
```
brandond, line 297: same nit

brandond (changes requested):

brandond (approved):

galal-hussein (approved):

PR #1859, opened 2021-09-20
Add release validation

galal-hussein:
#### Proposed Changes ####

Add release validation script, the script check for the following in order:

- The release branch: it makes sure that release branch matches the release tag version
- The kubernetes image tag: it makes sure that image tag matches the release tag version
- The kubernetes version: it makes sure that kubernetes version env variable matches the release tag version since its used in building both the runtime and windows runtime images as well as the test image
- The Kube-proxy version: it makes sure that kube proxy chart version matches the release tag version
- crictl and calico windows binaries: it makes sure that both binaries matches the Linux versions

#### Types of Changes ####

CI Feature


dereknola (approved):

scripts/validate-release
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
briandowns, line 1: Should we stay consistent with the rest of the scripts?

scripts/validate-release
```diff
@@ -0,0 +1,76 @@
...
+set -e
+
+BUILD_REGEX="build[0-9]+"
+
+info() {
```
briandowns, line 6: At some point we're gonna need a shell package to source these types of things in since they're used in some many placed. No action. Just sayin... :)

briandowns (approved):

Issue #1940, opened 2021-10-08
[Release-1.19] Upgrade Nginx Ingress chart


galal-hussein:
**Environmental Info:**
RKE2 Version: 

1.19.x

**Describe the bug:**
Nginx chart is outdated in 1.19, we need to upgrade nginx chart to 1.0.x version

PR #1943, opened 2021-10-08
[Release-1.19] Upgrade nginx chart to version 4.0.305

galal-hussein:
#### Proposed Changes ####

Upgrade Nginx ingress to version 4.0.3

#### Types of Changes ####

chart upgrade

#### Verification ####

- Install rke2 cluster

#### Linked Issues ####

- https://github.com/rancher/rke2/issues/1940

### DIFF
--- a/scripts/test-run-basics
+++ b/scripts/test-run-basics
@@ -11,7 +11,6 @@ all_services=(
     kube-scheduler
     metrics-server
     rke2-ingress-nginx-controller
-    rke2-ingress-nginx-default-backend
 )
 
 export NUM_SERVERS=1
--- a/scripts/test-setup-sonobuoy
+++ b/scripts/test-setup-sonobuoy
@@ -11,7 +11,7 @@ all_services=(
     kube-scheduler
     metrics-server
 )
-export SERVER_ARGS='--disable=rke2-ingress-nginx-chart --kube-apiserver-arg=kubelet-preferred-address-types=InternalIP'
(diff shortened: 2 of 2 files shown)


scripts/build-images
```diff
@@ -21,7 +21,7 @@ xargs -n1 -t docker image pull --quiet << EOF >> build/images-core.txt
     docker.io/rancher/klipper-helm:v0.4.3-build20210225
     docker.io/rancher/pause:3.2
     docker.io/rancher/nginx-ingress-controller-defaultbackend:1.5-rancher1
```
brandond, line 23: Can drop this as it's not used with the newer chart.

brandond (changes requested):

scripts/build-images
galal-hussein: its just disabled in the new chart but if the user decided to enable it still needs the image

brandond: We don't include it in 1.21 which is on the new-style chart, so either we need to re-add it there, or remove it from here.
https://github.com/rancher/rke2/blob/release-1.21/scripts/build-images#L24

brandond (changes requested):

scripts/build-images
galal-hussein: yeah I think I will readd it in 1.21 and master, and will add webhook image here

brandond (approved):

PR #1951, opened 2021-10-12
Fix validate release for tags cut from master

galal-hussein:
#### Proposed Changes ####

- Fixing tags that are cut from master
- Skip check for kube proxy version from master

#### Types of Changes ####
bugfix

### DIFF
--- a/scripts/validate-release
+++ b/scripts/validate-release
@@ -31,7 +31,7 @@ function check_release_branch() {
     TAG_BRANCH=$(git branch --all -q --contains $GIT_TAG | grep origin | grep -v origin$ | grep -v "HEAD" | sed -e 's/^[[:space:]]*//')
     if [ "$TAG_BRANCH" == "remotes/origin/master" ]; then
         K8S_VERSION_GO_MOD=$(grep 'k8s.io/kubernetes v' go.mod | head -n1 | awk '{print $2}')
-        if [ "$MAJOR.$MINOR" == "$K8S_VERSION_GO_MOD" ]; then
+        if [ "v$MAJOR.$MINOR.$PATCH" == "$K8S_VERSION_GO_MOD" ]; then
             info "Tag $GIT_TAG is cut from master"
             return
         fi


scripts/validate-release
```diff
@@ -28,7 +28,11 @@ function parse_tag() {
 }
 
 function check_release_branch() {
-    TAG_BRANCH=$(git branch --all -q --contains $GIT_TAG | grep -v origin$ | grep -v "HEAD" | sed -e 's/^[[:space:]]*//')
+    TAG_BRANCH=$(git branch --all -q --contains $GIT_TAG | grep origin | grep -v origin$ | grep -v "HEAD" | sed -e 's/^[[:space:]]*//')
+    if [ "$TAG_BRANCH" == "remotes/origin/master" ] && [ "$MAJOR.$MINOR" == "1.22"  ]; then
```
brandond, line 32: Hmm, I wonder if we can avoid hardcoding the Kubernetes version that the master branch is on. Maybe if we're on master then we grab the version out of go.mod and check to see if that matches the tag?

`grep 'k8s.io/kubernetes v' go.mod | head -n1 | awk '{print $2}'`

brandond (changes requested):

scripts/validate-release
```diff
@@ -28,7 +28,14 @@ function parse_tag() {
 }
 
 function check_release_branch() {
-    TAG_BRANCH=$(git branch --all -q --contains $GIT_TAG | grep -v origin$ | grep -v "HEAD" | sed -e 's/^[[:space:]]*//')
+    TAG_BRANCH=$(git branch --all -q --contains $GIT_TAG | grep origin | grep -v origin$ | grep -v "HEAD" | sed -e 's/^[[:space:]]*//')
+    if [ "$TAG_BRANCH" == "remotes/origin/master" ]; then
+        K8S_VERSION_GO_MOD=$(grep 'k8s.io/kubernetes v' go.mod | head -n1 | awk '{print $2}')
+        if [ "$MAJOR.$MINOR" == "$K8S_VERSION_GO_MOD" ]; then
```
brandond, line 34: K8S_VERSION_GO_MOD is going to be v1.22.2 or something like that, so it won't compare properly. Maybe try using:
`K8S_VERSION_GO_MOD=$(grep 'k8s.io/kubernetes v' go.mod | awk -F. 'NR==1 {print "1." $3}')`

or compare to $MAJOR.$MINOR.$PATCH

brandond (changes requested):

PR #2036, opened 2021-10-26
Removed rke2-kube-proxy disable list option from documentation

PhilipSchmid:
... since it's not valid anymore. Use `disable-kube-proxy: true` instead.

#### Types of Changes ####

Documentation

#### Linked Issues ####

See https://github.com/rancher/rancher/issues/33785.


manuelbuil:
> This change is not reflected in the docs on this page. https://docs.rke2.io/advanced/

Thanks! I also found a related problem

PR #2084, opened 2021-11-07
testing: assert that install.sh works correctly

dweomer:
For every change that modifies the installer script, run some quick assertions against:
- CentOS 7
- CentOS 8
- MicroOS

Thanks, Github Actions and Vagrant!

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -1,0 +1,31 @@
+name: Installer
+on:
+  push:
+    paths:
+      - "install.sh"
+      - "tests/vagrant/install/**"
+  pull_request:
+    paths:
+      - "install.sh"
+      - "tests/vagrant/install/**"
+  workflow_dispatch: {}
+jobs:
+  vagrant:
+    name: "Vagrant"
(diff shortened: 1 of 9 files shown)


dweomer:
[Will It Blend](https://www.youtube.com/willitblend)

dweomer:
Flakes are detrimental to my chill.

tests/install/centos7/README.md
```diff
@@ -0,0 +1,46 @@
+RKE2 Install on CentOS 8
```
dereknola, line 1: All of theses READMEs are the same except for the OS that is being tested, can you make 1 document either under the `tests/install` folder or add a section to https://github.com/rancher/rke2/blob/master/developer-docs/testing.md

dereknola (changes requested):

tests/install/centos7/README.md
dweomer: You'll notice that the MicroOS has additional details. I expect these to diverge per OS over time and chose to not assume that someone looking to run some quick tests will know how to navigate our documentation.

dereknola (approved):

Issue #2091, opened 2021-11-09
[Docs] Add documentation for HTTP proxy configuration


bashofmann:
**Is your feature request related to a problem? Please describe.**
The docs are missing a section on how and where to configure HTTP proxy settings for environments, where you only have connectivity through a proxy.

**Additional context**
Related issues:
* https://github.com/rancher/rke2/issues/517
* https://github.com/rancher/rke2/issues/1834

gtato:
@bashofmann Hi, do you know how we can configure the http proxy for windows agents?

bashofmann:
I assume, you have to also set these as environment variables when starting the `rke2.exe`. But I have no idea how to do this  on Windows.

gtato:
Yup, I just did the following and it worked.
```
[Environment]::SetEnvironmentVariable("HTTP_PROXY", "http://ip:port", [EnvironmentVariableTarget]::Machine)
[Environment]::SetEnvironmentVariable("HTTPS_PROXY", "http://ip:port", [EnvironmentVariableTarget]::Machine)
[Environment]::SetEnvironmentVariable("NO_PROXY", "x,y,z", [EnvironmentVariableTarget]::Machine)

$env:HTTP_PROXY="http://ip:port"
$env:HTTPS_PROXY="http://ip:port"
$env:NO_PROXY="x,y,z"
```

PR #2092, opened 2021-11-09
[Docs] Configuring http proxy settings

bashofmann:
#### Proposed Changes ####

Add documentations on configuring http proxy settings.

#### Types of Changes ####

Docs

#### Linked Issues ####

Addresses https://github.com/rancher/rke2/issues/2091

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -22,6 +22,31 @@ For advanced customization of this file you can create another file called `conf
 
 The `config.toml.tmpl` will be treated as a Go template file, and the `config.Node` structure is being passed to the template. See [this template](https://github.com/k3s-io/k3s/blob/master/pkg/agent/templates/templates.go#L16-L32) for an example of how to use the structure to customize the configuration file.
 
+## Configuring an HTTP proxy
+
+If you are running RKE2 in an environment, which only has external connectivity through an HTTP proxy, you can configure your proxy settings on the RKE2 systemd service. These proxy settings will then be used in RKE2 and passed down to the embedded containerd and kubelet.
+
+Add the necessary `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` variables to the environment file of your systemd service, usually:
+
+- `/etc/default/rke2-server`
+- `/etc/default/rke2-agent`
+
+The `NO_PROXY` variable must include your internal networks, as well as the cluster pod and service IP ranges.
+
+```
+HTTP_PROXY=http://your-proxy.example.com:8888
+HTTPS_PROXY=http://your-proxy.example.com:8888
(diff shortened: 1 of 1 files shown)


Oats87 (approved):

brandond (approved):
Hmm. I thought that we added the cluster CIDRs and dns suffix to the NO_PROXY variable, but it looks like for some reason we only currently do that on server nodes, not agents. Until we fix that, this docs update should cover it.

docs/advanced.md
```diff
@@ -22,6 +22,23 @@ For advanced customization of this file you can create another file called `conf
 
 The `config.toml.tmpl` will be treated as a Go template file, and the `config.Node` structure is being passed to the template. See [this template](https://github.com/k3s-io/k3s/blob/master/pkg/agent/templates/templates.go#L16-L32) for an example of how to use the structure to customize the configuration file.
 
+## Configuring an HTTP proxy for containerd
```
brandond, line 25:
```suggestion
## Configuring an HTTP proxy
```

docs/advanced.md
```diff
@@ -22,6 +22,23 @@ For advanced customization of this file you can create another file called `conf
...
+- `/etc/default/rke2-agent`
+
+The `NO_PROXY` variable must include your internal networks, as well as the cluster pod and service IP ranges.
+
+```
+HTTP_PROXY=http://your-proxy.example.com:8888
+HTTPS_PROXY=http://your-proxy.example.com:8888
+NO_PROXY=127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
+```
+
```
brandond, line 41: You can also prefix environment variables with CONTAINERD_ and they will be passed through to containerd only, with the prefix removed. If your intent is to set the proxy variables just for containerd, that's probably a better way to do it. Would you mind adding this to the documentation as well? The currently recommended variables will set it for both RKE2, the Kubelet, and containerd.

brandond (changes requested):

docs/advanced.md
bashofmann: I added a section for this.

docs/advanced.md
```diff
@@ -22,6 +22,31 @@ For advanced customization of this file you can create another file called `conf
...
+If you just want to configure the proxy settings for containerd to pull container images through the proxy, you can alsoe prefix these variables with `CONTAINERD_`:
```
brandond, line 42:
```suggestion
If you want to configure the proxy settings for containerd without affecting RKE2 and the Kubelet, you can prefix the variables with `CONTAINERD_`:
```

brandond (changes requested):
nit

brandond (approved):

Issue #1755, opened 2021-09-03
Processes should stop when stopping rke2 process on windows agent


rancher-max:
**Environmental Info:**
RKE2 Version: 

v1.21.4-rc1+rke2r3

Node(s) CPU architecture, OS, and Version: 

Windows 2019

Cluster Configuration:

1 server 1 windows agent

**Describe the bug:**

When running the rke2 agent as a service, and then performing `Stop-Process rke2`, I expect the kubelet, kube-proxy, and containerd processes to also stop, similar as when stopping the rke2 binary process.

**Steps To Reproduce:**

- Install rke2 on windows node
- Run the rke2 agent as a service
- Stop the rke2 process

**Expected behavior:**

rke2, kubelet, kube-proxy, and containerd processes should all stop

**Actual behavior:**

Only rke2 process is stopped

**Additional context / logs:**

rancher-max:
When validating this, we should also bring up linux servers/agents and validate the running killall, uninstall, and cluster-reset all continue to function properly

brandond:
As well as just basic service start/stop behavior.

rancher-max:
This issue mentions the last step as: `Stop-Process rke2`, but this is actually not the correct behavior. In this case, we should do `Stop-Service rke2` (stopping the SERVICE, not the PROCESS), as this is what most Windows Admins will do anyway.

mdrahman-suse:
Tested rke2 using v1.21.6-rc2+rke2r2 and I am not seeing any other services (kube-proxy, kubelet, containerd, etc) other than rke2 getting stopped when I perform `Stop-Service rke2`. Please advice @luthermonson @rosskirkpat 
CC: @rancher-max

nickgerace:
Fleet `v0.3.8-rc2` (now on Rancher `v2.6-head`) contains the Wrangler module bump needed for this issue: https://github.com/rancher/fleet/pull/615

mdrahman-suse:
After debugging with @luthermonson, we came to conclusion that there are some additional code changes requires for this to fully work and it is working better than before and satisfies the scope of this issue. New issue will be raised and tracked for the additional work. At the time of testing this ticket it was observed that `Stop-Service rke2` stops most of the services used by rke2, except for one or two like kubelet and kube-proxy. For now they are stopped manually via `Stop-Process -Name <service>`

PR #2098, opened 2021-11-09
[master] updating to new signals package in wrangler

luthermonson:
#### Proposed Changes ####
Updates to using wrangler v0.8.8 which includes fixes to the signal handler to give a exported func to cancel the the context by hand. this is useful in windows services and gives us a way to cancel context to kill all sub-processes and exit gracefully when using windows service manager.

#### Types of Changes ####
vendor update, small code change to align k3s to use library slight different and pass in context to containerd

#### Verification ####
this will ultimately be used in rke2 windows services, this will be tested by QA when we work on closing the linked issue.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1755

#### User-Facing Change ####
none

### DIFF
--- a/pkg/windows/service_windows.go
+++ b/pkg/windows/service_windows.go
@@ -6,10 +6,13 @@ import (
 	"os"
 	"time"
 
+	"golang.org/x/sys/windows"
+
 	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/rancher/wins/pkg/logs"
 	"github.com/rancher/wins/pkg/profilings"
+	"github.com/rancher/wrangler/pkg/signals"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/windows/svc"
 )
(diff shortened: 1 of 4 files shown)


pkg/windows/service_windows.go
```diff
@@ -23,19 +26,29 @@ func (h *service) Execute(_ []string, requests <-chan svc.ChangeRequest, statuse
...
-			time.Sleep(5 * time.Second)
+			if !signals.RequestShutdown() {
+				logrus.Infof("Windows Service is shutting down")
+				statuses <- svc.Status{State: svc.Stopped}
+				os.Exit(0)
+			}
+
+			logrus.Infof("Windows Service is shutting down gracefully")
+			time.Sleep(10 * time.Second) // give context time to cancel
+			statuses <- svc.Status{State: svc.StopPending}
```
brandond, line 43: Go to pending for the duration of the sleep?
```suggestion
			statuses <- svc.Status{State: svc.StopPending}
			time.Sleep(10 * time.Second) // give context time to cancel
```

luthermonson: it literally never gets to this sleep tbh, once the context is cancelled it almost immediately exits as rke2 catches that cancel and exists.  that status just tells service manager it will stop soon and makes service manager not try any harder than it already did to cancel it. i could remove this sleep and it would probably behave exactly the same, this is basically the time i realized we should do a wait group or something

brandond: ah ok. I was just curious why we're transitioning from pending to stopped without any pause in between.

dweomer:
This passed for me locally:

dweomer (approved):

brandond (approved):

PR #2123, opened 2021-11-15
[release-1.21] version bumps for 1.21.7 prep

rosskirkpat:
#### Proposed Changes ####

- Remove flannel and wins from the Windows runtime file
- bump to dapper 0.5.8
- bump containerd to [1.5.7](https://github.com/containerd/containerd/releases/tag/v1.5.7) to remediate a [CVE](https://github.com/advisories/GHSA-c2h3-6mxw-7mvq)

#### Types of Changes ####

Version bumps to prep for 1.21.7 

#### Verification ####

Source of truth is in dockerfiles and drone 

#### Linked Issues ####

https://github.com/rancher/rke2/issues/2124

### DIFF
--- a/pkg/windows/service_linux.go
+++ b/pkg/windows/service_linux.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package windows
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,7 +9,7 @@ platform:
 
 steps:
   - name: validate-release
-    image: rancher/dapper:v0.5.5
+    image: rancher/dapper:v0.5.8
     commands:
       - docker pull --quiet rancher/hardened-build-base:v1.16.10b7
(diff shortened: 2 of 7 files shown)


luthermonson (approved):

dweomer (approved):

pkg/windows/service_linux.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
brandond, line 1: Do we need to migrate our build tags over to a new format or something?

rosskirkpat: Yes, I believe so. I bumped the version of golanglint-ci to 1.43 and it failed out on the old build tags. With go1.16.9 binary, running a go fmt on these files doesn't fix the build tags. I believe that 1.43.0 is using go1.17 which explains why we are suddenly seeing this.

PR #2163, opened 2021-11-23
Get node name from metadata if AWS cloud provider is enabled

thedadams:
The AWS cloud provider expects the node name of a node to have a
specific form. If the user sets the hostname of the of the underlying
machine to a custom name, then the node will fail to bootstrap because
the node will not have the expected name.

With this change, the node name is fetched from the metadata endpoint to
ensure it has the proper value. If this fails, then it will fallback to
the hostname command.

### DIFF
--- a/pkg/rke2/rke2_linux.go
+++ b/pkg/rke2/rke2_linux.go
@@ -1,13 +1,18 @@
+//go:build linux
 // +build linux
 
 package rke2
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/agent/config"
@@ -62,9 +67,13 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 			Path: cfg.CloudProviderConfig,
 		}
 		if clx.String("node-name") == "" && cfg.CloudProviderName == "aws" {
-			fqdn, err := hostnameFQDN()
-			if err != nil {
-				return nil, err
+			fqdn := hostnameFromMetadataEndpoint(context.Background())
+			if fqdn == "" {
+				hostFQDN, err := hostnameFQDN()
+				if err != nil {
+					return nil, err
+				}
+				fqdn = hostFQDN
 			}
 			if err := clx.Set("node-name", fqdn); err != nil {
 				return nil, err
@@ -198,4 +207,35 @@ func hostnameFQDN() (string, error) {
 	}
 
 	return strings.TrimSpace(b.String()), nil
+}
+
+func hostnameFromMetadataEndpoint(ctx context.Context) string {
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://169.254.169.254/latest/metadata/local-hostname", nil)
+	if err != nil {
+		logrus.Debugf("Failed to create request for metadata endpoint: %v", err)
+		return ""
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		logrus.Debugf("Failed to get local-hostname from metadata endpoint: %v", err)
+		return ""
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		logrus.Debugf("Metadata endpoint returned unacceptable status code %d", resp.StatusCode)
+		return ""
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		logrus.Debugf("Failed to read response body from metadata endpoint: %v", err)
+		return ""
+	}
+
+	return strings.TrimSpace(string(b))
 }


pkg/rke2/rke2_linux.go
```diff
@@ -199,3 +205,22 @@ func hostnameFQDN() (string, error) {
...
+
+func hostnameFromMetadataEndpoint() (string, error) {
+	resp, err := http.Get("http://169.254.169.254/latest/metadata/local-hostname")
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unexpected status code: %d", resp.StatusCode)
```
briandowns, line 217: Is there anything more descriptive that we could put in here??

pkg/rke2/rke2_linux.go
```diff
@@ -199,3 +205,22 @@ func hostnameFQDN() (string, error) {
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
briandowns, line 220: Generic return of bytes is conventionally referred to with a variable named `b`.

```suggestion
	b, err := ioutil.ReadAll(resp.Body)
```

pkg/rke2/rke2_linux.go
```diff
@@ -62,9 +65,12 @@ func initExecutor(clx *cli.Context, cfg Config, dataDir string, disableETCD bool
 			Path: cfg.CloudProviderConfig,
 		}
 		if clx.String("node-name") == "" && cfg.CloudProviderName == "aws" {
-			fqdn, err := hostnameFQDN()
+			fqdn, err := hostnameFromMetadataEndpoint()
 			if err != nil {
-				return nil, err
+				fqdn, err = hostnameFQDN()
```
briandowns, line 70: What happens to this error value?

briandowns: If we're logging relevant errors in the function itself and what gets returned is less meaningful and can be ignore, remove the error from the return and check here for an empty string, and do the additional call to `hostnameFQDN`.

thedadams: It gets return below if it is non-nil.

pkg/rke2/rke2_linux.go
thedadams: Will update.

pkg/rke2/rke2_linux.go
thedadams: I can put something more descriptive, but this error is never outputted to the user in the current implementation because if this function returns an error then we use `hostnameFQDN`.

briandowns: Let's remove this then if it's never going to happen.

pkg/rke2/rke2_linux.go
briandowns: Implicit reuse of err vals can lead to problematic code. It's generally best to `:=` your errors.

Oats87 (approved):

pkg/rke2/rke2_linux.go
thedadams: Updated.

pkg/rke2/rke2_linux.go
thedadams: Updated.

pkg/rke2/rke2_linux.go
thedadams: Updated.

briandowns (approved):

Oats87:
@galal-hussein Just so you can see this because I think you made these changes before -- do you have any concerns with this?

pkg/rke2/rke2_linux.go
```diff
@@ -199,3 +206,22 @@ func hostnameFQDN() (string, error) {
...
+
+func hostnameFromMetadataEndpoint() string {
+	resp, err := http.Get("http://169.254.169.254/latest/metadata/local-hostname")
+	if err != nil {
+		return ""
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return ""
```
galal-hussein, line 218: nit: can we leave a warning or a debug log to just report the error if its there

galal-hussein (approved):

pkg/rke2/rke2_linux.go
```diff
@@ -199,3 +206,22 @@ func hostnameFQDN() (string, error) {
 
 	return strings.TrimSpace(b.String()), nil
 }
+
+func hostnameFromMetadataEndpoint() string {
+	resp, err := http.Get("http://169.254.169.254/latest/metadata/local-hostname")
```
brandond, line 211: Access to the IMDS is usually done with a 1 second timeout to ensure that it doesn't hang indefinitely if the code isn't run on AWS. I think you need to pass through + wrap the context from initExecutor, and use `NewRequestWithContext` here instead of just Get.

brandond (changes requested):

pkg/rke2/rke2_linux.go
brandond: +1 on debug logging it.

pkg/rke2/rke2_linux.go
thedadams: Updated.

brandond (approved):

pkg/rke2/rke2_linux.go
thedadams: Updated.

thedadams:
The `*cli.Context` does not implement the `context.Context` interface, so I had to pass `context.Backgroup()` instead.

thedadams:
The `*cli.Context` does not implement the `context.Context` interface, so I had to pass `context.Backgroup()` instead.

brandond (approved):

brandond:
@thedadams I am guessing this needs to go back to release-1.21, and potentially release-1.20, as well?

thedadams:
Just 1.21, I believe. RKE2 provisioning from Rancher only supports RKE2 version 1.21+

PR #2171, opened 2021-11-25
Add s390x support

raulcabello:
Signed-off-by: raulcabello <raul.cabello@suse.com>

#### Proposed Changes ####

Add support for s390x

#### Further Comments ####

- No airgap for s390x
- Move manifest step to a new drone pipeline, so it also add s390x image
- Bump versions because new version are available on s390x
- Remove socat as it was dropped by containerd 1.4


thehejik (approved):
It was tested manually on s390x hardware, everything works as expected.

scripts/build-binary
```diff
@@ -24,7 +24,7 @@ VERSION_FLAGS="
         -X ${K3S_PKG}/pkg/version.Program=${PROG}
         -X ${K3S_PKG}/pkg/version.Version=${VERSION}
         -X ${RKE2_PKG}/pkg/images.DefaultRegistry=${REGISTRY}
-        -X ${RKE2_PKG}/pkg/images.DefaultEtcdImage=rancher/hardened-etcd:${ETCD_VERSION}-${IMAGE_BUILD_VERSION}
+        -X ${RKE2_PKG}/pkg/images.DefaultEtcdImage=rancher/hardened-etcd:${ETCD_VERSION}-build20211118
```
dweomer, line 27: We should retain the usage of `${IMAGE_BUILD_VERSION}` should we not?

scripts/build-s390x
```diff
@@ -0,0 +1,12 @@
...
+
+cd $(dirname $0)/..
+
+source ./scripts/version.sh
+mkdir -p build/images
+
+./scripts/build-binary
+./scripts/build-image-runtime
+./scripts/package-binary
+./scripts/package-bundle
```
dweomer, line 12: is this just a copy of scripts/build? with all of these other changes is `make build` on an s390x host not sufficient?

scripts/build-binary
raulcabello: I changed it because I saw in master was like this https://github.com/rancher/rke2/blob/2faa95b6a8b3a3abe12fb90c4529f7aba29803da/scripts/build-binary#L27

scripts/build-s390x
raulcabello: it is not a copy. We don't want to build window images and we don't support airgap on s390x

scripts/build-binary
brandond: It made sense to have a variable when it looked like we were going to build all the images at the same time, but now with things being as ad-hoc as they are - many images have different build tags -  it's not really getting used as I thought it would. I'm OK with dropping it.

scripts/build-s390x
brandond: Do we have a design document that captures the omission of airgap support for this architecture? That seems like an important thing to casually omit.

brandond (changes requested):
LGTM, once the docs are updated to cover lack of airgap support for s390x (if that is OK as per Product).

briandowns (approved):
LGTM but would like to see a docs update like @brandond mentioned.

agracey:
PM here. For now, the lack of airgap is expected. We will revisit in a few months after a few POCs

agracey:
To clarify. We need air gap for our GA which we are hoping to achieve in March. 

What is the amount of work required to get it now?

scripts/build-s390x
agracey: What is missing here to provide air gap? I was under the impression that air gap was effectively pulling the images down locally and "side-loading" them into the server(s) or a local registry. Is there more to it than that?

brandond: I would probably recommend
* Putting the windows bits of `./scripts/build` behind an architecture check, or perhaps modifying the various `scripts/*-windows-*` scripts to exit early on non-amd64 architectures so that they become a no-op.
* Modifying `scripts/build-images` to skip building airgap tarballs for the optional components on non-amd64 architectures.
Everything else should already work, as far as I can tell?

raulcabello: Yes, that should work. Thanks!

raulcabello:
@agracey it shouldn’t take long to add airgap for s390x. But I won’t be able to do it until next Tuesday, and I think code freeze is next week.
The work to do it in a month should be the same as doing it now. The only extra thing we would need to do is to update the docs to say we don’t support airgap for s390x. Or mention somewhere airgap is not available for the technical preview.

Makefile
```diff
@@ -55,6 +55,10 @@ build-image-kubernetes:                   ## Build the kubernetes image
 build-image-runtime:                      ## Build the runtime image
 	./scripts/build-image-runtime
 
+.PHONY: build-s390x
+build-s390x:                                   ## Build s390x
+	./scripts/build-s390x
+
```
brandond, line 61: I think this can go away now that the script has been removed

scripts/build
```diff
@@ -5,11 +5,14 @@ cd $(dirname $0)/..
...
-./scripts/package-windows-bundle
-./scripts/dev-runtime-image
-./scripts/build-image-test
+if [ "${GOARCH}" != "s390x" ]; then
+  ./scripts/build-windows-binary
+  ./scripts/build-windows-images
+  ./scripts/package-windows-bundle
+  ./scripts/dev-runtime-image
+  ./scripts/build-image-test
+fi
```
brandond, line 18: I am not confident that we can reorder these like this without breaking some of the build dependencies in ways that will not cause tests or the pipeline to fail, but may cause some artifacts to not be published. I'm also curious why we wouldn't want dev-runtime-image and build-image-test on s390x.

Can you leave the order here as-is, and add arch checks to the Windows scripts that just immediately `exit 0` on the wrong arch?

brandond (changes requested):

Dockerfile
```diff
@@ -113,7 +117,8 @@ RUN rm -vf /charts/*.sh /charts/*.md
 # This image includes any host level programs that we might need. All binaries
 # must be placed in bin/ of the file image and subdirectories of bin/ will be flattened during installation.
 # This means bin/foo/bar will become bin/bar when rke2 installs this to the host
-FROM rancher/hardened-kubernetes:v1.21.7-rke2r2-build20211206 AS kubernetes
+
```
brandond, line 120: Why the extra newline?

I think we also need to bump the image in the rke2-kube-proxy chart, and then bump the chart version up above.

brandond (changes requested):

Dockerfile
raulcabello: https://github.com/rancher/rke2-charts/pull/196 just bumped the chart, can you have a look?

Makefile
raulcabello: done in https://github.com/rancher/rke2/pull/2171/commits/133611b8d82876240448ae1357cdf8a86f60fe60

scripts/build
raulcabello: done in https://github.com/rancher/rke2/pull/2171/commits/75921506838b7dd92924ea953168af0522501cc8

Dockerfile
raulcabello: bump and remove line https://github.com/rancher/rke2/pull/2171/commits/c229aa7a368714892b692352324a38eb0e94a316

raulcabello: ah build failed because it could not fin chart https://drone-pr.rancher.io/rancher/rke2/2658/1/2 
Should pass after https://github.com/rancher/rke2-charts/pull/196 is merged

scripts/build-images
```diff
@@ -16,9 +16,9 @@ xargs -n1 -t docker image pull --quiet << EOF >> build/images-core.txt
     ${REGISTRY}/rancher/hardened-coredns:v1.8.5-build20211119
     ${REGISTRY}/rancher/hardened-cluster-autoscaler:v1.8.5-build20211119
     ${REGISTRY}/rancher/hardened-dns-node-cache:1.21.2-build20211119
-    ${REGISTRY}/rancher/hardened-etcd:${ETCD_VERSION}-${IMAGE_BUILD_VERSION}
```
brandond, line 19: if there's nothing left using `IMAGE_BUILD_VERSION` can you also remove it from version.sh to avoid confusion?

brandond (approved):
LGTM, with one nit about cleaning up the now-unused variable.

Make sure this gets squashed before merging?

scripts/build-images
raulcabello: It's not being used. Done in https://github.com/rancher/rke2/pull/2171/commits/68303a64afce935f5c5f0dd387c7c22d341587a1

raulcabello:
@brandond Can you squash it when merging in github? If not I can squash the commits in my local and do a force push. 
Do we need another approval? I don't have permission to merge, so someone else will need to do it.

Issue #2216, opened 2021-12-06
[Linux - Ubuntu] Bug - Uninstall script does not purge systemd files.


8times4:
**Environmental Info:**
RKE2 Version: 

v1.22.4+rke2r1

Node(s) CPU architecture, OS, and Version: 

Ubuntu 20.04 
5.4.0-90-generic

Cluster Configuration:

3m 3a

**Describe the bug:**

Uninstall script does not purge systemd files, effectively making redeployment via ansible fail. 

**Steps To Reproduce:**

- Installed RKE2: 
- run rke2-uninstall.sh on a debian based system.
- try to reinstall rke2 via ansible, pre-installed rke2 will be detected, and because the binaries were removed already, the playbook will fail. 
**Expected behavior:**

- rke2-uninstall should purge /etc/systemd/system/rke2*.service

**Actual behavior:**

- rke2-uninstall does not remove /etc/systemd/system/rke2*.service

**Additional context / logs:**

PR #2217, opened 2021-12-06
[BugFix - Linux - Ubuntu/Debian]Make sure to remove systemd service on debian based systems

8times4:
#### Proposed Changes ####

This patch makes sure to purge the rke2-*.services from systemd on debian based systems, as the current uninstall script does not remove these files, effectively making the rke2-ansible deployment to fail. 
This has been tested on Ubuntu 20.04.

#### Types of Changes ####

Bugfix 

#### Verification ####

Run the script on a debian based system(like Ubuntu), where rke2-*.services are running in /etc/systemd/system.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/2216

#### Further Comments ####

-


briandowns (approved):

Oats87 (approved):
This looks sane

Issue #1959, opened 2021-10-15
Request apiserver failed after deleting one of the master nodes


yaocw2020:
**Environmental Info:**
RKE2 Version: 

```
rke2 version v1.21.5+rke2r2 (9e4acdc6018ae74c36523c99af25ab861f3884da)
go version go1.16.6b7

CNI: canal
kube-proxy: iptables mode
```

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

After deleting the master node, some pods will fail to request the apiserver because TCP Conntrack entries remain stale.

**Additional context / logs:**

Related Kubernetes issue: https://github.com/kubernetes/kubernetes/issues/101607

brandond:
Do we need a tracking issue here if it's an upstream issue? Is there some action needed on our side?

gitlawr:
I believe it is related to the way rke2 runs apiservers. It does not happen in K3S.
Given that this is critical to RKE2 clusters, do you think we can mitigate the problem from RKE2 side?

cjellick:
@gitlawr any chance you can spin up rancher 2.5.x, provision an rke1 cluster and see if it repros there? Im just curious

cjellick:
@gitlawr can you expand on this:
> Delete the node that has the most contrack table entries.

What do you mean by "delete the node"? Is that `kubectl delete node` or delete the vm from aws or something else?

brandond:
In my case I left the node running. I'll see if stopping the service first and/or shutting down the node makes a difference.

brandond:
Just deleting the static pod manifests on the deleted node seems to fix things. Having RKE2 do this when it detects an etcd tombstone file might be a suitable workaround, although this relies on RKE2 being online and working in order to create the tombstone file, which makes it not a completely reliable fix for this issue.

It takes a moment for the kubelet to properly tear down all the static pods, and there are a bunch of errors, but it does eventually work. The orphaned workload pods keep running, which probably isn't great, but having at least the control-plane static pods gone seems to allow the rest of the cluster to function properly.

```console
root@ubuntu01:~# crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                            ATTEMPT             POD ID
be874f5df2911       7589738b9ae11       8 minutes ago       Running             coredns                         0                   51fe4a0225e70
0617a01f36c1b       69bca36d82432       10 minutes ago      Running             system-upgrade-controller       0                   b1b86b8c964c4
[...]

PR #2249, opened 2021-12-10
Add startup hook to watch/delete static pods on node delete

brandond:
#### Proposed Changes ####

Add a startup hook that starts a goroutine watching for deletion of the local node. If the local node is deleted, delete the static pod manifests so that the control-plane components will be cleaned up by the kubelet.

I think this is the best we can reasonably do. If the user stops RKE2 before deleting the node, but does not also run the rke2-killall script, that is an error on their part; RKE2 can't reasonably be expected to clean up after itself when it's not running.

#### Types of Changes ####

bugfix

#### Verification ####

1. Create multi-node server cluster
2. Delete one of the servers from the cluster
3. Note that the control-plane static pods manifests on the deleted server are gone from disk, and the static pods are stopped within a few minutes.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/1959#issuecomment-947362697

### DIFF
--- a/pkg/rke2/spc.go
+++ b/pkg/rke2/spc.go
@@ -1,0 +1,88 @@
+package rke2
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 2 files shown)


briandowns (approved):

dereknola (approved):

pkg/rke2/spc.go
```diff
@@ -0,0 +1,83 @@
...
+
+	logrus.Infof("Local Node deleted, cleaning up server static pods")
+	if err := cleanupStaticPods(dataDir); err != nil {
+		logrus.Fatalf("spc: failed to clean up static pods: %v", err)
+	}
+}
+
+// cleanupStaticPods deletes all the control-plane and etc static pod manifests.
+func cleanupStaticPods(dataDir string) error {
+	components := []string{"kube-apiserver", "kube-scheduler", "kube-controller-manager", "cloud-controller-manager", "etcd"}
```
Oats87, line 74: I wish we had a centralized list for these components.

Oats87 (approved):

dweomer (approved):

Issue #2288, opened 2021-12-21
On cis hardened cluster , RKE2 should refuse to start if the etcd user and group doesn't exist on the host


ShylajaDevadiga:
**Environmental Info:**
RKE2 Version: 
Any version
v1.20.14-rc2+rke2r1, v1.21.8-rc2+rke2r1, v1.22.5-rc1+k3s1

Node(s) CPU architecture, OS, and Version: 
SLES 15 SP2

Cluster Configuration:
3 server, 1 agent CIS enabled cluster
profile: cis-1.5
profile: cis-1.6

**Describe the bug:**
As per the docs, RKE2 should refuse to start if the etcd user and group doesn't exist on the host
```
ps -ef|grep "rke2 server"
root      2247     1  0 Dec20 ?        00:04:15 /usr/local/bin/rke2 server
```

**Steps To Reproduce:**
Install rke2 on SLES 15 in CIS enabled mode

**Expected behavior:**
RKE2 should refuse to start as per the docs.

**Actual behavior:**
RKE2 starts normally. When running CIS scan 1.1.12 fails, as the etcd user is created in users group as the cis requirement is not met.
Scan report:
```
1.1.12 | Ensure that the etcd data directory ownership is set to etcd:etcd (Scored)
```
```
stat -c %U:%G /var/lib/rancher/rke2/server/db/etcd
etcd:users
```
```
cat /etc/passwd |grep tail -1
etcd:x:475:100:etcd user:/home/etcd:/sbin/nologin

cat /etc/group |grep 100
users:x:100:
```
https://docs.rke2.io/security/cis_self_assessment16/#1112

brandond:
Huh, we've only ever checked for the user, not the group. Despite both the check and our docs referencing the group being required.

rancher-max:
Validated using v1.23.1-rc1+rke2r1 following the same steps as noted in the issue

PR #2290, opened 2021-12-21
add check for etcd group

briandowns:

#### Proposed Changes ####

Adds check for etcd group when running in CIS mode.

#### Types of Changes ####

Adds a check for the etcd group and errors if it doesn't exist.

#### Verification ####

Run rke2 in cis mode with only the etcd user create. Notice the error indicating that the etcd group does not exist.

#### Linked Issues ####

#2288

### DIFF
--- a/pkg/cli/cmds/root.go
+++ b/pkg/cli/cmds/root.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/rancher/rke2/pkg/images"
 	"github.com/rancher/rke2/pkg/rke2"
@@ -245,7 +246,10 @@ func validateCISReqs(role CLIRole) error {
 	// etcd user only needs to exist on servers
 	if role == Server {
 		if _, err := user.Lookup("etcd"); err != nil {
-			ce = append(ce, fmt.Errorf("missing required %w", err))
+			ce = append(ce, errors.Wrap(err, "missing required"))
+		}
+		if _, err := user.LookupGroup("etcd"); err != nil {
+			ce = append(ce, errors.Wrap(err, "missing required"))
 		}
 	}
 


pkg/cli/cmds/root.go
```diff
@@ -245,7 +246,10 @@ func validateCISReqs(role CLIRole) error {
 	// etcd user only needs to exist on servers
 	if role == Server {
 		if _, err := user.Lookup("etcd"); err != nil {
-			ce = append(ce, fmt.Errorf("missing required %w", err))
+			ce = append(ce, errors.Wrap(err, "missing required"))
```
brandond, line 249: I'll defer to @ShylajaDevadiga but I believe we also need to check that the `Gid` of the user returned by the `user.Lookup` call is the same as the `Gid` of the group from `user.LookupGroup`. If we ensure that here, the bits at https://github.com/rancher/rke2/blob/6f2632aaae172dda758813c1dc3c8928ba36682f/pkg/podexecutor/staticpod.go#L521 should fix the directory ownership for us.

brandond (changes requested):

pkg/cli/cmds/root.go
briandowns: There's a chance that the uid and gid don't match depending on how the operator has created the groups. I just did a test where the uid and gid are different and user and groups are assigned correctly in the pod manifest as well as the applied ownership of the db directory.

brandond: Ah, OK. The policy just says we need both an etcd user and group, and that the directory must be owned by them... but it doesn't have any requirement around the user's primary gid or membership in the etcd group. Fair enough.

brandond (approved):

pkg/cli/cmds/root.go
ShylajaDevadiga: uid and gid need not be the same while creating the user, we only need etcd user and etcd group to exist
etcd:x:990:987:etcd user:/home/etcd:/sbin/nologin

galal-hussein (approved):

dereknola (approved):

Issue #2335, opened 2022-01-09
Some options of default config.yml don't work as expectation


MrMYHuang:
**Environmental Info:**
RKE2 Version: 
rke2 version v1.22.5+rke2r1 (ce3e572376cbb1d8157f46e2ae29d7d7834067f1)
go version go1.16.10b7

Node(s) CPU architecture, OS, and Version: 
Linux myhsp7 5.10.74.3-microsoft-standard-WSL2+ #2 SMP Sat Jan 8 01:17:54 CST 2022 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
1 server

**Describe the bug:**
Some options, `node-ip, node-external-ip`, in the default config.yml, `/etc/rancher/rke2/config.yml`, doesn't work as expectation if start rke2 by `sudo systemctl start rke2-server`.  However, if start rke2 by `sudo rke2 server --node-ip www.xxx.yyy.zzz --node-external-ip www.xxx.yyy.zzz`, it works correctly. We can check IPs in etcd config, `sudo cat /var/lib/rancher/rke2/server/db/etcd/config`, to see the difference of two methods to start rke2.

**Steps To Reproduce:**
I created a vlan with a static IP 172.27.137.122 and followed the rke2 quick start with the default config.yml: https://docs.rke2.io/install/quickstart/
```
sudo ip link add link eth0 name eth0.1 type vlan id 1
sudo ip addr add 172.27.137.112/20 dev eth0.1
sudo ip link set dev eth0.1 up

curl -sfL https://get.rke2.io | sudo sh -

echo "cni: cilium
node-ip: 172.27.137.112
node-external-ip: 172.27.137.112" > config.yml
sudo mkdir -p /etc/rancher/rke2
sudo cp config.yml /etc/rancher/rke2

sudo systemctl start rke2-server.service
```
After etcd starts, run `sudo cat /var/lib/rancher/rke2/server/db/etcd/config` to check IPs and you will find they are not IP 172.27.137.112!

**Expected behavior:**

IPs in etcd config should be 172.27.137.112 of eth0.1, not the IP (e.g. 172.27.140.163) of eth0.

**Actual behavior:**

IPs in etcd config is the IP of eth0. `node-ip, node-external-ip` in the rke2 default config.yml don't work as expected. However, `cni: cilium` in it works (can be confirmed by `kubectl get pod -A`).

**Additional context / logs:**

[...]

PR #2336, opened 2022-01-10
Notice the default config file extension

MrMYHuang:
#### Proposed Changes ####
Notice user that correct default config file extension is yaml, not yml. Otherwise, it doesn't work.

#### Types of Changes ####
documentation

#### Verification ####
See https://github.com/rancher/rke2/issues/2335

#### Linked Issues ####
See https://github.com/rancher/rke2/issues/2335

#### Further Comments ####
None

### DIFF
--- a/docs/install/install_options/install_options.md
+++ b/docs/install/install_options/install_options.md
@@ -91,7 +91,7 @@ For details on configuring the RKE2 Windows agent, refer to the [Windows agent c
 
 ### Configuration File
 
-By default, RKE2 will launch with the values present in the YAML file located at `/etc/rancher/rke2/config.yaml`.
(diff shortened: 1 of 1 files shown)


docs/install/install_options/install_options.md
```diff
@@ -91,7 +91,7 @@ For details on configuring the RKE2 Windows agent, refer to the [Windows agent c
 
 ### Configuration File
 
-By default, RKE2 will launch with the values present in the YAML file located at `/etc/rancher/rke2/config.yaml`.
+By default, RKE2 will launch with the values present in the YAML (not .yml) file located at `/etc/rancher/rke2/config.yaml`.
```
brandond, line 94: The required filename including extension is literally four words later, does this really need to be called out?

briandowns: Agreed. I don't think the explicit negation is necessary.

Issue #2345, opened 2022-01-12
Windows Installer script output is not helpful if run as regular user


dereknola:
**Environmental Info:**
RKE2 Version: 

Master
Node(s) CPU architecture, OS, and Version: 

Windows

**Describe the bug:**

Installing RKE2 on Windows requires running `install.ps1` as admin. If not run as admin the output is confusing and the what is wrong is not clear. 

**Steps To Reproduce:**

Run the install.ps1 thru regular powershell

**Expected behavior:**

A clear message that powershell needs to be run as admin

**Actual behavior:**

![image](https://user-images.githubusercontent.com/11727736/149224177-a14b5d40-0b82-409e-8f80-baa2fe27158b.png)

**Additional context / logs:**

dereknola:
Closing this a the original issue was incorrect. This bug was caused by the user (me) and does not affect non-admin powershell.

PR #2346, opened 2022-01-12
Add Adminstrator check to windows install.ps1

dereknola:

#### Types of Changes ####
Added builtin powershell check for admin privileges with clear error message.

#### Verification ####
Run `install.ps1` on windows in an unprivileged powershell session

```
PS C:\Users\vagrant\Documents> .\install.ps1
.\install.ps1 : The script 'install.ps1' cannot be run because it contains a "#requires" statement for running as Administrator. The current Windows PowerShell session is not running as Administrator. Start
Windows PowerShell by  using the Run as Administrator option, and then try running the script again.
At line:1 char:1
+ .\install.ps1
+ ~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (install.ps1:String) [], ScriptRequiresException
    + FullyQualifiedErrorId : ScriptRequiresElevation
```
#### Linked Issues ####
https://github.com/rancher/rke2/issues/2345

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -76,6 +76,7 @@ param (
     $ChannelUrl = "https://update.rke2.io/v1-release/channels"
 )
 
+# --- bail if we are not administrator ---
 #Requires -RunAsAdministrator
 
 Set-StrictMode -Version Latest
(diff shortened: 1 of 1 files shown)


brandond (approved):

briandowns (approved):

install.ps1
```diff
@@ -76,6 +76,8 @@ param (
     $ChannelUrl = "https://update.rke2.io/v1-release/channels"
 )
 
+#Requires -RunAsAdministrator
```
rosskirkpat, line 79: We already have an admin check [here](https://github.com/rancher/rke2/blob/master/install.ps1#L125-L130)

dereknola: Yeah this doesn't work... The script wont even run at all when no elevated, see the issue this PR is attached to. So the check never happens because the script never gets that far. This requires script is always run regardless of where it is in the file. See https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_requires?view=powershell-7.2#rules-for-use

PR #2357, opened 2022-01-14
Revert "Add Adminstrator check to windows install.ps1"

dereknola:
Strong push-back from the windows rke2 team has necessitated reverting rancher/rke2#2346.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -76,9 +76,6 @@ param (
     $ChannelUrl = "https://update.rke2.io/v1-release/channels"
 )
 
-# --- bail if we are not administrator ---
-#Requires -RunAsAdministrator
-
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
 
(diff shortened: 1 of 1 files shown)


luthermonson:
asking derek to revert this commit, the #requires feature in powershell is a comment, it's not guaranteed to actually check anything if ran via some powershell scripting IDE etc. the standard way to guarantee the admin check is exactly how we have it in the script and will even function in powershell core.

luthermonson (approved):

brandond:
Can we do both? Put the code check back in for IDEs, but leave the `#requires` in for folks running it directly from the shell? Does having it in hurt anything?

rosskirkpat (approved):

luthermonson:
if you want to leave a comment in a script i wont stop you... but it's pointless with the real checks already happening. the original issue for this is already determined to be an error in the vagrant scripts. i dont see why this needs to be discussed further

brandond:
Just trying to learn something, sorry. I didn't block merging with my review :shrug:

brandond:
For the record, https://github.com/rancher/rke2/issues/2345 was caused by copying install.sh into the Windows vagrant host as install.ps1; there was not anything wrong with install.ps1 itself to necessitate the change, regardless of whether or not the `#requires` check is useful.

Issue #877, opened 2021-04-13
Add support for SLE Micro


davidnuzik:
QA test, see if can identify any issues. Does SELinux work when enforcing?

davidnuzik:
Nice to have in 1.21.1 for RKE2

cjellick:
@fapatel1 @davidnuzik can we confirm with @cloudnautique whether we can put this onto our backlog until it is absolutely needed?

davidnuzik:
Per @cloudnautique - moving this to the backlog for now until/if it comes up later. Additionally, later when we test the version may be 5.1 (not 5.0) - details can be determined later.

cwayne18:
While we have support for the tarball installation method, we should port the changes from the k3s installation script here to enable transactional updates.

katran001:
ADR was accepted and merged, but still in-progress. Moving to next milestone

dereknola:
We support SLE Micro, closing as OOD.

PR #2366, opened 2022-01-18
Modify install script to enable SLE Micro rpm installation

galal-hussein:
#### Proposed Changes ####

- Enable RPM installation via zypper and transactional-update in SLE Micro distro, which will in turn enable support for Selinux support as well since the rpm installation installs the following packages
  - rke2-selinux
  - rke2-common
  - rke2-server/rke2-agent  
 

#### Verification ####

- run install script in sle micro
- reboot

The install script should install the rpm packages via transactional-updates and will require rebooting of the machine to take effect

#### Linked Issues ####

- https://github.com/rancher/rke2/issues/877

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -109,9 +109,11 @@ setup_env() {
         INSTALL_RKE2_TYPE="${INSTALL_RKE2_EXEC:-server}"
     fi
 
-    # --- use yum install method if available by default
-    if [ -z "${INSTALL_RKE2_ARTIFACT_PATH}" ] && [ -z "${INSTALL_RKE2_COMMIT}" ] && [ -z "${INSTALL_RKE2_METHOD}" ] && command -v yum >/dev/null 2>&1; then
-        INSTALL_RKE2_METHOD="yum"
+    # --- use rpm install method if available by default
+    if [ -z "${INSTALL_RKE2_ARTIFACT_PATH}" ] && [ -z "${INSTALL_RKE2_COMMIT}" ] && [ -z "${INSTALL_RKE2_METHOD}" ] ;then
+	if command -v yum >/dev/null 2>&1 || command -v zypper >/dev/null 2>&1; then
+		INSTALL_RKE2_METHOD="rpm"
+	fi
     fi
 
     # --- install tarball to /usr/local by default, except if /usr/local is on a separate partition or is read-only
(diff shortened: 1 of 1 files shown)


dweomer (approved):

install.sh
```diff
@@ -404,18 +406,32 @@ install_airgap_tarball() {
...
-                maj_ver="7"
-                ;;
-        esac
+    . /etc/os-release
+    if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ] || [ "${ID_LIKE%%[ ]*}" = "suse"  ]; then
+        repodir=/etc/yum.repos.d
+        if [ -d /etc/zypp/repos.d ]; then
+            repodir=/etc/zypp/repos.d
+        fi
+        if [ "${ID_LIKE%%[ ]*}" = "suse" ]; then
```
brandond, line 415: The `"${ID_LIKE%%[ ]*}`  here looks like a bash-ism, does it work on busybox /bin/sh?

galal-hussein: I am not sure, but I guess not, but do we support installation on busybox anyway?

briandowns: I'd vote to make everything as POSIX compatible as possible.

brandond: We need to support basic Bourne shell syntax since we run the installer with /bin/sh; this means no bash-isms.

galal-hussein: makes sense, I will check for compatibility but FWIW we use the same syntax in k3s install script https://github.com/k3s-io/k3s/blob/master/install.sh#L463

brandond: I was mostly concerned about the bracket expansion; if it works in K3s it should work here too!
```console
/ # ID_LIKE="suse foo bar"; echo ${ID_LIKE%%[ ]*}
suse
```

brandond (approved):

install.sh
galal-hussein: yeah I tested it and its  working as expected in busybox /bin/sh

briandowns (approved):

galal-hussein:
BTW Smoke test for opensuse microos is failing because its using the stable channel which doesnt have the package right now, but should work with the testing channel, provided that the test will reboot the machine since its using transactional-update

Issue #2370, opened 2022-01-19
Update canal to latest flannel and calico images


manuelbuil:
**Is your feature request related to a problem? Please describe.**

Calico and flannel have new versions. We should move canal to them

**Describe the solution you'd like**

Use latest versions of canal and flannel

**Describe alternatives you've considered**

None

**Additional context**

dgiebert:
Btw this is the needed HelmChartConfig to get Wireguard to work:

```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rke2-canal
  namespace: kube-system
spec:
  valuesContent: |-
    flannel:
      image:
        repository: rancher/hardened-flannel
        tag: v0.16.1-build20220119
      backend: "wireguard"
```

Should I write a section and add it here: https://github.com/rancher/rke2/blob/master/docs/install/network_options.md

manuelbuil:
> Btw this is the needed HelmChartConfig to get Wireguard to work:
> 
> ```yaml
> apiVersion: helm.cattle.io/v1
> kind: HelmChartConfig
> metadata:
>   name: rke2-canal
>   namespace: kube-system
> spec:
>   valuesContent: |-
>     flannel:
>       image:
>         repository: rancher/hardened-flannel
>         tag: v0.16.1-build20220119
>       backend: "wireguard"
> ```
> 
> Should I write a section and add it here: https://github.com/rancher/rke2/blob/master/docs/install/network_options.md

That'd be nice! Under the `"Canal CNI plugin"` section. Please specify that this is a feature for v1.23 of rke2

bguzman-3pillar:
RKE2 commit: d038f7d16353a7de031e1d64bec2a090f986cbcb
Once RKE2 is running, deploy this:
```
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rke2-canal
  namespace: kube-system
spec:
  valuesContent: |-
    flannel:
      image:
        repository: rancher/hardened-flannel
        tag: v0.16.1-build20220119
      backend: "wireguard"
```
Get the output with the latest flannel and calico images:

`$ kubectl get all -A -o wide
`
```
.
.
.
NAMESPACE     NAME                                           DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE   CONTAINERS                      IMAGES                                                                                         SELECTOR
[...]

PR #2376, opened 2022-01-20
Add information about wireguard

dgiebert:
Added information about new option regarding wireguard option with Flannel upgrade in v1.23 (See #2370)

### DIFF
--- a/docs/install/network_options.md
+++ b/docs/install/network_options.md
@@ -29,6 +29,20 @@ The next tabs inform how to deploy each CNI plugin and override the default opti
           iface: "eth1"
     ```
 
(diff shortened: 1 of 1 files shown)


brandond (approved):

docs/install/network_options.md
```diff
@@ -29,6 +29,20 @@ The next tabs inform how to deploy each CNI plugin and override the default opti
           iface: "eth1"
     ```
 
+    Starting with RKE2 v1.23 it is possible to use flannels [wireguard backend](https://github.com/flannel-io/flannel/blob/master/Documentation/backends.md#wireguard) for in-kernel WireGuard encapsulatation and encryption ([Users of kernels < 5.6 need to install a module](https://www.wireguard.com/install/)). This can be achieved using the following config:
```
manuelbuil, line 32: encapsulation not encapsulatation

dgiebert: Good catch!

manuelbuil (approved):

PR #2424, opened 2022-02-02
E2E Validate Cluster test

dereknola:
#### Proposed Changes ####
New E2E framework similar to the one in K3s.
New Basic Cluster Validation test.
Bump Ginko to v2 (not backwards compatible with v1)
Expanded documentation

#### Types of Changes ####
New Tests

#### Verification ####
`go test -timeout=15m -v ./tests/e2e/... -run E2E

#### Linked Issues ####
N/A Internal testing change


tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -0,0 +1,191 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(func() (string, error) {
+			cmd = "kubectl exec volume-test cat /data/test --kubeconfig=" + kubeConfigFile
+			return e2e.RunCommand(cmd)
+		}, "180s", "2s").Should(ContainSubstring("local-path-test"))
+	})
+})
+
+var failed = false
```
briandowns, line 179:
```suggestion
var failed bool
```

tests/e2e/testutils.go
```diff
@@ -0,0 +1,208 @@
+package e2e
```
briandowns, line 1: Is this package / file necessary since everything is exported, there's no API really.

dereknola: Future tests need to be able to pull it in. I'm not quite sure what your proposing. Yes everything is exported, that's the point.

tests/e2e/validatecluster/validatecluster_test.go
dereknola: Fixed

tests/e2e/testutils.go
briandowns: Generally speaking, packages in Go are meant to define API boundaries. When everything is exported, it negates the need for import as the code can live next to where it's used. If it's going to be imported into other tests, then that makes sense, assuming the other tests will be in a different package from the existing tests that use this code. If they don't, this can live along side the test code that uses it.

dereknola: Other tests will be in other packages/folders. It will be how the e2e tests in K3s are https://github.com/k3s-io/k3s/tree/master/tests/e2e. This is a limitation of Ginkgo assuming only one type of test can exist within a package. Vagrant also suffers from this limitation with "one" Vagrantfile per folder.

briandowns: Gotcha. Makes sense then. I just want to avoid the creation of packages for file management purposes rather than creating APIs which they're meant for.

briandowns (approved):

brandond (approved):

Issue #510, opened 2020-10-29
Update docs to mention the need to run install as root


davidnuzik:
RKE2 requires you run the install as root. Compared to K3s install.sh there is no "auto-escalation" with sudo.
Due to this, we should tweak the RKE2 docs to make it clear that installation commands need to be ran as root.

Further, what about running kubectl? Anything else?
We should ensure the docs are friendly and call out anywhere a sudo command (or otherwise running as root) is required.

jossemargt-3pillar:
Addressed on #2434

PR #2434, opened 2022-02-04
Clarify the root user access req. for installation

jossemargt-3pillar:
Problem: RKE2 install script does not handle "auto-scalation" with sudo,
like k3s does.

Solution: Call out the requirement for either sudo or root user access.

#### Proposed Changes ####

Call out the requirement for either sudo or root user access in the documentation.

#### Types of Changes ####

Documentation

#### Verification ####

Proofread or manual provisioning following the proposed change in the documentation.

#### Linked Issues ####

#510

#### Further Comments ####

None

### DIFF
--- a/docs/install/airgap.md
+++ b/docs/install/airgap.md
@@ -2,29 +2,30 @@
 
 **Important:** If your node has NetworkManager installed and enabled, [ensure that it is configured to ignore CNI-managed interfaces.](https://docs.rke2.io/known_issues/#networkmanager)
 
-RKE2 can be installed in an air-gapped environment with two different methods.
-You can either deploy via the `rke2-airgap-images` tarball release artifact, or by using a private registry.
+RKE2 can be installed in an air-gapped environment with two different methods. You can either deploy via the `rke2-airgap-images` tarball release artifact, or by using a private registry.
 
 All files mentioned in the steps can be obtained from the assets of the desired released rke2 version [here](https://github.com/rancher/rke2/releases).
 
 If running on an air-gapped node with SELinux enabled, you must manually install the necessary SELinux policy RPM before performing these steps. See our [RPM Documentation](https://docs.rke2.io/install/methods/#rpm) to determine what you need.
- 
+
 If running on an air-gapped node running SELinux, CentOS, or RHEL 8, with SELinux enabled, the following are required dependencies when doing an [RPM install](https://docs.rke2.io/install/methods/#rpm):
-           
+
     Installing dependencies:
-    container-selinux           
-    iptables                    
-    libnetfilter_conntrack      
-    libnfnetlink                
-    libnftnl                    
+    container-selinux
+    iptables
+    libnetfilter_conntrack
+    libnfnetlink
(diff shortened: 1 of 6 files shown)


docs/install/ha.md
```diff
@@ -70,16 +71,19 @@ tls-san:
...
 
 ```
+
 As mentioned previously, you must have an odd number of server nodes in total.
 
 ### 4. Confirm cluster is functional
 Once you've launched the `rke2 server` process on all server nodes, ensure that the cluster has come up properly with
+
 ```bash
+# FOR PR: running kubectl as root only because of  /etc/rancher/rke2/rke2.yaml read access sounds like overkill, should it pointed out?
```
jossemargt-3pillar, line 81: This might be outside of the initial scope. But, should we tell the user that they don't need root access for this step if:

1. kubectl is already installed
2. kubectl context has been configured for this new cluster

docs/install/quickstart.md
```diff
@@ -13,28 +13,36 @@ If NetworkManager is installed and enabled on your hosts, [ensure that it is con
 
 - For RKE2 versions 1.21 and higher, if the host kernel supports [AppArmor](https://apparmor.net/), the AppArmor tools (usually available via the `apparmor-parser` package) must also be present prior to installing RKE2.
 
+- The RKE2 installation process must be run as the root user or through `sudo`.
```
jossemargt-3pillar, line 16: I initially thought to add `# Run as root user or through sudo` on every single code block but it gets tiresome quickly. Instead, I prefered to add this requirement here and on L61, then added `sudo` install in the comments when seems needed

docs/install/airgap.md
```diff
@@ -8,23 +8,23 @@ You can either deploy via the `rke2-airgap-images` tarball release artifact, or
 All files mentioned in the steps can be obtained from the assets of the desired released rke2 version [here](https://github.com/rancher/rke2/releases).
 
 If running on an air-gapped node with SELinux enabled, you must manually install the necessary SELinux policy RPM before performing these steps. See our [RPM Documentation](https://docs.rke2.io/install/methods/#rpm) to determine what you need.
- 
+
 If running on an air-gapped node running SELinux, CentOS, or RHEL 8, with SELinux enabled, the following are required dependencies when doing an [RPM install](https://docs.rke2.io/install/methods/#rpm):
```
jossemargt-3pillar, line 12: All the steps listed on this fie requires root user access, but since it is an advance topic I didn't thought we require to point it out. wdyt?

briandowns: Please update. We should try to keep things as consistent as possible.

docs/install/ha.md
briandowns: A reference to this might be helpful. https://docs.rke2.io/cluster_access/#cluster-access

Also, a config file can be placed on disk with the content below that won't require `root`.

```yaml
write-kubeconfig-mode: 644
```

briandowns (approved):

nikolaishields (approved):

PR #2522, opened 2022-03-01
Fix s390x build

thomasferrandiz:
This PR fixes two issues with the s390x build:
- we need to manually install the right version of yq since the default version it too old.
- we need to upgrade calico since the previous build was broken on s390x.

Related calico PR: https://github.com/rancher/image-build-calico/pull/21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,16 @@ RUN set -x \
     bsd-compat-headers \
     py-pip \
     pigz \
-    tar \
-    yq
+    tar
+    
 RUN if [ "${ARCH}" != "s390x" ]; then \
-    	apk --no-cache add mingw-w64-gcc; \
+    	apk --no-cache add mingw-w64-gcc yq; \
+    fi
+
+RUN if [ "${ARCH}" = "s390x" ]; then \
+        curl -LJO https://github.com/mikefarah/yq/releases/download/v4.6.1/yq_linux_s390x && \
+        mv yq_linux_s390x /usr/bin/yq && \
+        chmod +x /usr/bin/yq; \
     fi
 
 # Dapper/Drone/CI environment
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -15,10 +15,16 @@ RUN set -x \
...
+    tar
+    
 RUN if [ "${ARCH}" != "s390x" ]; then \
-    	apk --no-cache add mingw-w64-gcc; \
+    	apk --no-cache add mingw-w64-gcc yq; \
+    fi
+
+RUN if [ "${ARCH}" = "s390x" ]; then \
+        curl -LJO https://github.com/mikefarah/yq/releases/download/v4.6.1/yq_linux_s390x && \
+        mv yq_linux_s390x /usr/bin/yq && \
```
brandond, line 26: Why is the s390x build of yq the only one that's too old?

brandond (approved):

Dockerfile
thomasferrandiz: I don't know the reason but the versions of `yq` installed by `apk` are as follows:
On intel build: 
```
yq --version
yq version 4.6.3
```

On s390x build:
```
yq --version
yq version 3.4.1
```

brandond: Hmm, can you confirm what version of alpine is being used? According to the alpine package list, a much newer version that that should be available:
https://pkgs.alpinelinux.org/packages?name=yq&branch=v3.15&arch=s390x

thomasferrandiz: According to: https://github.com/rancher/image-build-base/blob/d28370a69aacdca3054c6a9b0faff47ba9f04e19/Dockerfile.s390x#L2

The s390x build uses alpine3.13 which uses an old version of yq:
https://pkgs.alpinelinux.org/packages?name=yq&branch=v3.13&arch=s390x

thomasferrandiz: After checking more, bumping the base image fixes the issue without needing a special version of yq for s390x.
See https://github.com/rancher/rke2/pull/2522/commits/d9090eb83a4d84dd73ac7b02af70e054936af325

brandond (approved):

briandowns (approved):

Issue #2543, opened 2022-03-04
Windows Install via `-Commit` fails


dereknola:
**Environmental Info:**
RKE2 Version: 

v1.22.7+rke2r1

Node(s) CPU architecture, OS, and Version: 

Windows Server

Cluster Configuration:

1 server, 1 linux agent, 1 windows agent
**Describe the bug:**

Windows agent fails to startup due to missing tar images.
```
Failed to load runtime 
image index.docker.io/rancher/rke2-runtime:v1.22.7-dev-90d94ddb from tarball: no 
local image available for 
index.docker.io/rancher/rke2-runtime:v1.22.7-dev-90d94ddb: directory 
C:\var\lib\rancher\rke2\agent\images does not exist: image not found<
```

**Steps To Reproduce:**

`./install.ps1 -Commit 90d94ddbd8601c4b1bc15460811239e1c99ba1b3`

**Expected behavior:**

Install.ps1 should place the image.tar under /var/lib/rancher/rke2/agent/images

**Actual behavior:**

No /var/lib/rancher/rke2/agent/images directory is ever created

**Additional context / logs:**

rosskirkpat:
```powershell
./install.ps1 -Commit 058823838da854109d8159911bbcd5f2d5030ad3
[INFO] Windows feature: 'Containers' is installed. Installation will proceed.
Using stable channel of rke2 for installation
[INFO] using commit 058823838da854109d8159911bbcd5f2d5030ad3 as release
[INFO] Version: commit 058823838da854109d8159911bbcd5f2d5030ad3
Storage URL: https://storage.googleapis.com/rke2-ci-builds
Github URL: https://github.com/rancher/rke2
Binary Checksums: C:\Users\Administrator\AppData\Local\Temp\rke2-install\rke2.checksums
Image Checksums: C:\Users\Administrator\AppData\Local\Temp\rke2-install\rke2-images.checksums
downloading image checksum for commit: 058823838da854109d8159911bbcd5f2d5030ad3 at https://storage.googleapis.com/rke2-ci-builds/rke2-images.windows-amd64-058823838da854109d8159911bbcd5f2d5030ad3.tar.zst.sha256sum
[INFO] downloading airgap tarball with commit 058823838da854109d8159911bbcd5f2d5030ad3 from https://storage.googleapis.com/rke2-ci-builds/rke2-images.windows-amd64-058823838da854109d8159911bbcd5f2d5030ad3.tar.zst
[...]

Issue #2125, opened 2021-11-15
Cannot install rke2 on windows via commitid because install script does not add tar file


rancher-max:
**Environmental Info:**
RKE2 Version: 

Commitid install, any branch. Validated using master branch commit `12e640bd`

Node(s) CPU architecture, OS, and Version: 

Windows 2019

Cluster Configuration:

1 server running on ubuntu, 1 agent on windows datacenter 2019

**Describe the bug:**

Install via commitid fails on the windows agent because it doesn't install the necessary tar file(s) into `/var/lib/rancher/rke2/agent/images` so tries to pull the rke2-runtime image from dockerhub, even though it doesn't exist.

**Steps To Reproduce:**

- Install RKE2 on windows via commitid
- Start the rke2 service

**Expected behavior:**

It should run and join the cluster successfully 

**Actual behavior:**

It fails to pull rke2-runtime image:
```
failed to get runtime image index.docker.io/rancher/rke2-runtime:v1.22.3-dev-12e640bd: GET https://index.docker.io/v2/rancher/rke2-runtime/manifests/v1.22.3-dev-12e640bd: MANIFEST_UNKNOWN: manifest unknown; unknown tag=v1.22.3-dev-12e640bd
```

**Additional context / logs:**

Version:
```
PS C:\Users\Administrator> rke2 --version
rke2.exe version v1.22.3-dev+12e640bd (12e640bd0106812bc552490c3ee1c87e65cde6fd)
go version go1.16.7b7
```

Example what I see on a server node:
```
Nov 15 19:33:18 ip-172-31-17-75 rke2[1664]: time="2021-11-15T19:33:18Z" level=info msg="Checking local image archives in /var/lib/rancher/rke2/agent/images for index.docker.io/rancher/rke2-runtime:v1.22.3-dev-12e640bd"
...
Nov 15 19:33:34 ip-172-31-17-75 rke2[1664]: time="2021-11-15T19:33:34Z" level=info msg="Imported docker.io/rancher/rke2-runtime:v1.22.3-dev-12e640bd"

...
$ sudo ls /var/lib/rancher/rke2/agent/images
rke2-images.linux-amd64.tar.zst
```

briandowns:
Resolved via #1890

PR #2544, opened 2022-03-04
Fix windows commit installs

dereknola:

#### Proposed Changes ####
Add missing `-Not` to two if checks (Parameter was always passed and we WANT to check if it is empty)
Checks if `/var/lib/rancher/rke2/agent/images` directory exists before trying to make a new one
Removes duplicate directory creation line

#### Types of Changes ####
Bugfix

#### Verification ####
Setup RKE2 linux server with a commit 
`curl -sfL https://get.rke2.io | INSTALL_RKE2_COMMIT=<COMMIT> sh -`

Install RKE2 on windows with commit 
My yaml config
```
kube-proxy-arg: "feature-gates=IPv6DualStack=false"
node-external-ip: #{win_agent_ip}
node-ip: #{win_agent_ip}

server: https://#{server_ip}:9345
token: vagrant-rke2
```
`./install.ps1 -Commit <COMMIT>` now installs correctly, and RKE2 agent will connect to linux server
Additionally, running the `install.ps1` again will no longer break complaining about an existing /var/lib/rancher/rke2/agent/images directory

#### Linked Issues ####
https://github.com/rancher/rke2/issues/2543
https://github.com/rancher/rke2/issues/2125


install.ps1
```diff
@@ -635,22 +635,24 @@ function Install-AirgapTarball() {
...
+    if (-Not $ExpectedImageAirGapChecksum) {
         return
     }
 
     $archInfo = Get-ArchitectureInfo
     $suffix = $archInfo.Suffix    
     $arch = $archInfo.Arch
 
+    if (Test-Path $InstallAgentImageDir) {
+        Remove-Item -Path $InstallAgentImageDir -Force -Recurse
```
brandond, line 647: Why are we nuking the agent image dir? What if the user had placed some of their own image tarballs there prior to install?

phillipsj (approved):

install.ps1
phillipsj: Yeah, this is new behavior as compared to the Linux install script.

brandond (approved):

phillipsj (approved):
LGTM!

install.ps1
```diff
@@ -600,7 +600,7 @@ function Test-AirgapTarballChecksum() {
         return
     }
 
-    if ($ExpectedImageAirGapChecksum) {
+    if (-Not $ExpectedImageAirGapChecksum) {
```
rosskirkpat, line 603: The install script in master is working based on my testing https://github.com/rancher/rke2/issues/2543#issuecomment-1062217804, have you verified that you are running powershell as an administrator?

rosskirkpat (changes requested):

install.ps1
rosskirkpat: I had an incorrect understanding of the issue, removing my block and approving as I have validated this fix myself as well.

rosskirkpat (approved):

rosskirkpat:
Thank you for your patience while I figured this one out @dereknola! It's greatly appreciated

rosskirkpat (changes requested):
blocking this as there are lower level issues that cause these changes to not work due to using zstd compressed tarballs for the dev-runtime-image outputs.

I am taking the design issue offline with @dereknola

rosskirkpat (approved):
LGTM

PR #2588, opened 2022-03-11
remove line breaks

briandowns:


Oats87 (approved):

brandond (approved):
LGTM but I'm confused as to what it's fixing.
Edit: Oh I see, the grep in the version check script expects them to be on one line. Bummer.

PR #2595, opened 2022-03-11
WIP: Build rke2 with tekton

nikolaishields:
#### Proposed Changes ####
Build RKE2 with tekton via the manifests present in /tekton.


nikolaishields:
This PR isn't necessary at the moment, as we are still in the discovery phase of Tekton.

Issue #2161, opened 2021-11-23
Unable to start RKE2 Windows Agent due to npipe transportation error


ckehoe:
**Environmental Info:**
RKE2 Version:

rke2.exe version v1.21.6+rke2r1 (b915fc9)
go version go1.16.6b7

I have also tried it with rke2 v1.22.3+rle2r1

Node(s) CPU architecture, OS, and Version:

64-bit, Microsoft Windows Server 2019 Standard, 10.0.17763

Cluster Configuration:

One centos 7.9 server, one centos 7.9 agent, and one Windows 2019 Server
Used calico as CNI and was able to join the centos 7.9 agent
Running on VSphere

**Describe the bug:**

RKE2 is unable to communicate with containerd after startup due to an npipe transportation error:

`failed to dial \"npipe:////./pipe/containerd-containerd\": connection error: desc = \"transport: error while dialing: open npipe:////./pipe/containerd-containerd: The filename, directory name, 
or volume label syntax is incorrect.\"`

This looks the the exact error described [here](https://github.com/rancher/windows/issues/72) but the solution for that was installing the latest Windows Patches. It also wasn't supposed to apply to the version of Windows server that I'm using. I have the latest November 2021 Cumulative update installed so I don't believe that is my issue.

ckehoe:
I believe the issue is that `npipe:////./pipe/containerd-containerd` is being passed as the address to ctr.exe which invalid. I copied containerd.exe, crictl.exe, and ctr.exe from `c:\var\lib\rancher\rke2\data\{version}\bin` to my `c:\` drive and ran the following:

```powershell
# Start Containerd
PS> C:\containerd.exe -c C:\var\lib\rancher\rke2\agent\etc\containerd\config.toml
# Connect crictl to containerd endpoint using npipe
PS> $env:CONTAINER_RUNTIME_ENDPOINT = "npipe:////./pipe/containerd-containerd"
PS> C:\crictl.exe images
IMAGE               TAG                 IMAGE ID            SIZE
```
Everything looks good and crictl is able to connect to containerd using the npipe:// address. Then I tried the same address for ctr.exe:

```powershell
PS > $env:CONTAINERD_ADDRESS = "npipe:////./pipe/containerd-containerd"
PS > C:\ctr.exe images list
[...]

Issue #1954, opened 2021-10-13
Uninstall and reinstall on windows 2022 hits router panic


rancher-max:
**Environmental Info:**
RKE2 Version: 

v1.22.2-rc1+rke2r1

Node(s) CPU architecture, OS, and Version: 

```
PS C:\Users\Administrator> Get-ComputerInfo

WindowsBuildLabEx                                       : 20348.1.amd64fre.fe_release.210507-1500
WindowsCurrentVersion                                   : 6.3
WindowsEditionId                                        : ServerDatacenter
WindowsInstallationType                                 : Server
WindowsInstallDateFromRegistry                          : 10/13/2021 1:07:56 AM
WindowsProductId                                        : 00454-60000-00001-AA008
WindowsProductName                                      : Windows Server 2022 Datacenter
WindowsRegisteredOrganization                           : Amazon.com
WindowsRegisteredOwner                                  : EC2
WindowsSystemRoot                                       : C:\Windows
WindowsVersion                                          : 2009
```
Using AWS: `Windows_Server-2022-English-Full-ContainersLatest-2021.09.15 - ami-0d55df877a8eada89`

Cluster Configuration:

server running calico, windows agent running calico

**Describe the bug:**

Initial install worked fine. After running the uninstall script, deleting the node from the cluster, and trying to reinstall, I see this error:
```
time="2021-10-13T23:32:34Z" level=info msg="Generating HNS networks, please wait"
panic: router only implemented in linux

goroutine 1 [running]:
github.com/google/gopacket/routing.New(...)
        /go/pkg/mod/github.com/google/gopacket@v1.1.19/routing/other.go:14
github.com/rancher/rke2/pkg/windows.setMetaDataServerRoute(0xc00177a5a0, 0xd, 0x0, 0x0)
        /source/pkg/windows/calico.go:478 +0xcb
github.com/rancher/rke2/pkg/windows.generateCalicoNetworks(0x7ff7d47b3063, 0x5, 0xc0029055c0, 0x1)
        /source/pkg/windows/calico.go:310 +0x19f
github.com/rancher/rke2/pkg/windows.(*Calico).Setup(0x7ff7d7d38d68, 0x7ff7d52551d8, 0xc0019e8ac0, 0x7ff7d47f8d53, 0x15, 0xc000101400, 0xc001ac7440, 0xc0014a4000, 0x207, 0x19)
        /source/pkg/windows/calico.go:111 +0x717
[...]

PR #2700, opened 2022-03-30
[master] k3s pin for containerd npipe fix + calico panic fix

luthermonson:
Problem
During the installation process of calico they destroy all networking on the box and put it back. Part of that process is to blow away the routing and on ec2/gce you will lose the route to the metadata server. They account for this by validating they're on one of these providers first and later in their scripts they readd the route.

When we install we actually check for the cloud platform too late as we've always already blown out the route so the router is skipped. Calico itself has some logic to add this route back via an outbound NAT rule which isn't blown away while we install calico and the second time around you run rke2 it has the route and hits code in RKE2 which panics as the routing library go as it wasn't noticed that it didn't actually implement the `New-NetRoute` things you'd expect in windows. This wasn't caught because it was working on installation and only broke in ec2/gce if you tried to reuse the nodes.

Solution
Grabbing an updated netroute library which supports windows fully to apply the metadata route properly and moving the platform check to the top of `generateCalicoNetworks` so it's checked before the route is blown away which is how calico did it upstream.

https://github.com/rancher/rke2/issues/1954
https://github.com/rancher/rke2/issues/2161

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,10 +65,10 @@ require (
 	github.com/Microsoft/hcsshim v0.9.2
 	github.com/containerd/continuity v0.2.2
 	github.com/google/go-containerregistry v0.7.0
-	github.com/google/gopacket v1.1.19
 	github.com/iamacarpet/go-win64api v0.0.0-20210311141720-fe38760bed28
 	github.com/k3s-io/helm-controller v0.12.0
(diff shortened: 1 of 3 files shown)


phillipsj (approved):
LGTM!

Issue #2599, opened 2022-03-14
masquerade NAT rules remain after uninstall


ShylajaDevadiga:
Environmental Info:
RKE2 Version: v1.23.4+rke2r2

Describe the bug:
rke2-uninstall.sh does not clean up the masquerade rules

Node(s) CPU architecture, OS, and Version:
NA, Tested on Ubuntu

Cluster Configuration:
Single node or multi node configuration

Steps To Reproduce:
- Install rke2 with default settings
- Run rke2-uninstall.sh
- Check iptables using

$ sudo iptables-save |grep -i masq
```
# Warning: iptables-legacy tables present, use iptables-legacy-save to see them
-A POSTROUTING -s 10.42.0.0/16 ! -d 224.0.0.0/4 -j MASQUERADE --random-fully
-A POSTROUTING ! -s 10.42.0.0/16 -d 10.42.0.0/16 -j MASQUERADE --random-fully
```
Expected behavior:
All rules applied during installation should be removed at uninstall

Actual behavior:
Some rules remain

Backporting

[*] Needs backporting to older releases

rancher-max:
Reopening because this does not work when doing an upgrade from v1.23.4+rke2r2 to v1.23.5-rc4+rke2r1. Checked using SLES 15 SP3.

manuelbuil:
> Reopening because this does not work when doing an upgrade from v1.23.4+rke2r2 to v1.23.5-rc4+rke2r1. Checked using SLES 15 SP3.

Yes, this is expected. Flannel is not capable of detecting the old iptables rules because they are missing the new arg "--comment" which allows us to identify those rules and delete them. I don't think there is a good workaround to avoid this :(

brandond:
This should only be an issue following upgrades, correct? If you reboot or otherwise clear out the iptables rules and let flannel put them back, the uninstall script should clean them up once they have the correct comments on them.

rancher-max:
> This should only be an issue following upgrades, correct? If you reboot or otherwise clear out the iptables rules and let flannel put them back, the uninstall script should clean them up once they have the correct comments on them.

Yep that's correct!

manuelbuil:
I agreed wtih @rancher-max to add a comment in the clean-up script informing people that they might need to remove those rules if they are coming from an upgrade

ShylajaDevadiga:
Validated masquerade NAT rules are removed using v1.23.6-rc3+rke2r1 install and upgrade
[...]

PR #2727, opened 2022-04-06
Add message in the killall script

manuelbuil:

#### Proposed Changes ####

In old flannel versions, it is not possible to remove the iptable rule created for snatting because it does not include anything flannel specific. This changes in the new versions of flannel. However, if customer comes from an old version of flannel, those old iptable rules will remain after destroying the cluster.

#### Types of Changes ####

Adding log/message

#### Verification ####

When destroying the cluster, the message should appear

#### Linked Issues ####

https://github.com/rancher/rke2/issues/2599

### DIFF
--- a/bundle/bin/rke2-killall.sh
+++ b/bundle/bin/rke2-killall.sh
@@ -84,4 +84,10 @@ if [ -d /sys/class/net/nodelocaldns ]; then
 fi
 
 rm -rf /var/lib/cni/
+
+# Delete iptables created by CNI plugins or Kubernetes (kube-proxy)
 iptables-save | grep -v KUBE- | grep -v CNI- | grep -v cali- | grep -v cali: | grep -v CILIUM_ | grep -v flannel | iptables-restore
+echo 'If this cluster was upgraded from an older release of the Canal CNI, you may need to manually remove some flannel iptables rules:'
+echo -e '\texport cluster_cidr=YOUR-CLUSTER-CIDR'
+echo -e '\tiptables -D POSTROUTING -s $cluster_cidr -j MASQUERADE --random-fully'
+echo -e '\tiptables -D POSTROUTING ! -s $cluster_cidr -d  -j MASQUERADE --random-fully'


jossemargt-3pillar (approved):

briandowns (approved):

bundle/bin/rke2-killall.sh
```diff
@@ -84,4 +84,9 @@ if [ -d /sys/class/net/nodelocaldns ]; then
 fi
 
 rm -rf /var/lib/cni/
+
+# Delete iptables created by CNI plugins or Kubernetes (kube-proxy)
 iptables-save | grep -v KUBE- | grep -v CNI- | grep -v cali- | grep -v cali: | grep -v CILIUM_ | grep -v flannel | iptables-restore
+echo
+echo if this cluster has been upgraded and was using canal, you may need to remove the flannel iptables with 'iptables -D POSTROUTING -s $cluster-cidr -j MASQUERADE --random-fully' and 'iptables -D POSTROUTING ! -s $cluster-cidr -d  -j MASQUERADE --random-fully'
```
brandond, line 91: Might it be simpler to tell users to just reboot or manually clear all the iptables rules? If we want to print commands, lets at least give them a couple lines to copy-paste:
```suggestion
echo 'If this cluster was upgraded from an older release of the Canal CNI, you may need to manually remove some flannel iptables rules:'
echo -e '\texport cluster_cidr=YOUR-CLUSTER-CIDR'
echo -e '\tiptables -D POSTROUTING -s $cluster_cidr -j MASQUERADE --random-fully'
echo -e '\tiptables -D POSTROUTING ! -s $cluster_cidr -d  -j MASQUERADE --random-fully'
```

brandond (changes requested):

bundle/bin/rke2-killall.sh
manuelbuil: Thanks for the suggestion. Yes, that would be simpler but rebooting might not be an option and if there are other tools using iptables, it might be difficult to find the exact rules. Copy-paste lines is a good idea

PR #2833, opened 2022-04-30
Bump dynamiclistener to fix apiserver outage issue

brandond:
#### Proposed Changes ####

Bump dynamiclistener to fix an issue where circular outages can prevent the supervisor from accepting new connections:
1. supervisor on apiserver-only node tries to talk to bootstrap node during startup to reconcile bootstrap data (can't do it locally, no etcd)
2. when accepting the TLS connection from the apiserver-only node, the etcd-only bootstrap node tries to talk to an apiserver to update the dynamiclistener certificate secret
3. apiserver on apiserver-only node is still running in static pod from previous startup, but it is not responsive because it can't talk to any etcd servers through load-balancer tunnel, since it doesn't come up until step 1 is done
4. apiserver-only rke2 supervisor startup fails due to TLS handshake eventually timing out

Running rke2-killall.sh on the apiserver-only node allows the start to succeed, as it kills the apiserver static pod, which in turn causes the dynamiclistener certificate update to fail (which IS handled properly) instead of just hanging.

* Waiting on merge of https://github.com/rancher/dynamiclistener/pull/57

#### Types of Changes ####

bugfix / version bump

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/2834

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ replace (
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.3
 	github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
 	github.com/opencontainers/selinux => github.com/opencontainers/selinux v1.10.1
+	github.com/rancher/dynamiclistener => github.com/rancher/dynamiclistener v0.3.2
 	github.com/rancher/remotedialer => github.com/rancher/remotedialer v0.2.0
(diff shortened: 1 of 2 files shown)


thedadams:
For reference, the created issue is [here](https://github.com/rancher/rke2/issues/2834).

briandowns (approved):

briandowns (approved):

briandowns (approved):

PR #2873, opened 2022-05-09
docs: fix invalid path

Jellyfrog:

### DIFF
--- a/docs/security/hardening_guide.md
+++ b/docs/security/hardening_guide.md
@@ -48,7 +48,7 @@ The location of the config file depends on the installation method used.
 
 If RKE2 was installed via RPM, YUM, or DNF (the default on OSes that use RPMs, such as CentOS), run the following command(s):
 ```bash
-sudo cp -f /usr/share/rke2/rke2-cis-sysctl.conf /etc/sysctl.d/60-rke2-cis.conf
+sudo cp -f /usr/local/share/rke2/rke2-cis-sysctl.conf /etc/sysctl.d/60-rke2-cis.conf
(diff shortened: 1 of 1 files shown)


docs/security/hardening_guide.md
```diff
@@ -48,7 +48,7 @@ The location of the config file depends on the installation method used.
 
 If RKE2 was installed via RPM, YUM, or DNF (the default on OSes that use RPMs, such as CentOS), run the following command(s):
 ```bash
-sudo cp -f /usr/share/rke2/rke2-cis-sysctl.conf /etc/sysctl.d/60-rke2-cis.conf
+sudo cp -f /usr/local/share/rke2/rke2-cis-sysctl.conf /etc/sysctl.d/60-rke2-cis.conf
```
brandond, line 51: What distro are you on? As far as I know, all of the RPMs all put it in the documented path.

```console
opensuse01:~ # rpm -ql rke2-common | grep sysctl
/usr/share/rke2/rke2-cis-sysctl.conf
```

Jellyfrog: oh, its via rancher on rhel8, file doesn't belong to any package according to rpm (ie, it doesnt install via rpm it seems)

brandond: Hmm, did someone perhaps install from tarball on that host, instead of using the RPMs? This section of the install documentation is specific to RPM-based installs.

Jellyfrog: > Note that installs from the Rancher Cluster Provisioning UI always use tarball, not RPM.

Thats what I realized.

PR #2880, opened 2022-05-09
Move windows agent dependencies check to CI

jossemargt-3pillar:
Currently windows specific binaries are checked upon release only, making any binary drift visible until last minute.

#### Proposed Changes ####

Move window binaries check to CI validation

#### Types of Changes ####

CI

#### Verification ####

CI working correctly

#### Linked Issues ####

#2911

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -6,6 +6,30 @@ if [ -n "${SKIP_VALIDATE}" ]; then
     exit 0
 fi
 
+fatal() {
+    echo '[ERROR] ' "$@" >&2
+    exit 1
+}
+
+function check_win_binaries() {
+    CRICTL_WINDOWS_VERSION=$(grep 'CRICTL_VERSION=' Dockerfile.windows | cut -d '=' -f 2- | grep -oE "v([0-9]+)\.([0-9]+)")
+    if [ ! "$CRICTL_WINDOWS_VERSION" = "v$VERSION_MAJOR.$VERSION_MINOR" ]; then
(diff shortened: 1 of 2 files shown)


scripts/validate
```diff
@@ -6,6 +6,26 @@ if [ -n "${SKIP_VALIDATE}" ]; then
     exit 0
 fi
 
+function check_win_binaries() {
+    # --- disabled until upstream releases a 1.23 release of cri-tools
+    #CRICTL_WINDOWS_VERSION=$(grep 'CRICTL_VERSION=' Dockerfile.windows | cut -d '=' -f 2- | grep -oE "v([0-9]+)\.([0-9]+)")
+    #if [ ! "$CRICTL_WINDOWS_VERSION" = "v$MAJOR.$MINOR" ]; then
+    #    fatal "crictl windows binary version [$CRICTL_WINDOWS_VERSION] does not match kubernetes version"
+    #fi
```
brandond, line 14: there are now 1.23 and 1.24 versions of cri-tools

brandond (approved):
LGTM other than one nit about cri-tools

scripts/validate
jossemargt-3pillar: I'd rather remove those commented out lines, since I don't know if we want to bump `cri-tools` version in `Dockerfile.windows`

brandond: I would recommend keeping them, and fixing the version in the Dockerfile. Now, between releases, is a good time to do that.

brandond (approved):

PR #3039, opened 2022-06-10
Update Canal issues with IP exhaustion

manuelbuil:

#### Proposed Changes ####

Update docs with potential canal issue

#### Types of Changes ####

Docs change


docs/known_issues.md
```diff
@@ -92,3 +92,7 @@ Wicked configures the networking settings of the host based on the sysctl config
 net.ipv4.conf.all.forwarding=1
 net.ipv6.conf.all.forwarding=1
 ```
+
+## Canal and IP exhaustion
+
+By default Canal keeps track of the used IPs by saving them in `/var/lib/cni/networks/k8s-pod-network`. Each IP belongs to a pod and will be deleted as soon as the pod is removed. However, in the unlikely event that containerd loses track of the running pods, there might be IPs which are left over. Canal will not be able to reuse those IPs anymore and you could experience IP exhaustion errors. In such cases, please remove the left over IPs from that directory and report the problem, specifying how it was triggered
```
brandond, line 98: Can we quote the specific error message that users will see when IPs are exhausted?

Couple nits on the spelling and punctuation.
```suggestion
By default Canal keeps track of pod IPs by creating a lock file for each IP in `/var/lib/cni/networks/k8s-pod-network`. Each IP belongs to a single pod and will be deleted as soon as the pod is removed. However, in the unlikely event that containerd loses track of the running pods, lock files may be leaked, and Canal will not be able to reuse those IPs anymore. If this occurs, you may experience IP exhaustion errors. To resolve this, you can manually remove unused IPs from that directory. If you need to do this, please report the problem via GitHub, making sure to specify how it was triggered.
```

brandond (changes requested):

Martin-Weiss:
Maybe we could also add the information that a workaround for RKE2 is to drain the node, stop rke2-server or rke2-agent and run rke2-killall.sh.. then start server/agent again and uncordon..

manuelbuil:
> Maybe we could also add the information that a workaround for RKE2 is to drain the node, stop rke2-server or rke2-agent and run rke2-killall.sh.. then start server/agent again and uncordon..

The reason for this is that the `rke2-killall.sh` removes the directory `/var/lib/cni/networks/k8s-pod-network`. My opinion is that draining the node, stopping rke2, etc means more hassle than removing files. Therefore, it seems like a more complicated workaround, or?

docs/known_issues.md
manuelbuil: Thanks, done!

docs/known_issues.md
```diff
@@ -92,3 +92,13 @@ Wicked configures the networking settings of the host based on the sysctl config
 net.ipv4.conf.all.forwarding=1
 net.ipv6.conf.all.forwarding=1
 ```
+
+## Canal and IP exhaustion
+
+By default Canal keeps track of pod IPs by creatig a lock file for each IP in `/var/lib/cni/networks/k8s-pod-network`. Each IP belongs to a single pod and will be deleted as soon as the pod is removed. However, in the unlikely event that containerd loses track of the running pods, lock files may be leaked and Canal will not be able to reuse those IPs anymore. If this occurs, you may experience IP exhaustion errors, for example:
```
brandond, line 98: nit: typo
```suggestion
By default Canal keeps track of pod IPs by creating a lock file for each IP in `/var/lib/cni/networks/k8s-pod-network`. Each IP belongs to a single pod and will be deleted as soon as the pod is removed. However, in the unlikely event that containerd loses track of the running pods, lock files may be leaked and Canal will not be able to reuse those IPs anymore. If this occurs, you may experience IP exhaustion errors, for example:
```

docs/known_issues.md
```diff
@@ -92,3 +92,13 @@ Wicked configures the networking settings of the host based on the sysctl config
...
+
+```bash
```
brandond, line 100: nit: console or just no syntax for log messages; bash will highlight weird stuff if you use it for log output.
```suggestion
```console
```

brandond (changes requested):

docs/known_issues.md
manuelbuil: aaaaah ok, thanks!

docs/known_issues.md
manuelbuil: good catch

dereknola (approved):

Issue #3048, opened 2022-06-13
Create release Channel for 1.24


alb-dev:
**Is your feature request related to a problem? Please describe.**
Create a release channel for 1.24 so that it can be used as a reference in other sotware products. 

**Describe the solution you'd like**
Create a release channel like the following "https://update.rke2.io/v1-release/channels/v1.23" for 1.24

**Describe alternatives you've considered**

**Additional context**
This would be greate for the updates (system upgrade controller) so that minor release can be installed automatically

rancher-max:
Validated v1.24 channel is now present:
```
$ curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_CHANNEL=v1.24 sh -
[INFO]  finding release for channel v1.24
[INFO]  using v1.24.1+rke2r2 as release
[INFO]  downloading checksums at https://github.com/rancher/rke2/releases/download/v1.24.1+rke2r2/sha256sum-amd64.txt
[INFO]  downloading tarball at https://github.com/rancher/rke2/releases/download/v1.24.1+rke2r2/rke2.linux-amd64.tar.gz
[INFO]  verifying tarball
[INFO]  unpacking tarball file to /usr/local
```

PR #3053, opened 2022-06-15
Update channel server for May patch release

brooksn:
#### Proposed Changes ####

Updates stable channel to v1.22.10+rke2r2, and adds release channel for v1.24 in channel server #3048

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,6 +1,6 @@
 channels:
 - name: stable
-  latest: v1.22.10+rke2r2
+  latest: v1.23.7+rke2r2
 - name: latest
   latestRegexp: .*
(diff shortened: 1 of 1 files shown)


briandowns (approved):

cwayne18 (approved):

channels.yaml
```diff
@@ -1,6 +1,6 @@
 channels:
 - name: stable
-  latest: v1.22.9+rke2r2
+  latest: v1.22.10+rke2r2
```
brandond, line 3: Are we still waiting for Rancher to catch up before promoting 1.23 to stable, or can we do that now since 2.6.5 is out?

brandond (changes requested):

Issue #3004, opened 2022-05-31
`rke2-server.service` fails to activate during `systemctl restart` following `rke2 secrets-encrypt reencrypt`


jakefhyde:
**Environmental Info:**
RKE2 Version: v1.23.6+rke2r2

Cluster Configuration:

1 server (rancher provisioned, but reproducible standalone)

**Describe the bug:**

In a single node cluster, `rke2-server.service` is stuck activating during `systemctl restart rke2-server.service` following `rke2 secrets-encrypt reencrypt` command.

**Steps To Reproduce:**

- Installed RKE2: Rancher provisioned, but also reproducible standalone
- Create a script that does the following:
  - Run `rke2 secrets-encrypt prepare`
  - Scrape `rke2 secrets-encrypt status` to verify that the current stage is `prepare`
  - Restart the node using `systemctl restart rke2-server.service`
  - Run `rke2 secrets-encrypt rotate`
  - Scrape `rke2 secrets-encrypt status` to verify that the current stage is `rotate`
  - Restart the node using `systemctl restart rke2-server.service`
  - Run `rke2 secrets-encrypt reencrypt`
  - Scrape `rke2 secrets-encrypt status` to verify that the current stage is `reencrypt_request`, `reencrypt_active`, or `reencrypt_finished`
  - Continue scraping status until the current stage is `reencrypt_finished`
  - Restart the node using `systemctl restart rke2-server.service`

**Expected behavior:**

`rke2-server.service` successfully restarts following `rke2 secrets-encrypt reencrypt`

**Actual behavior:**

`rke2-server.service` is stuck in `Activating` state, and the `kube-apiserver` pod will eventually restart, however it produces no logs after the restart and the `Waiting for API server to become available` message never goes away.

**Additional context / logs:**

### `journalctl -u rke2-server.service`
```
May 31 17:56:09 jhyde-do-pfipa-pool0-77085e31-lwq7z rke2[20284]: time="2022-05-31T17:56:09Z" level=info msg="Waiting for API server to become available"
May 31 17:56:09 jhyde-do-pfipa-pool0-77085e31-lwq7z rke2[20284]: time="2022-05-31T17:56:09Z" level=info msg="Waiting to retrieve kube-proxy configuration; server is not ready: https://127.0.0.1:9345/v1-rke2/readyz: 500 Internal Server Error"
[...]

PR #3054, opened 2022-06-15
Add Static Pod Startup Hook + K3s Bump + CoreDNS bump

dereknola:
#### Proposed Changes ####
1) Add a new startup hook which waits for the correct static pods to be deployed when kubeapiserver is present. This prevents issues when the static manifests are changed during a rke2 restart. Previously, the old pods would be deployed, then later, the new manifests would be generated and replace the old pods, with no waiting. 
2) Bump K3s to support waiting until all startup hooks are finished before reporting that the rke2 service is ready. https://github.com/k3s-io/k3s/pull/5649
3) Bump coreDNS 

#### Types of Changes ####
BugFix

#### Verification ####
See https://github.com/rancher/rke2/issues/3004, on single node, restarting rke2 after running secrets encryption commands no longer causes rke2 to hang.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/3004
https://github.com/rancher/rke2/issues/3055


pkg/rke2/spw.go
```diff
@@ -0,0 +1,140 @@
...
+					logrus.Info("Waiting for etcd manifest")
+					return false, nil
+				}); err != nil {
+					logrus.Fatal(err)
+				}
+
+			} else if !errors.Is(err, os.ErrNotExist) {
+				// Since split-role servers exist, we don't care if no etcd manifest is found
+				logrus.Fatal(err)
+			}
```
brandond, line 101: It looks like this block is just copy-pasted down below, with the only differences being `s/etcd/kube-apiserver/` in the variable names and log messages. I think this would best be handled as a loop that iterates over `[]string{"etcd", "kube-apiserver"}`.

brandond (changes requested):

pkg/rke2/spw.go
dereknola: Consolidated into single for loop

pkg/rke2/spw.go
```diff
@@ -0,0 +1,104 @@
...
+
+// checkStaticManifests validates that the pods started with rke2 match the static manifests
+// provided in /var/lib/rancher/rke2/agent/pod-manifests. When restarting rke2, it takes time
+// for any changes to static manifests to be pulled by kubelet. Additionally this prevents errors
+// where something is wrong with the static manifests and RKE2 starts anyways.
+func checkStaticManifests(dataDir string) cmds.StartupHook {
+	return func(ctx context.Context, wg *sync.WaitGroup, args cmds.StartupHookArgs) error {
+		go func() {
+			defer wg.Done()
+			<-args.APIServerReady
```
brandond, line 45: Do we need to wait for this? We're only talking to containerd, and the apiserver manifest is one of the things we're examining, so waiting for it to start first is counterintuitive to me.

pkg/rke2/spw.go
```diff
@@ -0,0 +1,104 @@
...
+				manifestFile := filepath.Join(manifestDir, pod+".yaml")
+				if f, err := os.Open(manifestFile); err == nil {
+					if err != nil {
+						logrus.Fatal(err)
+					}
+					podManifest := v1.Pod{}
+					decoder := yaml.NewYAMLToJSONDecoder(f)
+					err = decoder.Decode(&podManifest)
+					if err != nil {
+						logrus.Fatal(err)
```
brandond, line 73:
```suggestion
						logrus.Fatalf("Failed to decode %s manifest: %v", pod, err)
```

pkg/rke2/spw.go
```diff
@@ -0,0 +1,104 @@
...
+						for _, c := range resp.Containers {
+							if c.Labels["io.kubernetes.pod.uid"] == string(podManifest.UID) {
+								logrus.Infof("Latest %s manifest deployed", pod)
+								return true, nil
+							}
+						}
+						logrus.Infof("Waiting for %s manifest", pod)
+						return false, nil
+					}); err != nil {
+						logrus.Fatal(err)
```
brandond, line 94:
```suggestion
						logrus.Fatalf("Failed to wait for latest %s manifest to be deployed", pod, err)
```

pkg/rke2/spw.go
```diff
@@ -0,0 +1,104 @@
...
+			defer conn.Close()
+
+			manifestDir := podManifestsDir(dataDir)
+
+			for _, pod := range []string{"etcd", "kube-apiserver"} {
+				manifestFile := filepath.Join(manifestDir, pod+".yaml")
+				if f, err := os.Open(manifestFile); err == nil {
+					if err != nil {
+						logrus.Fatal(err)
+					}
```
brandond, line 68: This is right below an `err == nil` test
```suggestion
```

pkg/rke2/spw.go
```diff
@@ -0,0 +1,104 @@
...
+							}
+						}
+						logrus.Infof("Waiting for %s manifest", pod)
+						return false, nil
+					}); err != nil {
+						logrus.Fatal(err)
+					}
+				} else if !errors.Is(err, os.ErrNotExist) {
+					// Since split-role servers exist, we don't care if no manifest is found
+					logrus.Fatal(err)
```
brandond, line 98:
```suggestion
					logrus.Fatalf("Failed to open %s manifest: %v", pod, err)
```

brandond (changes requested):
One thing that I think got left over in the refactor, and a couple nits on the Fatal errors. We should always provide some extra context when Fataling so that we have a string to search for if it happens.

pkg/rke2/spw.go
dereknola: Oops, I think that was still left in from old commits

pkg/rke2/spw.go
dereknola: Removed, this was leftover from all the other startup hooks that I used as a base for this one.

brandond (approved):

briandowns (approved):

PR #3111, opened 2022-06-28
updating dev doc

matttrach:
updating dev doc for validating CNIs in release notes

CNI validation was not included before, this gives good locations for validating the CNI versions when generating release notes.

### DIFF
--- a/developer-docs/upgrading_kubernetes.md
+++ b/developer-docs/upgrading_kubernetes.md
@@ -125,18 +125,29 @@ Release notes should be drafted before the release is complete. This happens in 
 The 2 primary sections of the release notes are the "Changes since ..." and the "Package Component Versions". The other sections need to be reviewed as well. The "Changes since ..." section can be fleshed out by reviewing the closed issues and pull requests for the matching milestone.
 
 #### Packaged Components
-It can be confusing to track where each number for a component is getting pulled from. Below provides
-some locations to check component versions:
+It can be confusing to track where each number for a component is getting pulled from.
 
+Here are some locations to check component versions:
+- Kubernetes: Look in `Dockerfile` for `FROM rancher/hardened-kubernetes`.
 - Etcd: Look in `scripts/version` for `ETCD_VERSION`.
-- Containerd: Look in `Dockerfile` for `rancher/hardened-containerd`.
-- Runc: Look in `Dockerfile` for `rancher/hardened-runc`.
-- Flannel: Look in `scripts/build-images` for `rancher/hardened-flannel`.
-- Calico: Look in `scripts/build-images` for `rancher/hardened-calico`.
+- Containerd: Look in `Dockerfile` for `FROM rancher/hardened-containerd`.
+- Runc: Look in `Dockerfile` for `FROM rancher/hardened-runc`.
 - Metrics-server: Look in `scripts/build-images` for `rancher/hardened-k8s-metrics-server`.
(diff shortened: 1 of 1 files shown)


developer-docs/upgrading_kubernetes.md
```diff
@@ -125,19 +125,30 @@ Release notes should be drafted before the release is complete. This happens in
...
+- Runc: Look in `Dockerfile` for `FROM rancher/hardened-runc`.
 - Metrics-server: Look in `scripts/build-images` for `rancher/hardened-k8s-metrics-server`.
 - CoreDNS: Look in `scripts/build-images` for `rancher/hardened-coredns`.
-- Ingress-Nginx: Look in `Dockerfile` find the line with `charts/rke2-ingress-nginx`.
+- Ingress-Nginx: Look in `Dockerfile` find the line with `CHART_FILE=/charts/rke2-ingress-nginx`.
 - Helm-controller: Look in `go.mod` for `helm-controller`.
 
+#### Available CNIs
+
+All CNI versions are in `scripts/build-images`.
```
brandond, line 142: You might also mention that these need to match the versions referenced by the chart version specified for that CNI in the Dockerfile.

briandowns (approved):

developer-docs/upgrading_kubernetes.md
matttrach: what do we do if there is a conflict between these two sources of information?

brandond: The chart determines what will be deployed. If the build-images scripts say something different than the chart, then someone forgot to update the scripts when bumping the chart and we have a problem.

nikolaishields (approved):

developer-docs/upgrading_kubernetes.md
matttrach: I took some time to trace the code and it looks like the build-images script generates a list of the images which will be built and uploaded to the chart repo, then later on (when building rke2) the Dockerfile has the versions of what charts to pull from the chart repo, with help from the build-chart.sh. 
Am I missing something?
How do we validate the build-images script?
Is that just dependent on the individuals adding charts to rke2 to keep up to date?

matttrach: I went ahead and added a section talking about how to validate without the build script, just in case.

matttrach: @brandond is this good?

brandond: The suggestions are good, but the bit about `The build images script might not be up to date` should probably call out that this is a problem state - if the build-images script is not up to date, we have a release-blocking issue with the airgap image list, and the issue will need to be corrected so that the script generates an image list that is in sync with the dockerfile and charts.

matttrach: This sort of indicates that the release team should be validating the build-images script, right?
Meaning that the release captain should be validating the CNI container versions for the charts every time and updating the build-images script if they find discrepancies?

brandond: If our process is working correctly, then:
1. Anyone who bumps a chart or component that changes an image tag would also bump the tag in build-images.
2. QA will validate airgap installs as part of the release process; if an image is missing from the list the release will be failed at the RC stage.
3. By the time we've made it to a release and are writing release notes, the airgap image list (and build-image script) will have been validated by QA.

brandond (approved):
thanks for being so comprehensive with the process docs!

PR #3154, opened 2022-07-14
Add a readme to explain our implementation of ADRs

matttrach:
Signed-off-by: matttrach <matt.trachier@suse.com>

#### Proposed Changes ####

This adds a README.md to our ADR directory so that readers are able to easily understand what it is for and why they exist.
This was confusing to me (I did not realize what ADR stands for or the format for it), so I am hoping to give the context I have to the next person. 
My understanding (based on how we are using this tool), is to document and version decisions that are made in team design discussions.

#### Types of Changes ####

Documentation change, potential process change (the previous process is not fully documented)

#### Further Discussion ####

I am presenting this as a work in progress because the adoption of this tool is low and I would like to discuss the ideas which I have presented.
I am hoping for a few outcomes:
1. a definition of the process
2. a definition of the format for the file
3. a review period, so that we recognize a reasonable timeframe for a decision to be made
4. a reconsider date, so that we make sure to review decisions we have made, add context where necessary, and make sure we still agree with those decisions

### DIFF
--- a/docs/adrs/README.md
+++ b/docs/adrs/README.md
@@ -1,0 +1,67 @@
+# Architectural Decision Records (ADRs)
+
+ADRs are a record of the arguments and decisions made to change process or software architecture.
+The idea is for these records to be regularly reviewed, updated, and documented so that new people or external parties
+ to a project can read along and understand the points of a decision, the context, 
+ and in general can make educated decisions based on previous discussions.
+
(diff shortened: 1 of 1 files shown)


brandond (approved):

nikolaishields (approved):
This is a great idea @matttrach!

PR #3166, opened 2022-07-19
Consolidate staticPod timeout to static 30 minutes

dereknola:

#### Proposed Changes ####
Simplify and consolidate the timeout for Static Pod Manifest check. Now there is only 1 timeout, a static 20 second check for 30 minutes.

#### Verification ####
RKE2 startup on single node.

#### Linked Issues ####
https://github.com/harvester/harvester/issues/2336

### DIFF
--- a/pkg/rke2/spw.go
+++ b/pkg/rke2/spw.go
@@ -2,6 +2,7 @@ package rke2
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
@@ -11,26 +12,17 @@ import (
 	"github.com/k3s-io/k3s/pkg/cli/cmds"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
-// total timeout of 4 minutes
+// Check every 20 seconds for a total timeout of 30 minutes
 var podCheckBackoff = wait.Backoff{
-	Steps:    12,
-	Duration: 15 * time.Second,
+	Steps:    90,
+	Duration: 20 * time.Second,
 	Factor:   1.0,
-	Jitter:   0.1,
-}
-
-// total timeout of 2047 seconds (34 minutes)
-var criBackoff = wait.Backoff{
-	Steps:    12,
-	Duration: 1 * time.Second,
-	Factor:   2,
 	Jitter:   0.1,
 }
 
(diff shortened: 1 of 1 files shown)


pkg/rke2/spw.go
```diff
@@ -42,57 +34,55 @@ func checkStaticManifests(dataDir string) cmds.StartupHook {
 	return func(ctx context.Context, wg *sync.WaitGroup, args cmds.StartupHookArgs) error {
 		go func() {
 			defer wg.Done()
+			if err := wait.ExponentialBackoff(podCheckBackoff, func() (done bool, err error) {
```
brandond, line 37: I think at this point we have just recreated [PollImmediate](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate)
```suggestion
			if err := wait.PollImmediate(20 * time.Second, 30 * time.Minute, func() (done bool, err error) {
```

brandond (changes requested):

pkg/rke2/spw.go
```diff
@@ -42,57 +26,55 @@ func checkStaticManifests(dataDir string) cmds.StartupHook {
...
+					if f, err := os.Open(manifestFile); err == nil {
+						podManifest := v1.Pod{}
+						decoder := yaml.NewYAMLToJSONDecoder(f)
+						err = decoder.Decode(&podManifest)
+						if err != nil {
+							logrus.Fatalf("Failed to decode %s manifest: %v", pod, err)
+						}
+						podFilter := &runtimeapi.ContainerFilter{
+							LabelSelector: map[string]string{
+								"io.kubernetes.container.name": pod,
```
brandond, line 52: This will cause problems if someone ever deploys a pod with a container named 'etcd' or 'kube-apiserver. We should just filter on the pod UID we're looking for:
```suggestion
								"io.kubernetes.pod.uid": podManifest.UID,
```

pkg/rke2/spw.go
```diff
@@ -42,57 +26,55 @@ func checkStaticManifests(dataDir string) cmds.StartupHook {
...
+							},
+						}
 						resp, err := cRuntime.ListContainers(ctx, &runtimeapi.ListContainersRequest{Filter: podFilter})
 						if err != nil {
 							return false, err
 						}
+						found := false
```
brandond, line 59: If we filter containers on `io.kubernetes.pod.uid` then we can just set `found := len(resp.Containers) > 0`

brandond (changes requested):

pkg/rke2/spw.go
```diff
@@ -42,57 +26,55 @@ func checkStaticManifests(dataDir string) cmds.StartupHook {
...
 								logrus.Infof("Latest %s manifest deployed", pod)
-								return true, nil
+								found = true
 							}
 						}
-						logrus.Infof("Waiting for %s manifest", pod)
-						return false, nil
-					}); err != nil {
-						logrus.Fatalf("Failed to wait for latest %s manifest to be deployed: %v", pod, err)
+						if !found {
```
brandond, line 66: If we filter containers on `io.kubernetes.pod.uid` then we can just do:
```suggestion
						if len(resp.Containers) < 1 {
```

brandond (changes requested):

brandond (approved):

PR #3192, opened 2022-07-28
ADR, uncheck prerelease last

matttrach:
Signed-off-by: matttrach <matt.trachier@suse.com>

### DIFF
--- a/docs/adrs/when-prerelease-is-unchecked.md
+++ b/docs/adrs/when-prerelease-is-unchecked.md
@@ -1,0 +1,42 @@
+# When Prerelease Box is Unchecked
+
+The release order is important, we need to make sure we are not introducing any bugs in the order of release.
+
+## Established
+
+2022-07-28
+
+## Revisit by
+
+2023-07-28
+
+## Subject
+
+Given approval for GA release, after channel server is updated,
+  the last operation we perform is to uncheck the prerelease box on the releases.
+
+## Status
+
+Approved
+
+## Context
+
+### Strength of doing process
+
+- this is a good final marker for the release
+
+### Weakness of doing process
+
+- the channel server will not be aware of the release until the box is unchecked
+
+### Threats involved in not doing process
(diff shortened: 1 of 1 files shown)


brandond:
Based on the discussion from the design meeting, lets update this ADR to capture the current interactions between GH releases, the channel server, and downstream tooling such as the install script and System Upgrade Controller. In particular I'd like to make sure we document why the pre-release flag is cleared as soon as all GH artifacts are published and packaging has been tagged.

Also, upon taking a second look at the channel server config, I believe that we should also be clearing the pre-release flag for non-GA releases, as the `testing` channel is set up to specifically include only non-GA tags: https://github.com/rancher/rke2/blob/727b1b1518fa9e4e0cd387f6f16adf018bdd19b6/channels.yaml#L13-L14

Without doing this, we're not honoring the contract of publishing releases to the testing channel.

brandond (changes requested):
Based on the discussion from the design meeting, lets update this ADR to capture the current interactions between GH releases, the channel server, and downstream tooling such as the install script and System Upgrade Controller. In particular I'd like to make sure we document why the pre-release flag is cleared as soon as all GH artifacts are published and packaging has been tagged.

Also, upon taking a second look at the channel server config, I believe that we should also be clearing the pre-release flag for non-GA releases, as the `testing` channel is set up to specifically include only non-GA tags: https://github.com/rancher/rke2/blob/727b1b1518fa9e4e0cd387f6f16adf018bdd19b6/channels.yaml#L13-L14

Without doing this, we're not honoring the contract of publishing releases to the testing channel.

This makes the pre-release checkbox essentially an indicator that CI is not done yet, removing any residual question around whether or not that flag is used to denote GA releases, and makes the tags the sole indicator of whether a release is GA or not.

nikolaishields (approved):

matttrach:
closing this as stale

Issue #3104, opened 2022-06-23
Static pod manifest customization


QtSYnT:
**Is your feature request related to a problem? Please describe.**

RKE2 provides limited ability to configure the generated static pod manifest files for control plane components. Some items like liveness probes are unable to be configured due to this abstraction.

**Describe the solution you'd like**

A generic way to override values in the generated manifests would be ideal or alternatively a way to supply an existing manifest file to be used instead.

**Describe alternatives you've considered**

There are limited workarounds that can be applied like for example disabling certain control plane components and then supplying the manifest ourselves but this isn't possible for all components.

**Additional context**

N/A

brandond:
Are there any specific bits of the probes you'd like to see changed? Generic customization of manifest fields is unlikely to land without a specific use case, but if there's something in particular you're wanting to change we may be able to accommodate that more quickly.

QtSYnT:
A good start would be the ability to manipulate the timing/threshold values for the existing probes. Next level up from that would be a way to define probes that don't already exist and their respective values.

I'm open to "hacky" solutions that can achieve this so if even some unsupported pathways exists for it or could be opened up that would be great.

brandond:
Can you describe what you think we're missing from the current control-plane component static pod health checks? I believe that most of ours mirror the kubeadm defaults pretty closely.

QtSYnT:
I'd like to see a way to define startup/liveness probes for static pods to facilitate monitoring. So for example right now the api server and kube proxy don't have any probes defined.

brandond:
See what you think of that PR. If all you want is to add health checks to those two static pods, that should do it.

QtSYnT:
Sorry about the delay on this I needed to verify a few more things before I got back to you. So it's great that the liveness probes now exist for all static pods but the following are missing if we're using kubeadm as a reference.

The following are missing startup probes:

- etcd
- kube-apiserver
- kube-controller-manager
- kube-scheduler
- kube-proxy

The kube-apiserver is also missing the readiness probe.

The startup probe is critical in accurately reporting the "Ready" status of the pod. When it's not defined the only time the pod will report a "Not Ready" status is a split second when restarting from the liveness probe and then again once the pod goes on CrashLoopBackOff. That's the functionality observed in RKE2 while in contrast kubeadm installs will retain a "Not Ready" status until the probe succeeds unlike RKE2/k3s.

[...]

PR #3204, opened 2022-08-02
Improve static pod probing

brandond:
#### Proposed Changes ####

Improve static pod probing

* Add Startup/Readiness probes
* Sync default probe thresholds and timing with kubadm
* Make probe thresholds and timing configurable from the CLI, using an interface similar to control-plane resources and requests:
  `--control-plane-probe-configuration=etcd-startup-initial-delay-seconds=42,kube-apiserver-readiness-period-seconds=15`

#### Types of Changes ####

Enhancement

#### Verification ####

See linked issue

#### Linked Issues ####

https://github.com/rancher/rke2/issues/3104

#### User-Facing Change ####
```release-note
RKE2 static pods now include readiness, liveness, and startup probes with defaults that match those configured by kubeadm.
RKE2 static pod probe timings and thresholds can be customized with the `--control-plane-probe-configuration` flag.
```

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -86,20 +86,30 @@ type ControlPlaneMounts struct {
 	CloudControllerManager []string
 }
 
+type ControlPlaneProbeConfs struct {
+	KubeAPIServer          staticpod.ProbeConfs
+	KubeScheduler          staticpod.ProbeConfs
+	KubeControllerManager  staticpod.ProbeConfs
+	KubeProxy              staticpod.ProbeConfs
+	Etcd                   staticpod.ProbeConfs
+	CloudControllerManager staticpod.ProbeConfs
+}
+
 type StaticPodConfig struct {
-	ManifestsDir          string
-	ImagesDir             string
-	Resolver              *images.Resolver
-	CloudProvider         *CloudProviderConfig
-	DataDir               string
-	AuditPolicyFile       string
-	KubeletPath           string
-	ControlPlaneResources ControlPlaneResources
-	ControlPlaneMounts    ControlPlaneMounts
-	ControlPlaneEnv       ControlPlaneEnv
-	CISMode               bool
-	DisableETCD           bool
-	IsServer              bool
+	ControlPlaneResources
+	ControlPlaneProbeConfs
+	ControlPlaneEnv
+	ControlPlaneMounts
+	ManifestsDir    string
+	ImagesDir       string
+	Resolver        *images.Resolver
+	CloudProvider   *CloudProviderConfig
+	DataDir         string
+	AuditPolicyFile string
+	KubeletPath     string
+	CISMode         bool
+	DisableETCD     bool
+	IsServer        bool
 }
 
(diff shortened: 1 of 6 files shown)


pkg/rke2/rke2_linux.go
```diff
@@ -192,6 +202,188 @@ func initExecutor(clx *cli.Context, cfg Config, isServer bool) (*podexecutor.Sta
...
+			InitialDelaySeconds: 10,
+			TimeoutSeconds:      5,
+			FailureThreshold:    24,
+			PeriodSeconds:       10,
+		},
+	}
+
+	var parsedProbeConf = make(map[string]int32)
+
+	if cfg.ControlPlaneProbeConf != "" {
```
dereknola, line 358: Could you convert 358-380 into a separate function? and then write a unit test for that function?

dereknola (changes requested):

pkg/rke2/rke2_linux.go
brandond: Honestly this whole function has sort of grown out of control. The bits for resources, environment variables, mounts, and probe configs probably all merit going into their own functions.

pkg/rke2/rke2_linux_test.go
```diff
@@ -0,0 +1,37 @@
+//go:build linux
+// +build linux
+
+package rke2
+
+import (
+	"flag"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
dereknola, line 10: Thanks for adding a unit test!! However, we don't use testify in our Unit tests, we generate them via `gotests` tool. Please convert this over to that format. See https://github.com/rancher/rke2/blob/master/developer-docs/testing.md#unit-tests for more details.

dereknola (changes requested):

pkg/rke2/rke2_linux_test.go
brandond: I wrap this in a table-driven design, but I don't see that use of table-driven tests are mutually exclusive with use of `assert` to check the actual values of the returned data. gotest still leaves it up to you to check the type and value of your data against the desired output, create an error message, and so on. The basic assert functions are a very useful shorthand for that.

dereknola (approved):

pkg/podexecutor/staticpod.go
```diff
@@ -86,20 +86,30 @@ type ControlPlaneMounts struct {
...
+	ImagesDir       string
+	Resolver        *images.Resolver
+	CloudProvider   *CloudProviderConfig
+	DataDir         string
+	AuditPolicyFile string
+	KubeletPath     string
+	CISMode         bool
+	DisableETCD     bool
+	IsServer        bool
+	ControlPlaneResources
```
briandowns, line 109: Generally, convention is to have the embedded structs at the top of the struct. Less important, it's best to keep the bools at the bottom.

briandowns (approved):

pkg/podexecutor/staticpod.go
brandond: done, thanks!

Issue #3247, opened 2022-08-18
Add a flag to skip systemctl deamon restart in install script


dramich:
**Is your feature request related to a problem? Please describe.**
Currently the install script assumes that systemctl is available when doing a tar install, so when using the script to build images and the system init hasn't been ran the install script fails.
**Describe the solution you'd like**
Add a flag to skip the reload

mdrahman-suse:
### Validated on RC v1.25.0-rc5+rke2r1
  
## Environment Details

  
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
Linux ip-172-31-42-243 5.15.0-1011-aws #14-Ubuntu SMP Wed Jun 1 20:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
```
  
*Cluster Configuration:*

```
Single node
```
  
*Config.yaml:*

```
write-kubeconfig-mode: 644
cni: calico
token: summerheat
```
  
## Testing Steps

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/rke2 && sudo cp config.yaml /etc/rancher/rke2
```
2. Install RKE2 as root: 
  - Replication: `curl -sfL https://get.rke2.io | INSTALL_RKE2_VERSION=v1.24.4+rke2r1 sh -`
  - Validation: `curl -sfL https://get.rke2.io | INSTALL_RKE2_VERSION=v1.25.0-rc5+rke2r1 INSTALL_RKE2_SKIP_RELOAD=true sh -`
3. Check logs: `journalctl --system --no-pager -n 20 | grep Reloading`
  
**Replication Results:**
- rke2 version used for replication:

```
rke2 version v1.24.4+rke2r1 (749c87a7da6482dec999df44f25578486158b21d)
go version go1.18.1b7
```
 

- Logs shows
```
$ journalctl --system --no-pager -n 20 | grep Reloading
Sep 14 23:03:52 ip-172-31-42-243 systemd[1]: Reloading.
Sep 14 23:03:52 ip-172-31-42-243 systemd[1]: Reloading.
Sep 14 23:03:52 ip-172-31-42-243 systemd[1]: Reloading.
Sep 14 23:04:13 ip-172-31-42-243 systemd[1]: Reloading.
```
 
[...]

PR #3248, opened 2022-08-18
Update install script with option to skip reload

dramich:
#### Proposed Changes ####
Add a new envvar INSTALL_RKE2_SKIP_RELOAD that skips the systemctl
daemon-reload call. This is useful when using the install script
to build images that contain rke2 but init has not run yet.

#### Types of Changes ####

Enhancement

#### Verification ####

Run the script with and without the new flag set. On a fresh box systemctl won't see the rke2 services until after the reload is called, so running the script with and without the flag should show the difference. Or run in a container before init is called and watch systemctl bomb 😄 

#### Linked Issues ####
https://github.com/rancher/rke2/issues/3247

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -55,6 +55,10 @@ fi
 #     rather than the downloading the files from the internet.
 #     Default is not set.
 #
+#   - INSTALL_RKE2_SKIP_RELOAD
+#     If set, the install script will skip reloading systemctl daemon after the tar has been extracted and systemd units
+#     have been moved.
+#     Default is not set.
 
 
 # info logs the given argument at info log level.
@@ -561,7 +565,10 @@ do_install_tar() {
     install_airgap_tarball
     verify_tarball
     unpack_tarball
-    systemctl daemon-reload
+
+    if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
+        systemctl daemon-reload
+    fi
 }
 
 do_install() {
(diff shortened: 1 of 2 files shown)


brandond:
Is the issue that systemd isn't installed yet when building the image? In the K3s installer we wrap things in a check to see if `systemctl` is available, perhaps we should do that here as well?

https://github.com/k3s-io/k3s/blob/master/install.sh#L694-L698

dramich:
`systemctl` does exist at this point but `systemd` is not running since init has not run yet, so the check passes but still fails in the same way as before.

install.sh
```diff
@@ -561,6 +565,9 @@ do_install_tar() {
     install_airgap_tarball
     verify_tarball
     unpack_tarball
+
+    [ "${INSTALL_RKE2_SKIP_RELOAD}" = true ] && return
+
     systemctl daemon-reload
```
brandond, line 571: I think this would better match the convention we use for other vars in this script
```suggestion
    if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
        systemctl daemon-reload
    fi
```

brandond (changes requested):
one nit; makes sense otherwise

install.sh
dramich: Sounds good, updated.

brandond (approved):

dramich:
@brandond Can you run the workflows?

galal-hussein (approved):

mdrahman-suse:
@dramich can I get some assistance on how to test this? I tried installing rke2 on fresh node by running `curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION=v1.25.0-rc4+rke2r1 INSTALL_RKE2_SKIP_RELOAD=true sh -` and when I try to check the status of the service using `$ systemctl status rke2-server`, I do see the service getting loaded... same happens if I dont use the var `INSTALL_RKE2_SKIP_RELOAD=true`
```
rke2-server.service - Rancher Kubernetes Engine v2 (server)
     Loaded: loaded (/usr/local/lib/systemd/system/rke2-server.service; disabled; vendor preset: enabled)
     Active: inactive (dead)
       Docs: https://github.com/rancher/rke2#readme
```

Issue #3254, opened 2022-08-22
Error in the logs even when serviceLB is disabled and config.ServiceLBNamespace is empty 


mdrahman-suse:
**Environmental Info:**
RKE2 Version: 

```
rke2 version v1.24.4-rc1+rke2r1 (48205d8a01cfb3084d2a430bd01a005451abc9cf)
go version go1.18.1b7
```

Node(s) CPU architecture, OS, and Version: 

```
Linux  5.3.18-59.37-default #1 SMP Mon Nov 22 12:29:04 UTC 2021 (d10168e) x86_64 x86_64 x86_64 GNU/Linux
NAME="SLES"
VERSION="15-SP3"
VERSION_ID="15.3"
```

Cluster Configuration:

Single server

**Describe the bug:**

The log is getting flooded with errors below even when serviceLB is disabled and config.ServiceLBNamespace is empty 

**Steps To Reproduce:**

- config.yaml
```
write-kubeconfig-mode: 644
token: testing
cni: calico
```
- Installed RKE2: `curl -sfL https://get.rke2.io | INSTALL_RKE2_VERSION=v1.24.4-rc1+rke2r1 sh -`
- Enable and Start rke2 service: `sudo systemctl enable rke2-server.service --now`

**Expected behavior:**

- Cluster is functional with no errors

**Actual behavior:**

- Cluster is functional with errors
```
Aug 22 21:19:16 ip-172-31-35-183 rke2[2714]: time="2022-08-22T21:19:16Z" level=error msg="error syncing 'calico-system/calico-kube-controllers-metrics': handler svccontroller: an empty namespace may not be set when a resource name is provided, requeuing"
Aug 22 21:19:36 ip-172-31-35-183 rke2[2714]: time="2022-08-22T21:19:36Z" level=error msg="error syncing 'kube-system/rke2-coredns-rke2-coredns': handler svccontroller: an empty namespace may not be set when a resource name is provided, requeuing"
Aug 22 21:19:55 ip-172-31-35-183 rke2[2714]: time="2022-08-22T21:19:55Z" level=error msg="error syncing 'calico-system/calico-typha': handler svccontroller: an empty namespace may not be set when a resource name is provided, requeuing"
Aug 22 21:20:20 ip-172-31-35-183 rke2[2714]: time="2022-08-22T21:20:20Z" level=error msg="error syncing 'default/kubernetes': handler svccontroller: an empty namespace may not be set when a resource name is provided, requeuing"
[...]

PR #3255, opened 2022-08-22
Work around issue with empty servicelb namespace

brandond:
#### Proposed Changes ####

Work around issue with empty servicelb namespace

There is some oddness to how servicelb works when disabled - the the wrangler handlers are registered even when it's disabled, to cleanup any svclb pods that were created when it was enabled. In RKE2 we really just want it off completely, we don't even need to worry about cleanup.

For the next release we should fix this on the k3s side by properly defaulting the value, or (better yet) not doing any of the setup when it's disabled and the namespace is empty. 

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3254


briandowns (approved):

wgsuse (approved):
LGTM

PR #3295, opened 2022-09-01
Update Kubernetes image to v1.25.0-rke2r1-build20220901

galal-hussein:
Signed-off-by: galal-hussein <hussein.galal.ahmed.11@gmail.com>


brandond:
I think the check is raising a valid error. The hardened-kubernetes tag is `v1.25.0-rke2r2` but you tagged RKE2 as rke2r1. The two release versions must match or things will be very confused downstream. Why did we skip directly to r2 of the Kubernetes image?

https://github.com/rancher/rke2/blob/v1.25.0-rc1%2Brke2r1/Dockerfile#L135

cwayne18 (approved):

briandowns (approved):

PR #3338, opened 2022-09-13
Fix config directory permissions

galal-hussein:
#### Proposed Changes ####

Fix permissions of /etc/rancher/rke2 when created during startup.

#### Types of Changes ####

bugfix

#### Verification ####

see linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3339

#### Further Comments ####

Will need additional refactoring later to consolidate PSA and Audit Policy file creation logic.


brandond (changes requested):
Lets fix https://github.com/rancher/rke2/blob/d32859fee17610b8f2220abb8358654d4e160106/pkg/podexecutor/staticpod.go#L595 also. Later on we can refactor these to use a common implementation for writing files to /etc/rancher/rke2 that uses consistent permissions.

brandond (approved):

brooksn (approved):

PR #3371, opened 2022-09-21
update k8s to 1.23.12

luisrojas2:


luisrojas2:
Hardened build should be created first.

Issue #2848, opened 2022-05-03
Recommended Configurations for Fapolicyd


andygodish:
**Is your feature request related to a problem? Please describe.**

When working on RHEL 8 machines running OpenSCAP, checks are being applied against fapolicy. 

For example:
```
error="failed to create containerd task: : fork/exec /var/lib/rancher/rke2/data/v1.21.11-rke2r1-a2c827fdad5b/bin/containerd-shim-runc-v2: operation not permitted: unknown"
``` 
Disabling fapolicyd will resolve this, but is not always a valid option due to the security posture of individual organizations. Is there an established set of configurations that allow for openscap compliance as it pertains to fapolicy findings? 

**Describe the solution you'd like**

Establishment of a set of configurations that allow for openscap compliance as it pertains to fapolicy findings.

**Describe alternatives you've considered**

I was able to bypass the `containerd exec` error by added the following fapolicyd rule: 

```
"allow perm=any uid=0 : dir.=/var/lib/rancher/rke2/”
```
However, this results in another error in reference to CoreDNS that I have not been able to resolve.

FruityWelsh:
In prelim testing I found this config seems to be working: 
```
allow perm=any all : dir=/var/lib/rancher/
allow perm=any all : dir=/opt/cni/
allow perm=any all : dir=/run/k3s/
allow perm=any all : dir=/var/lib/kubelet/
```

This is on a previously running system (with fapolicy set permissive). I then added each one of these rules and rebooted the node to test that rke2-server would start without errors and without audit exceptions.

Checking audit errors with ```sudo ausearch --start today -m fanotify -i```.

rancher-max:
### Validated using install script from `master` branch

Note that I am not an `fapolicyd` expert by any means, so if anything from the below steps appears incorrect, please let me know. I do believe this is working based on my results though, so I am closing this issue.
  
## Environment Details

  
*Infrastructure*
- [x] Cloud (AWS)
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
$ uname -a && cat /etc/os-release 
Linux ip-172-31-29-132.us-east-2.compute.internal 4.18.0-372.9.1.el8.x86_64 #1 SMP Fri Apr 15 22:12:19 EDT 2022 x86_64 x86_64 x86_64 GNU/Linux
NAME="Red Hat Enterprise Linux"
VERSION="8.6 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
[...]

PR #3416, opened 2022-10-03
Add fapolicyd configuration rules

galal-hussein:
#### Proposed Changes ####

- Added fapolicyd configuration rules on RHEL systems

The install script will check for fapolicy existence and if it exists it will add the file  `/etc/fapolicyd/rules.d/80-rke2.rules` to the rules directory which contains:
```
allow perm=any all : dir=/var/lib/rancher/
allow perm=any all : dir=/opt/cni/
allow perm=any all : dir=/run/k3s/
allow perm=any all : dir=/var/lib/kubelet/
```
- Add uninstall section for fapolicyd configuration rules as well

#### Types of Changes ####

bug fix

#### Verification ####
On RHEL machine:

- Install fapolicyd `yum install -y fapolicyd`
- Install rke2 using the install script

#### Linked Issues ####
- https://github.com/rancher/rke2/issues/2848
- 
- #### Further Comments ####

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -595,8 +595,8 @@ allow perm=any all : dir=/opt/cni/
 allow perm=any all : dir=/run/k3s/
 allow perm=any all : dir=/var/lib/kubelet/
 EOF
-        fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
         if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
+            fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
             systemctl restart fapolicyd
         fi
     fi
--- a/install.sh
+++ b/install.sh
@@ -191,6 +191,17 @@ verify_downloader() {
     return 0
 }
 
+# verify_fapolicyd verifies existence of
+# fapolicyd executable.
+verify_fapolicyd() {
+    cmd="$(command -v "fapolicyd")"
+    if [ -z "${cmd}" ]; then
+        return 1
+    fi
+
+    return 0
+}
+
 # setup_tmp creates a temporary directory
 # and cleans up when done.
 setup_tmp() {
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -574,6 +585,21 @@ do_install_tar() {
...
+        verify_fapolicyd || return
+        # setting rke2 fapolicyd rules
+        cat <<-EOF >>"/etc/fapolicyd/rules.d/80-rke2.rules"
+allow perm=any all : dir=/var/lib/rancher/
+allow perm=any all : dir=/opt/cni/
+allow perm=any all : dir=/run/k3s/
+allow perm=any all : dir=/var/lib/kubelet/
+EOF
+        fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
+        systemctl restart fapolicyd
```
brandond, line 599: We should probably honor `INSTALL_RKE2_SKIP_RELOAD` and not try to make any changes to the running system if it's set - see https://github.com/rancher/rke2/pull/3248

brandond (changes requested):

install.sh
galal-hussein: will fix

install.sh
```diff
@@ -574,6 +585,23 @@ do_install_tar() {
...
+        # setting rke2 fapolicyd rules
+        cat <<-EOF >>"/etc/fapolicyd/rules.d/80-rke2.rules"
+allow perm=any all : dir=/var/lib/rancher/
+allow perm=any all : dir=/opt/cni/
+allow perm=any all : dir=/run/k3s/
+allow perm=any all : dir=/var/lib/kubelet/
+EOF
+        fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
+        if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
+            systemctl restart fapolicyd
```
brandond, line 600: I think this should go in there too... if this script is being run during system image creation (which is what that flag was added to support) then we don't need to load the rules either, as fapolicyd won't be running yet, for the same reason that systemd isn't running either.

```suggestion
        if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
            fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
            systemctl restart fapolicyd
```

brandond (changes requested):

dereknola (approved):

briandowns (approved):

brandond (approved):

Issue #3469, opened 2022-10-17
RKE2 truncates manifests that contain a HelmChart resource as the first document


brandond:
**Environmental Info:**
RKE2 Version: 
n/a

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
RKE2 truncates manifests that contain a HelmChart resource as the first document

**Steps To Reproduce:**
* Drop the following file in `/var/lib/rancher/rke2/server/manifests` prior to starting RKE2
* Note that the file on disk is truncated to just the cert-manager HelmChart resource

```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  namespace: kube-system
  name: cert-manager
spec:
  targetNamespace: cert-manager
  version: v1.6.1
  chart: cert-manager
  repo: https://charts.jetstack.io
  set:
    installCRDs: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: cert-manager
---
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  name: rancher
  namespace: kube-system
spec:
  targetNamespace: cattle-system
  version: v2.6.8
  chart: rancher
  repo: https://releases.rancher.com/server-charts/latest
  set:
    ingress.tls.source: "rancher"
    hostname: "rancher.example.com"
    antiAffinity: "required"
    replicas: 1
---
apiVersion: v1
kind: Namespace
metadata:
  name: cattle-system
```

**Expected behavior:**
Cluster configuration values are passed in to all charts in the manifest; other resource types are not changed or lost

**Actual behavior:**
Everything after the first HelmChart resource is removed from the file

**Additional context / logs:**
* Mentioned at https://github.com/rancher/rke2/issues/527#issuecomment-742119875
* https://github.com/rancher/rke2/blob/ace2b88fecedacc0a5aec5c734bbe162f1f80a50/pkg/bootstrap/bootstrap.go#L253-L255

VestigeJ:
Reproduced using VERSION=v1.25.3+rke2r1 
Validated using COMMIT=9364f9d4cdeb7081122491a9f9e4e3b1a6d2ae53

##Environment Details 
*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and Version:*

Linux 5.14.21-150400.24.11-default x86_64 GNU/Linux SUSE Linux Enterprise Server 15 SP4

*Cluster Configuration:*
```
[...]

Issue #527, opened 2020-11-11
Should not rewrite HelmChart manifests using HelmChart type


brandond:
As per @ibuildthecloud:

>  https://github.com/rancher/rke2/blob/master/pkg/bootstrap/bootstrap.go#L365  We probably dont' want to read/write this file using a struct.  While the typed nature is nice, the downside is if HelmChart has a new field in the struct we don't know about, they we will drop it.  So it's better to just treat this is an unstructured object.

Discussed using the unstructured type from https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured instead

brandond:
While we're here we should probably handle manifests with multiple documents as well. Failure to do so is the cause of https://github.com/rancher/rancher/issues/30012#issuecomment-741228968 - the first document is a namespace, so we don't handle helmChart portion. If the HelmChart had happened to be first, we would have dropped the namespace when rewriting and broke it in a different way.

ShylajaDevadiga:
Closing this issue in favor of validation covered in https://github.com/rancher/rke2/issues/3469

PR #3470, opened 2022-10-18
Fix handling of manifests with multiple resources

brandond:
#### Proposed Changes ####

Use wrangler's YAML helpers to read/write the manifest with proper support for multiple documents (objects).

Also uses Unstructured helpers instead of the native type to rewrite HelmChart resources without the potential for dropping unknown fields.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3469
* https://github.com/rancher/rke2/issues/527

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -1,12 +1,9 @@
 package bootstrap
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -287,77 +283,94 @@ func setChartValues(manifestsDir string, nodeConfig *daemonconfig.Node, cfg cmds
...
-	}
+	var changed bool
 
-	// Generally we should avoid using Set on HelmCharts since it cannot be overridden by HelmChartConfig,
-	// but in this case we need to do it in order to avoid potentially mangling the ValuesContent field by
-	// blindly appending content to it in order to set values.
-	if chart.Spec.Set == nil {
-		chart.Spec.Set = map[string]intstr.IntOrString{}
-	}
+OBJECTS:
```
dereknola, line 310: What does this line do? Are we using a `goto` statement somewhere else to reference this?

brandond: See a few lines below where if the `spec.set` field is improperly formatted, we break out of processing this object instead of failing over and over again for every key.

dereknola (approved):

PR #3478, opened 2022-10-19
Prevent script fail when `fapolicyd` doesn't exist

### LINKED ISSUE #3477: RKE2 install script fails when fapolicyd does not exist
**Environmental Info:**
RKE2 Version: N/A

Node(s) CPU architecture, OS, and Version: N/A

Cluster Configuration: N/A

**Describe the bug:**

#3416 adds installation and uninstallation of `fapolicyd` configuration rules on RHEL systems
However, if `fapolicyd` does not exist, the script fails instead of proceeding

https://github.com/rancher/rke2/blob/c51f865a630672fcf7b36b8ea4818d5fd9426ad6/install.sh#L194-L203

https://github.com/rancher/rke2/blob/c51f865a630672fcf7b36b8ea4818d5fd9426ad6/install.sh#L588-L590

This is because an explicit return code is not passed to `return` and as per `return --help`:
```console
$ return --help
return: return [n]
    Return from a shell function.

    Causes a function or sourced script to exit with the return value
    specified by N.  If N is omitted, the return status is that of the
    last command executed within the function or script.

    Exit Status:
    Returns N, or failure if the shell is not executing a function or script.
```

so it should be `verify_fapolicyd || return 0`

**Steps To Reproduce:**

Install RKE2 on a system that does not have `fapolicyd` using the install script. In my case, I used:
```console
curl -vsfL https://get.rke2.io | sh -
```

Here are the full steps to reproduce the issue:

```console
> docker run -it --rm registry.access.redhat.com/ubi8/ubi:latest
[root@d364ac4b088d /]# export INSTALL_RKE2_TYPE=server
[root@d364ac4b088d /]# export INSTALL_RKE2_VERSION=v1.24.6+rke2r1
[root@d364ac4b088d /]# export DEBUG=1
[root@d364ac4b088d /]# curl -vsfL https://get.rke2.io | sh -
```

**Expected behavior:**

Installation proceeds as usual

**Actual behavior:**

Installation fails with non-zero exit code

**Additional context / logs:**
```console
[...]
+ setup_fapolicy_rules
+ '[' -r /etc/redhat-release ']'
+ verify_fapolicyd
++ command -v fapolicyd
+ cmd=
+ '[' -z '' ']'
+ return 1
+ return
```

sereneshikari:
#### Proposed Changes ####

Fixes #3477 

#### Types of Changes ####

Fixes bug introduced by #3416 

#### Verification ####

Follow reproduction steps outlined in #3477 and confirm that actual behaviour is the same as expected behaviour

#### Linked Issues ####

#3477 
#3416 

#### Further Comments ####

I know this is a very trivial fix but it broke the automated RKE2 installation in my CI/CD environment 😅

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -587,7 +587,7 @@ do_install_tar() {
 
 setup_fapolicy_rules() {
     if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ] || [ -r /etc/rocky-release ]; then
-        verify_fapolicyd || return
+        verify_fapolicyd || return 0
         # setting rke2 fapolicyd rules
         cat <<-EOF >>"/etc/fapolicyd/rules.d/80-rke2.rules"
 allow perm=any all : dir=/var/lib/rancher/


briandowns (approved):

sereneshikari:
Thanks for the approval @briandowns! I'm a first-time contributor so I "need a maintainer to approve running workflows."

galal-hussein (approved):

PR #3530, opened 2022-11-05
Take nodeIP into account to configure the calico networks

manuelbuil:

#### Proposed Changes ####

This PR reads the nodeIP parameters and finds the interface with that IP. Then, uses that interface as the vxlan adapater to create the vxlan network that calico uses for inter-node communication.

Without this PR, Calico chooses the interface randomly. If there are multiple interfaces, it might choose the wrong interface and thus inter-node communication will not work. (As reported in https://jira.suse.com/browse/SURE-5486?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel)

#### Types of Changes ####

Bugfix

#### Verification ####

1 - Deploy rke2 in linux-windows. The windows agent must have more than 1 interface
2 - In the windows deployment, list the hns-networks ==> Get-HnsNetwork
3 - Both External and Calico networks should have "ManagementIP = node-ip"

#### Linked Issues ####

https://github.com/rancher/rke2/issues/3533

### DIFF
--- a/pkg/windows/calico.go
+++ b/pkg/windows/calico.go
@@ -277,12 +277,23 @@ func (c *Calico) Start(ctx context.Context) error {
 	return nil
 }
 
+// generateCalicoNetworks creates the overlay networks for internode networking
 func (c *Calico) generateCalicoNetworks() error {
 	if err := deleteAllNetworksOnNodeRestart(); err != nil {
 		return err
 	}
 
-	mgmt, err := createHnsNetwork(c.CNICfg.Mode, os.Getenv("VXLAN_ADAPTER"))
+	vxlanAdapter := os.Getenv("VXLAN_ADAPTER")
+
+	if vxlanAdapter == "" && c.CNICfg.IP != "" {
+		iFace, err := findInterface(c.CNICfg.IP)
+		if err != nil {
+			return err
+		}
+		vxlanAdapter = iFace
+	}
+
+	mgmt, err := createHnsNetwork(c.CNICfg.Mode, vxlanAdapter)
 	if err != nil {
 		return err
 	}
@@ -293,7 +304,6 @@ func (c *Calico) generateCalicoNetworks() error {
 			return err
 		}
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


brandond (approved):

manuelbuil:
> How does this interact with the existing interface selection code?
> 
> https://github.com/rancher/rke2/blob/d6a8d34d6623a81a319021b328a051aaaf3a795b/pkg/windows/calico.go#L262-L286
> 
> Does this replace that, or just allow someone to override it?
> 
> Is there any way we could be more intelligent about interface selection without requiring users to hardcode the node-ip?

IMO, that part of the code is not useful because as soon as it finds one interface with an address which is not 127.0.0.1 or 169.254.x.x, returns true. It does not really select any interface, as it lists all of them. I think this code requires a bit of refactoring that I wanted to do after code freeze with less pressure

TBH, I am starting to think that the Windows pick of the interface when creating the external network is not random but picks the interface with the default gateway. I'm trying to find docs about this but so far unsuccessfully. If this was the case, node-ip would only be needed if you want an interface which is not the one with the default gateway. That's the same case in linux

manuelbuil:
> How does this interact with the existing interface selection code?
> 
> https://github.com/rancher/rke2/blob/d6a8d34d6623a81a319021b328a051aaaf3a795b/pkg/windows/calico.go#L262-L286
> 
> Does this replace that, or just allow someone to override it?
> 
> Is there any way we could be more intelligent about interface selection without requiring users to hardcode the node-ip?

Oh! I now see that I'm currently forcing users to define the node-ip always. I'll getInterface only when nodeIP != ""

pkg/windows/calico.go
```diff
@@ -173,7 +172,7 @@ func getDefaultConfig(config *CNIConfig, dataDir string, nodeConfig *config.Node
 		Platform:              platformType,
 		StartUpValidIPTimeout: 90,
 		LogDir:                "",
-		IP:                    "autodetect",
+		IP:                    nodeConfig.AgentConfig.NodeIP,
```
brandond, line 175: we're unconditionally hardcoding node IP here? If this is unset by the user, should it default back to autodetect?

tests/e2e/mixedos/Vagrantfile
```diff
@@ -98,13 +98,16 @@ def provision(vm, role, role_num, node_num)
     # end
     # If using libvirt, use virt-viewer for GUI after bring up the node
     vm.provision :rke2, run: 'once' do |rke2|
-      rke2.env = [install_type]
+      #rke2.env = [install_type]
+      rke2.env = %W[Commit=c351c1dce61b6952265aae8e78e7fe2ed452a343]
```
brandond, line 102: Is this meant to be merged?

manuelbuil: Crap! This is how I was testing this PR. This is wrong

brandond (approved):

manuelbuil:
Just tested connectivity and works!

rbrtbnfgl (approved):

PR #3543, opened 2022-11-09
Refactor Windows Calico code

manuelbuil:
NO MERGE. CODE FREEZE!

#### Proposed Changes ####

Refactors windows code so that it is more readable and fixes several bugs. This is a high level summary:

== calico.go changes ==
1 - Include the CNI Config inside the Calico struct. That way we don't have to carry it around and pass it to almost all functions
2 - Same for DataDir
3 - Simplify Setup function:
   3.1 - Everything related to initial config, (mainly defaults and calicoKubeconfig parameters) goes into "func initializeConfig"
   3.2 - Then, we check rke2-chart and the possible overriding config in "func overrideCalicoConfigByHelm"
   3.3 - Then, as config is ready, we write the config files required by Calico in "func writeConfigFiles"
   3.4 - Finally, we create the calico network which connects the nodes
4 - Refactor deleteAllNetworksOnNodeRestart and remove the useless backend map and thus the backend arg
5 - Change "createExternalNetwork" name by "createHnsNetwork" to be coherent with CalicoHnsNetworkName const
6 - Remove the potential infinite for loop covering the whole logic when creating the hns network. Replace it with:
```
	// Check if network exists. If it does not after 5 minutes, fail
	for start := time.Now(); time.Since(start) < 5*time.Minute; {
		network, err := hcsshim.GetHNSNetworkByName(CalicoHnsNetworkName)
		if err == nil {
			return network.ManagementIP, nil
		}
	}
```
that will fail if the network was not created in 5 minutes
7 - Remove shallow functions:
     7.1 - "func coreClient" (one liner. Now part of "func createCalicoKubeConfig" called by "func initializeConfig")
     7.2 - Combine "func createKubeConfig", "func CreateCNIConfig", "func initializeConfig" into "func writeConfigFiles"
     7.3 - "func checkIfNetworkExists" (one liner. It is the code doing the 5 minutes loop I described in point 6)
     7.4 - "func waitForManagementIP" (The 5 minutes loop now returns the network.ManagementIP. There is no need to wait again)
     7.5 - "func checkForCorrectInterface()" (I think we can assume there is going to be one interface)
8 - Remove:
```
	"github.com/k3s-io/k3s/pkg/daemons/agent"
	"github.com/k3s-io/k3s/pkg/daemons/config"
```
agent was only used for `agent.NetworkName = CalicoHnsNetworkName` which AFAIK does not make any sense. 
config is imported twice. The second one with the name "daemonconfig". We keep that one
9 - Remove util2 from the import "util2 github.com/k3s-io/k3s/pkg/agent/util". There is no other util being used

== types.go changes ==

1 - the import "github.com/k3s-io/k3s/pkg/daemons/config" is duplicated again
2 - CNIConfig is now part of the Calico struct. We don't need to return it in Setup and carry it around. We can get it by calling GetConfig()
3 - Setup does not require a string arg. It was used to pass the  to pass the DataDir
4 - I don't it makes sense that the CNIConfig includes `NodeConfig *config.Node`. I removed it

== pebinary.go changes ==
1 - Now that Calico struct includes the CNIConfig, PEBinaryConfig struct does not need `cniConig *win.CNIConfig`. We can get the config with cni.GetConfig() if we need it
2 - Include KubeConfigKubeProxy in PEBinaryConfig struct. Initialize it in "func Bootstrap" with:
```
p.KubeConfigKubeProxy = nodeConfig.AgentConfig.KubeConfigKubeProxy
```
This was the only reason why CNIConfig included NodeConfig *config.Node. Adding it to  PEBinaryConfig struct makes more sense.
3 - p.cni.Setup(...) and p.cni.Start(...) do not require the CNIConfig anymore as it is part of the Calico struct
4 - Use p.cni.GetConfig() to get the required fields by KubeProxy

== templates.go ==
Remove it

#### Types of Changes ####

Code refactoring


pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
-		},
+	if err := c.writeConfigFiles(nodeConfig.AgentConfig.CNIConfDir, nodeConfig.AgentConfig.NodeName); err != nil {
+		return err
 	}
 
-	token, err := serviceAccounts.CreateToken(ctx, calicoNode, &req, metav1.CreateOptions{})
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to create token for service account (%s/%s)", CalicoSystemNamespace, calicoNode)
+	logrus.Info("Generating HNS networks, please wait")
+	if err := c.generateCalicoNetworks(); err != nil {
```
briandowns, line 70:
```suggestion
	return c.generateCalicoNetworks(); err != nil {
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+		return err
 	}
-	calicoKubeConfig.Token = token.Status.Token
 
-	cfg := &CNIConfig{NodeConfig: nodeConfig, NetworkName: "Calico", BindAddress: nodeConfig.AgentConfig.NodeIP}
+	return nil
+}
 
-	hc, err := helm.NewFactoryFromConfig(restConfig)
+func (c *Calico) GetConfig() *CNIConfig {
```
briandowns, line 77: It doesn't appear as though this getter is necessary since the field being returned is exported.

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+	}
+	
+	// create calico kubeconfig, calico config and node name files
+	if err = util.WriteFile(filepath.Join("c:\\", c.DataDir, "agent", CalicoKubeConfigName), templates[calicoKubeConfigTemplate]); err != nil {
+		return err
+	}
+	if err = util.WriteFile(filepath.Join(CNIConfDir, CalicoConfigName), templates[calicoConfigTemplate]); err != nil {
+		return err
+	}
+    if err = util.WriteFile(c.CalicoCNICfg.CalicoConfig.KubeConfig.Path, NodeName); err != nil {
```
briandowns, line 145:
```suggestion
    return util.WriteFile(c.CalicoCNICfg.CalicoConfig.KubeConfig.Path, NodeName);
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
-		return nil, err
+// writeConfigFiles writes the three required files by Calico
+func (c *Calico) writeConfigFiles(CNIConfDir string, NodeName string) error {
+	// parse the two config files required by Calico
+	templates, err := parseCalicoTemplatesFromConfig(c.CalicoCNICfg)
+	if err != nil {
+		return err
+	}
+	
+	// create calico kubeconfig, calico config and node name files
```
briandowns, line 138: This err value should be scope locked and not use the err value from above. ie, add a ":" to this equal sign and the ones below.

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
 	}
-	config.CalicoConfig = calicoCfg
+
 	return nil
 }
 
+func getNodeAddressAutodetection(autoDetect opv1.NodeAddressAutodetection) string {
+	if autoDetect.FirstFound != nil && *autoDetect.FirstFound {
+		return "first-found"
+	} else if autoDetect.CanReach != "" {
```
briandowns, line 263: Can all of the same logic be had wihtout all of the "else" statements?

pkg/windows/calico.go
```diff
@@ -285,106 +358,55 @@ func checkForCorrectInterface() (bool, error) {
...
+	// Check if network exists. If it does not after 5 minutes, fail
+	for start := time.Now(); time.Since(start) < 5*time.Minute; {
+		network, err := hcsshim.GetHNSNetworkByName(CalicoHnsNetworkName)
+		if err == nil {
+			return network.ManagementIP, nil
 		}
 	}
-	return nil
+
+	return "", fmt.Errorf("Failed to create %s network", CalicoHnsNetworkName)
```
briandowns, line 409: Error strings shouldn't contain punctuation.

```suggestion
	return "", fmt.Errorf("failed to create %s network", CalicoHnsNetworkName)
```

pkg/windows/templates.go
```diff
@@ -85,18 +85,22 @@ users:
...
-			return strings.ReplaceAll(s, "\\", "\\\\")
-		},
-	}
-	t := template.Must(template.New("compiled_template").Funcs(funcs).Parse(templateBuffer))
-	if err := t.Execute(out, config); err != nil {
-		return "", err
-	}
-	return out.String(), nil
+// parseCalicoTemplatesFromConfig returns parsed configs for Calico in windows
+func parseCalicoTemplatesFromConfig(config *CNIConfig) (map[string]string, error) {
```
briandowns, line 89: This code needs to be fmt'd.

pkg/windows/calico.go
```diff
@@ -39,7 +37,10 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-type Calico struct{}
+type Calico struct{
+	CalicoCNICfg *CNIConfig
```
briandowns, line 41: We should try to avoid stutters in naming.
```suggestion
        CNICfg *CNIConfig
```

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -41,18 +41,18 @@ var (
...
-	ImagesDir       string
-	Resolver        *images.Resolver
-	CloudProvider   *CloudProviderConfig
-	CISMode         bool
-	DataDir         string
-	AuditPolicyFile string
-	KubeletPath     string
-	DisableETCD     bool
-	IsServer        bool
-	cniConig        *win.CNIConfig
```
brandond, line 54: image

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+
+
+// createCalicoKubeConfig creates all needed for Calico to contact kube-api
+func (c *Calico) createCalicoKubeConfig(ctx context.Context, restConfig *rest.Config) (CalicoKubeConfig, error) {
+
+	// Fill all information except for the token
+	calicoKubeConfig := CalicoKubeConfig{
+		Server:               "https://127.0.0.1:6443",
+		CertificateAuthority: filepath.Join("c:\\", c.DataDir, "agent", "server-ca.crt"),
+		Token:                "",
```
brandond, line 160: No need to set a field to its default value
```suggestion
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
 	}
-	if err := createKubeConfig(cfg, dataDir); err != nil {
-		return nil, err
+
+	return nil
+}
+
+
+// createCalicoKubeConfig creates all needed for Calico to contact kube-api
+func (c *Calico) createCalicoKubeConfig(ctx context.Context, restConfig *rest.Config) (CalicoKubeConfig, error) {
```
brandond, line 154: Should this perhaps return:
```suggestion
func (c *Calico) createCalicoKubeConfig(ctx context.Context, restConfig *rest.Config) (*CalicoKubeConfig, error) {
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+		},
 	}
 
-	logrus.Info("Generating HNS networks, please wait")
-	if err := generateCalicoNetworks(cfg.CalicoConfig.NetworkingBackend); err != nil {
-		return nil, err
+	// Register the token in the Calico service account
+	client, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return calicoKubeConfig, err
```
brandond, line 175: If we change the function sig, we can just return nil instead of an unusable kubeconfig
```suggestion
		return nil, err
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+	}
+	serviceAccounts := client.CoreV1().ServiceAccounts(CalicoSystemNamespace)
+	token, err := serviceAccounts.CreateToken(ctx, calicoNode, &req, metav1.CreateOptions{})
+	if err != nil {
+		return calicoKubeConfig, errors.Wrapf(err, "failed to create token for service account (%s/%s)", CalicoSystemNamespace, calicoNode)
```
brandond, line 180: same comment here - return nil pointer instead of unusable struct
```suggestion
		return nil, errors.Wrapf(err, "failed to create token for service account (%s/%s)", CalicoSystemNamespace, calicoNode)
```

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
 	}
 
-	return cfg, nil
+	calicoKubeConfig.Token = token.Status.Token
+
+
+	return calicoKubeConfig, nil
```
brandond, line 186:
```suggestion
	return &calicoKubeConfig, nil
```

pkg/windows/calico.go
brandond, line 145: the indentation here appears to be off? I'm also confused what's going on, we're writing a file containing the NodeName where the calico kubeconfig should be?

pkg/windows/calico.go
```diff
@@ -52,147 +53,226 @@ const (
...
+// writeConfigFiles writes the three required files by Calico
+func (c *Calico) writeConfigFiles(CNIConfDir string, NodeName string) error {
+	// parse the two config files required by Calico
+	templates, err := parseCalicoTemplatesFromConfig(c.CalicoCNICfg)
+	if err != nil {
+		return err
+	}
+	
+	// create calico kubeconfig, calico config and node name files
+	if err = util.WriteFile(filepath.Join("c:\\", c.DataDir, "agent", CalicoKubeConfigName), templates[calicoKubeConfigTemplate]); err != nil {
```
brandond, line 139: We have a similar but different version of this at https://github.com/k3s-io/k3s/blob/master/pkg/daemons/control/deps/deps.go#L75 - I wonder if we could replace both of of these with something more streamlined that uses proper Kubernetes types and serialization instead of just rendering a string template?

brandond (changes requested):

pkg/windows/calico.go
manuelbuil: This is the reasoning: pebinary is using the CNI interface which defines functions "Start" and "Setup". Calico struct complies with the CNI interface and it includes the CalicoCNIConfig. However, as golang interfaces only define functions, pebinary has no way to access CalicoCNIConfig (I think). That's why I created a third function (GetConfig() ) in the CNI interface to retrieve the config.

pkg/windows/calico.go
manuelbuil: Good idea!

pkg/windows/calico.go
manuelbuil: good idea too!

pkg/windows/calico.go
manuelbuil: In my defense, I must say that this is old code ;). Do you mean adding separate if conditions? As it evaluates different subfields of `autoDetect` we can't have a `switch` :(

pkg/windows/calico.go
manuelbuil: good to know

pkg/windows/templates.go
manuelbuil: done

pkg/windows/calico.go
manuelbuil: good, done!

pkg/pebinaryexecutor/pebinary.go
manuelbuil: :shrug:

pkg/windows/calico.go
manuelbuil: very true! Done

pkg/windows/calico.go
manuelbuil: Yes! Pointers improve efficiency

pkg/windows/calico.go
manuelbuil: yes

pkg/windows/calico.go
manuelbuil: yes

pkg/windows/calico.go
manuelbuil: correct

pkg/windows/calico.go
manuelbuil: For some reason, VisualStudio does not digest well files starting with the lines:
```
//go:build windows
// +build windows
```
so the automatic fmt I always get is not working anymore. I `gofmt` it manually and it should be fixed now.

It seems it is reusing the path of Kubeconfig.... but why it does not use that when creating the kubeconfig? I need to change this and maybe use a more general path name

pkg/windows/calico.go
manuelbuil: let me have a look and try to replicate the same

pkg/windows/calico.go
brandond: yeah, I have the same problem with vim - unless I override the GOARCH when starting vim it won't autoformat or autocomplete in the windows-tagged files.

pkg/windows/calico.go
brandond: Does the `cni` field have to be of type `win.CNI`, or could we just make it of type `Calico`? As long as `Calico` implements the `win.CNI` interface we can refer to it as the actual type within this module.

manuelbuil: @brandond you mean adding calico as value for cni field when initializing pebinaryexecutor? https://github.com/rancher/rke2/blob/master/pkg/rke2/rke2_windows.go#L61-L72. I guess we could do that by using 
```
if clx.String("cni") == calico {
    cni = windows.Calico
}
```
And then:
```
	return &pebinaryexecutor.PEBinaryConfig{
		Resolver:        resolver,
		ImagesDir:       agentImagesDir,
		ManifestsDir:    agentManifestsDir,
		CISMode:         isCISMode(clx),
		CloudProvider:   cpConfig,
		DataDir:         dataDir,
		AuditPolicyFile: clx.String("audit-policy-file"),
		KubeletPath:     cfg.KubeletPath,
		DisableETCD:     clx.Bool("disable-etcd"),
		IsServer:        isServer,
		cni:                cni,
	}, nil
```
And then we need to import "github.com/rancher/rke2/windows/". Is that what you mean?

manuelbuil: On reflection, I guess you mean adding Calico struct directly as the value of cni in `type PEBinaryConfig struct {`. We can do that but in that case, I think we should remove: `type CNI interface {` because we don't use it anymore. TBH, as it does not seem that we will add any other CNI plugin in the foreseeable future, I think it is fine

briandowns: Let's remove it and keep it explicit for now. Generally, there's never a time to add an abstraction when there's only one implementation.

brandond: @briandowns you're suggesting we remove the windows-specific `CNI` interface that only has the single implementation? If so I am +1 on that.

brandond (approved):

pkg/windows/calico.go
manuelbuil: done

manuelbuil:
@virtualmo could you please test the last commit? I test and it seems to work but just in case :)

pkg/windows/calico.go
```diff
@@ -7,23 +7,20 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"html/template"
```
briandowns, line 10: Does "html/template" provide you specific html functionality you're using? If not, it'd be preferred to use "text/template".

pkg/windows/calico.go
```diff
@@ -52,147 +137,235 @@ const (
...
 	}
 
-	client, err := coreClient(restConfig)
+	logrus.Info("Generating HNS networks, please wait")
+	return c.generateCalicoNetworks()
+}
+
+// initializeConfig sets the default configuration in CNIConfig
+func (c *Calico) initializeConfig(ctx context.Context, nodeConfig *daemonconfig.Node, restConfig *rest.Config) error {
+	c.CNICfg = &CalicoConfig{}
```
briandowns, line 161: It would appear as though this line isn't necessary since you're doing that and more on line 107,

pkg/windows/calico.go
```diff
@@ -52,147 +137,235 @@ const (
...
-	cfg := &CNIConfig{NodeConfig: nodeConfig, NetworkName: "Calico", BindAddress: nodeConfig.AgentConfig.NodeIP}
+	return nil
+}
 
-	hc, err := helm.NewFactoryFromConfig(restConfig)
+// writeConfigFiles writes the three required files by Calico
+func (c *Calico) writeConfigFiles(CNIConfDir string, NodeName string) error {
+
+	// Create CalicoKubeConfig and CIPAutoDetectionMethodalicoConfig files
+	err := c.renderCalicoConfig(c.CNICfg.KubeConfig.Path, calicoKubeConfigTemplate)
```
briandowns, line 209: Single err returns should remain scope locked on a single line.

pkg/windows/calico.go
```diff
@@ -52,147 +137,235 @@ const (
...
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	if err := getDefaultConfig(cfg, dataDir, nodeConfig); err != nil {
-		return nil, err
+	err = c.renderCalicoConfig(filepath.Join(CNIConfDir, CalicoConfigName), calicoConfigTemplate)
```
briandowns, line 214: Single line here to scope lock err. Avoids problems with value reuse.

pkg/windows/calico.go
```diff
@@ -52,147 +137,235 @@ const (
...
+
+	if autoDetect.SkipInterface != "" {
+		return "skip-interface=" + autoDetect.SkipInterface, nil
+	}
+
+	if len(autoDetect.CIDRS) > 0 {
+		return "cidr=" + strings.Join(autoDetect.CIDRS, ","), nil
+	}
+
+	return "", fmt.Errorf("the passed autoDetect value is not supported. Please read Calico docs")
```
briandowns, line 366:
```suggestion
	return "", errors.New("the passed autoDetect value is not supported. Please read Calico docs")
```

pkg/windows/calico.go
manuelbuil: Not really. I was just reusing that import (moved the templating stuff to calico.go). Thanks, I'll change that

pkg/windows/calico.go
manuelbuil: correct, good catch!

pkg/windows/calico.go
manuelbuil: Correct again, I missed this

pkg/windows/calico.go
manuelbuil: yes, good catch (with the lacrosse stick)

pkg/windows/calico.go
manuelbuil: good idea. Another left over :/

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -173,12 +155,9 @@ func (p *PEBinaryConfig) Kubelet(ctx context.Context, args []string) error {
 // KubeProxy starts the kubeproxy in a subprocess with watching goroutine.
 func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
 	extraArgs := map[string]string{
-		"hostname-override": p.cniConig.CalicoConfig.Hostname,
-		"v":                 "4",
-		"proxy-mode":        "kernelspace",
-		"kubeconfig":        p.cniConig.NodeConfig.AgentConfig.KubeConfigKubeProxy,
-		"network-name":      p.cniConig.NetworkName,
-		"bind-address":      p.cniConig.BindAddress,
+		"v":            "4",
```
brandond, line 158: We don't normally override the log level for components, is there some particular reason to ship this with a higher-than-normal log verbosity?

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -187,15 +166,18 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
...
-		endpoint, err := hcsshim.GetHNSEndpointByName("Calico_ep")
-		if err != nil {
-			logrus.WithError(err).Warningf("can't find %s, retrying", "Calico_ep")
-			continue
+	if p.cni.CNICfg.Name == "Calico" {
+		var vip string
+		for range time.Tick(time.Second * 5) {
+			endpoint, err := hcsshim.GetHNSEndpointByName("Calico_ep")
+			if err != nil {
+				logrus.WithError(err).Warningf("can't find %s, retrying", "Calico_ep")
```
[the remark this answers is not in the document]
manuelbuil: yep, good idea

manuelbuil:
Waiting on https://github.com/rancher/rke2/pull/3577 to merge

pkg/windows/calico.go
```diff
@@ -234,157 +404,73 @@ func startCalico(ctx context.Context, config *CalicoConfig) error {
...
+					AddressPrefix:  "192.168.255.0/30",
+					GatewayAddress: "192.168.255.1",
 				},
-			}
+			},
 		}
+	}
+
+	if _, err := network.Create(); err != nil {
+		return "", fmt.Errorf("Error creating the %s network: %v", CalicoHnsNetworkName, err)
```
briandowns, line 462:
```suggestion
		return "", fmt.Errorf("error creating the %s network: %v", CalicoHnsNetworkName, err)
```

pkg/windows/calico.go
```diff
@@ -234,157 +404,73 @@ func startCalico(ctx context.Context, config *CalicoConfig) error {
...
-						Policies: []json.RawMessage{
-							[]byte("{ \"Type\": \"VSID\", \"VSID\": 9999 }"),
-						},
+// createHnsNetwork creates the network that will connect nodes and returns its managementIP
+func createHnsNetwork(backend string, vxlanAdapter string) (string, error) {
+	var network hcsshim.HNSNetwork
+	if backend == "vxlan" {
+		// Ignoring the return because both true and false without an error represent that the firewall rule was created or already exists
+		if _, err := wapi.FirewallRuleAdd("OverlayTraffic4789UDP", "Overlay network traffic UDP", "", "4789", wapi.NET_FW_IP_PROTOCOL_UDP, wapi.NET_FW_PROFILE2_ALL); err != nil {
+			return "", fmt.Errorf("Error creating firewall rules: %v", err)
```
briandowns, line 431:
```suggestion
			return "", fmt.Errorf("error creating firewall rules: %v", err)
```

pkg/windows/calico.go
```diff
@@ -52,147 +137,232 @@ const (
...
-func getDefaultConfig(config *CNIConfig, dataDir string, nodeConfig *config.Node) error {
-	platformType, err := getPlatformType()
+	cniChartConfig, err := hc.Helm().V1().HelmChartConfig().Get(metav1.NamespaceSystem, CalicoChart, metav1.GetOptions{})
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to check for %s HelmChartConfig", CalicoChart)
+	}
+	if cniChartConfig == nil {
```
briandowns, line 318: Should we log here to know why this may have returned nil?

pkg/windows/calico.go
```diff
@@ -52,147 +137,232 @@ const (
...
+
+	if c.CNICfg.Platform == "ec2" || c.CNICfg.Platform == "gce" {
+		logrus.Debugf("recreating metadata route because platform is: %s", c.CNICfg.Platform)
+		if err := setMetaDataServerRoute(mgmt); err != nil {
+			return err
+		}
+	}
+	if c.CNICfg.Mode == "windows-bgp" {
+		// Not supported yet, this does work.
+		_ = wapi.StopService("RemoteAccess")
```
briandowns, line 298: Should these be commented out? Are there side effects?

pkg/windows/calico.go
manuelbuil: thanks

pkg/windows/calico.go
manuelbuil: thanks

pkg/windows/calico.go
manuelbuil: good idea, let me add a log when being in debug mode

pkg/windows/calico.go
manuelbuil: Good catch, we are right now hardconding the mode to be vxlan, so this code can be removed

brandond (approved):

PR #3570, opened 2022-11-18
Add ADR for security bumps automation

macedogm:
#### Proposed Changes ####

This PR adds an ADR to propose the automation of versions and security bump of dependencies.

#### Further Comments ####

Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

### DIFF
--- a/docs/adrs/005-security-updates-automation.md
+++ b/docs/adrs/005-security-updates-automation.md
@@ -1,0 +1,80 @@
+# Security updates and versions bump automation
+
+Date: 2022-11-18
+
+## Status
+
+Accepted
+
+## Context
+
+The current process of keeping third-party dependencies and container images up to date is manual and time consuming, which makes us lack behind upstream updates. This leads to RKE2 sometimes shipping outdated dependencies and images that can introduce security issues (CVEs) in our code and in users environments. A process is needed to automate the discovery of updates and PRs, so developers don't need to spend valuable time with manual tasks that can be easily automate.
+
+The basic requirements that we need for such automation regarding the tooling are:
+
+1. The tool is free and open-source.
+2. Supports major packages ecosystems - Docker, Go, Python etc.
+3. Automatically opens pull requests (PRs).
+4. Preferably supports bumping versions in unstructured formats, e.g., `ENV` vars in Dockerfiles and by matching regular expressions.
+5. Preferably supports scripting the automation process, e.g., update package A in repo B after package X in repo Y matches a pre-defined version criteria.
+
+There are well known free and/or open-source tools available for this kind of automation:
+
+### 1. [Dependabot](https://docs.github.com/en/code-security/dependabot)
+
+- Provides PR automation for version and security updates.
+- Is provided as a service by GitHub and free for public and private repos.
+  - There is no need to add a GH app or token in the repo.
+- Simple to configure in the repo settings in GitHub and with an YAML configuration `.github/dependabot.yml` file.
+- Supports bumping of images in Dockerfiles and major packages ecosystem. See full list [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem).
+- Doesn't support:
+  - Bumping versions in unstructured formats.
+  - Orchestrating updates.
+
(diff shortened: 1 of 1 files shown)


brandond (approved):

dereknola (approved):

briandowns (approved):
Dig it.

cwayne18 (approved):
love it, thank you

macedogm:
@cwayne18 should we hold the approval/merge of this ADR in order for it to reflect the K3s one - https://github.com/k3s-io/k3s/pull/6559#issuecomment-1334498599?

nikolaishields (approved):

matttrach:
These tools simply generate PRs, in that way they are not destructive and adding automation to gather the data into a single PR is a pretty good improvement to our current process. I am excited to see that improvement and how we can build on it.

We need to be careful not to merge the PRs blindly.
Each PR will need to be inspected in a very similar manner to how we approach releases now.
Compatibility with underlying systems is a complex topic, for instance, the containerd version on the default version of an operating system we support might effect what version of cilium we choose, using the latest version of everything might not work for our entire support matrix.

We should be careful how much we depend on updateCLI.
updateCLI only compares the versions from a source and makes a PR to a target, it doesn't inspect if the source's artifacts are present much less working as expected. That being said, we could certainly spend time building plugins for updateCLI if that is what we want to do.

Dependabot is much better, it has its own CVE database which it compares against a deep scan of the dependencies involved. I would consider Dependabot a standard in security scanning system systems.

I hope these comments add context to the decision, which I approve of, and that we consider them as we implement the tools mentioned.

matttrach (approved):

macedogm:
> We need to be careful not to merge the PRs blindly.

That is right. Reviewing and merging is still a manual task. This is mentioned as outside the scope of the automation in the related PR for K3s. See in https://github.com/k3s-io/k3s/pull/6583/files#diff-7034c9aa94c5412bc39399ff885593b1bf25fe91c0bab79f4923fe2c7d61ae16R22-R23.

> We need to be careful not to merge the PRs blindly.
Each PR will need to be inspected in a very similar manner to how we approach releases now.

IMO this automation doesn't change the current process, because of:
1. We are not doing auto-merge; and
2. The automation is initially only targeting safe updates, like of projects that we control, for example, Helm, Klipper etc. In the K3s PR we demoed updating upstream projects, like Sonobuoy, but we don't need to target them at this phase. 

> Compatibility with underlying systems is a complex topic, for instance, the containerd

Such updates we should let Renovate/Dependabot and similar tools handle, because they already have builtin logic to check for compatibility (which is not bulletproof, but it's there).

> updateCLI only compares the versions from a source and makes a PR to a target, it doesn't inspect if the source's artifacts are present much less working as expected

It does check if the the artifacts are present and available, see the `sources` and `conditions` sections. How much we want to check and double check is up to us when implementing each update pipeline.

> Dependabot is much better, it has its own CVE database which it compares against a deep scan of the dependencies involved. I would consider Dependabot a standard in security scanning system systems.

That is not entirely true and Dependabot is not a `security scanning system systems`. Just to reiterated as explained in the ADR, Dependabot is good at [targeting updates in packages ecosystems](https://github.com/rancher/rke2/pull/3570/files#diff-0e1de499783aef66ff8c087bcd8263f84a03153503fe42ed5ff08f823603d9f5R61), Renovate goes a bit further and allows updates in [unstructured formats](https://github.com/rancher/rke2/pull/3570/files#diff-0e1de499783aef66ff8c087bcd8263f84a03153503fe42ed5ff08f823603d9f5R63) like ENV vars, and Updatecli is good at allowing the orchestration of when and how to do the updates in basically any kind of file and unstructured format. It can also be used in other use cases.

macedogm:
@matttrach please let me know if I was able to address all of your concerns above ^.

macedogm:
Note: moving this PR to draft, because I want to do some updates in the ADR before it gets merged.

macedogm:
After chatting offline with @matttrach, I'm moving this PR to ready, since we are getting more mature and confident with using Updatecli (see https://github.com/k3s-io/k3s/issues/7040 for the related updates in K3s). If Dependabot gets too noisy, we can also test Renovate ([rancher/renovate-config](https://github.com/rancher/renovate-config)).

PR #3608, opened 2022-11-30
Add rke2 standalone install script for Windows

mdrahman-suse:
#### Proposed Changes ####

This PR includes a script that automates installing RKE2 in Windows instances to reduce the manual effort needed for setup and installation. 

That can be decided based on feedback from the team

#### Types of Changes ####

No changes to the RKE2 codebase 

#### Verification ####

For testing,
- Setup a server node and install rke2
- Setup an Windows instance
- Download the script
- Run the script by providing serverIP, token and rke2 version

#### Linked Issues ####

https://github.com/rancher/qa-tasks/issues/151

### DIFF
--- a/windows/rke2-quickstart.ps1
+++ b/windows/rke2-quickstart.ps1
@@ -1,0 +1,104 @@
+<#
+.SYNOPSIS
+  A quickstart script to Setup and Install standalone RKE2 in Windows to be used as Worker Nodes.
+  This script enables features, sets up environment variables and adds default configuration that are needed to install rke2 in Windows and join a cluster
+.DESCRIPTION
+  Run the script to setup and install all RKE2 related needs and to join a cluster.
+.Parameter ServerIP
+    Server IP of Primary server where rke2 is already installed and the worker will join.`
+.Parameter Token
+    Token of Primary server.`
+.Parameter Version
+    Version of rke2 to download from github.`
+.Parameter Commit
+    Commit of RKE2 to download from temporary cloud storage.
+    If set, this forces Method=tar in install script.
+    * (for developer & QA use only)`
+.EXAMPLE
+  Usage:
+    Invoke-WebRequest ((New-Object System.Net.WebClient).DownloadString('https://github.com/rancher/rke2/blob/master/windows/rke2-quickstart.ps1'))
+    ./rke2-quickstart.ps1 $ServerIP <server-IP> $Token <server-token> $Version <rke2-version>
+#>
+
+[CmdletBinding()]
+param (
+    [Parameter()]
+    [String]
+    $ServerIP,
+    [Parameter()]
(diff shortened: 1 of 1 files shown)


brandond:
Can we maybe call this a "quickstart" script (and name it similarly) to avoid any confusion around the intent and purpose of this script? We already have an install script for Windows, but the RKE2 installers for both Linux and Windows intentionally don’t handle configuration or start the service. That will probably make documenting it easier as well.

windows/rke2-standalone-install.ps1
```diff
@@ -0,0 +1,103 @@
...
+    }
+    elseif ($Version) {
+        $Version = $Version
+    }
+    return $Version
+}
+
+function Enable-Features() {
+    Enable-WindowsOptionalFeature -Online -FeatureName Containers -All
+}
```
dereknola, line 63: Doesn't the node need to restart after this is enabled?

brandond: I think there is a -Restart flag that will reboot it if necessary. If this is our YOLO script we might as well add that?

mdrahman-suse: As per the doc, it is required to reboot. But we saw rke2 was installed and started successfully even with no restart. So need to check if that is a hard requirement. I will check on this, thanks!

windows/rke2-quickstart.ps1
```diff
@@ -0,0 +1,105 @@
...
+    }
+    elseif ($Version) {
+        $Version = $Version
+    }
+    return $Version
+}
+
+function Enable-Features() {
+    Enable-WindowsOptionalFeature -Online -FeatureName Containers -All
+    Restart-Computer -Wait -For PowerShell -Timeout 300 -Delay 2
```
brandond, line 64: It looks like prompting to reboot is actually the default, so unconditionally rebooting probably isn't the right answer.

https://learn.microsoft.com/en-us/powershell/module/dism/enable-windowsoptionalfeature?view=windowsserver2022-ps

mdrahman-suse: So when rke2 joins successfully, the system reboots automatically.. removing this Restart command should be fine then?

mdrahman-suse: I have tested with both with and without this command (single node and HA) and was successfully able to join the windows node to the cluster

rancher-max (approved):

windows/rke2-quickstart.ps1
mdrahman-suse: Removed this Restart command. Will address any issues if observed on a separate PR

brandond (approved):

mdrahman-suse:
Merging in 5 mins if no concerns

PR #3611, opened 2022-12-01
Fix Jenkinsfile typo and clarify support for oracle in TF automation

rancher-max:
#### Proposed Changes ####

There was a typo in the Jenkinsfile where a letter is missing, so I fixed that. Also included `oracle` as a working node_os option so that we don't have to put `rhel8` every time we use this TF automation for an Oracle machine.


brandond (approved):

ShylajaDevadiga (approved):
LGTM

PR #3661, opened 2022-12-07
December 2022 channels.yaml update

briandowns:


matttrach (approved):

cwayne18 (approved):

brandond (approved):
I don't like that we're now two minors behind, but I guess we're waiting on Rancher to get their act together with 1.25 support...

Issue #3440, opened 2022-10-07
Add pod-security-admission-config-file flag to the doc


PhanLe1010:
We added the new flag `pod-security-admission-config-file` [here](https://github.com/rancher/rke2/blob/ffba111f8f3d9933d9e1f08caa29944352ab4739/pkg/rke2/rke2_linux.go#L119-L125), so maybe considering adding it to [the doc](https://github.com/rancher/rke2/blob/master/docs/install/install_options/server_config.md)?

Context: I was working on testing some Pod Security Admission and tried to modify the default behavior. Looking through the doc doesn't seems to find a way to set it until I read [the PR](https://github.com/rancher/rke2/pull/3282/files#diff-0cdb72b73bef2169e6b872613aefd8949abef5f6b9dceae387ba9a1222487c90).

dereknola:
It exists in the docs https://docs.rke2.io/security/pod_security_standards#pod-security-standards, we were transitioning the Rke2 docs to a new website, so there was about a month where the old docs still existed but were not getting updated.

sureshgoli25:
I did roundtrip to here to know that, it is not yet covered in here https://docs.rke2.io/reference/server_config but available :)

caroline-suse-rancher:
Closing because the documentation is available in an appropriate location

PR #3665, opened 2022-12-08
fix(pkg/rke2): psa custom path

perriea:
#### Proposed Changes ####

In fact, a `pod-security-admission-config-file` flag exists to define an arbitrary path for the PSA. But even if the flag is set, the file is still created at `/etc/rancher/rke2/rke2-pss.yaml` (`defaultPSAConfigFile`). This is incorrect behavior.

#### Types of Changes ####

Bugfix

#### Linked Issues ####

/!\ This flag isn't registered in the documentation #3440


perriea:
Anyone is available to review this ?

brandond:
I believe this was intended behavior - we only write the default config file. If you provide your own config file path, it will be used, but you are responsible for populating it with the correct content. I would defer to @galal-hussein on that though.

brandond:
To be clear @perriea - you object to the fact that RKE2 is writing the default configuration file out to the fixed location, even if you've specified a custom path that already contains your desired configuration?

perriea:
Yes, otherwise I don't understand the existence of this flag which at this moment is useless 😅

erikfuego:
Correct me if i'm wrong @perriea, but I think the point of this flag is for us to provide our own version of the `admission-control-config.yaml` right? Which is why the path stays the same on the node

Issue #855, opened 2021-04-07
rke2-killall.sh does not clean up all network namespaces?


Martin-Weiss:
We had rke2 1.20.5 deployed on sles15sp2 and then did a rke2-killall.sh, deleted the installation and re-deployed (no reboot)

Even after the killall we can see this:

zrz-ux-vm11634:~ # ip netns
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
cni-3b14aef7-1868-f8cc-422c-df5a5c62084c (id: 0)
cni-5f7c07bc-67a0-cb4e-9f10-2e14eb312711 (id: 3)
cni-2b859e51-1be8-7cd8-bb9f-956c77698475 (id: 1)
Error: Peer netns reference is invalid.
cni-b61067e2-ec37-9b02-971c-166c54468ea4
Error: Peer netns reference is invalid.
cni-f4d1dab9-9124-c063-0b1a-7d3a78694ceb
Error: Peer netns reference is invalid.
cni-b0dbfb39-fe4b-9fb7-3f2e-f5416f230d0c

--> could it be that the killall does not clean up properly and is missing something on the cni side of things?

(Workaround was to reboot - but maybe we could get the deletion of the namespaces added to the rke2-killall, too?)

brandond:
Probably just needs the equivalent of https://github.com/k3s-io/k3s/pull/2542

Martin-Weiss:
> 
> 
> Probably just needs the equivalent of [k3s-io/k3s#2542](https://github.com/k3s-io/k3s/pull/2542)

Yeah - looks like.. how can we get this added in RKE2 uninstall?

kmahyyg:
same problem here.

mdrahman-suse:
### Validated on master branch with commit 81b04f085bd73d7a285f63087489600fb011a7a4
  
## Environment Details

  
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
SUSE Linux Enterprise Server 15 SP4
```
  
*Cluster Configuration:*

```
1 server
```
  
*Config.yaml:*

```
write-kubeconfig-mode: 644
node-name: server
debug: true
```
  
## Testing Steps

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/rke2 && sudo cp config.yaml /etc/rancher/rke2
```
2. Install RKE2
[...]

PR #3954, opened 2023-03-02
Remove unmounts in killall script

dereknola:
#### Proposed Changes ####
In the killall script, remove the folders/files that we just unmounted. 

#### Types of Changes ####
Uninstall cleanup

#### Verification ####
Verified that `/var/lib/kubelet/pods` and other mounts are deleted after uninstall.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/855

### DIFF
--- a/bundle/bin/rke2-killall.sh
+++ b/bundle/bin/rke2-killall.sh
@@ -44,7 +44,7 @@ do_unmount_and_remove() {
     if [ -n "${MOUNTS}" ]; then
         set -x
         umount ${MOUNTS}
-        rm -rf ${MOUNTS}
+        rm -rf --one-file-system ${MOUNTS}
     else
         set -x
     fi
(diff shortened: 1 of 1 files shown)


bundle/bin/rke2-killall.sh
```diff
@@ -44,6 +44,7 @@ do_unmount() {
     if [ -n "${MOUNTS}" ]; then
         set -x
         umount ${MOUNTS}
+        rm -rf ${MOUNTS}
```
brandond, line 47: I feel like this either shouldn't be recursive (there should not be anything left under these paths after the filesystems have been unmounted, or we should we use `--one-file-system` just to be extra safe, in case the unmount fails for some reason. I'd hate to accidentally delete a bunch of files out of a volume during a killall.

brandond (approved):

briandowns (approved):

PR #3975, opened 2023-03-08
Remove unnecessary bits from testing dockerfile

rancher-max:
#### Proposed Changes ####

Take away the `go get` bits from the Dockerfile that is used for terraform tests. Having `go get` in this file causes failures due to not strictly adhering to the go.mod, so when Go version is bumped, things break.

#### Types of Changes ####

Bugfix 

#### Verification ####

Run terraform tests

#### Linked Issues ####

N/A

#### Further Comments ####

N/A


VestigeJ (approved):
LGTM

brandond (approved):

PR #4009, opened 2023-03-14
Add Make commands to terraform tests

fmoral2:
#### Proposed Changes ####

- Add Make commands to terraform tests to be able to running automation easier

- Refactor of dockerfile to also accept arm arch

- Add golangci lint and vet on the tests directory

- Change file createcluster to create cluster folder

#### Types of Changes ####

Doc only - README

#### Verification ####

Run the make commands

#### Linked Issues ####

n/a

#### Further Comments ####

- Forgot to squash commit. Sorry! :)

### DIFF
--- a/tests/terraform/README.md
+++ b/tests/terraform/README.md
@@ -37,13 +37,12 @@ Test Flags:
 We can also run tests through the Makefile:
 ```bash
 Args:
-*All args are optional and cab be used with `$make tf-tests-run` and `$make tf-tests-logs`
+*All args are optional and can be used with `$make tf-tests-run` and `$make tf-tests-logs`
 
 - ${IMGNAME}     append any string to the end of image name
-- ${TEST}        call a specific tests directory
 - ${ARGNAME}     name of the arg to pass to the test
 - ${ARGVALUE}    value of the arg to pass to the test
-- ${TESTDIR}     path to the test directory
+- ${TESTDIR}     path to the test directory 
 
 Commands:
 $ make tdf-tests-up   # create the image from Dockerfile.build
@@ -55,9 +54,9 @@ $ make tf-tests       # clean resources + remove images + run tests
 $ make vet-lint       # runs go vet and go lint
 
 Examples:
-$ make tf-tests-run NAME=1 TEST=upgradecluster ARGNAME=upgradeVersion ARGVALUE=v1.24.8+rke2r1
-$ make tf-tests-run TEST=createcluster
-$ make tf-tests-logs NAME=1
+$ make tf-tests-run IMGNAME=2 TESTDIR=terraform/upgradecluster ARGNAME=upgradeVersion ARGVALUE=v1.24.8+rke2r1
+$ make tf-tests-run TESTDIR=terraform/upgradecluster
+$ make tf-tests-logs IMGNAME=1
 ```
 
 
--- a/tests/terraform/README.md
+++ b/tests/terraform/README.md
@@ -17,6 +17,8 @@ See the [create cluster test](../tests/terraform/createcluster_test.go) as an ex
 Before running the tests, it's required to create a tfvars file in `./tests/terraform/modules/config/local.tfvars`. This should be filled in to match the desired variables, including those relevant for your AWS environment. All variables that are necessary can be seen in [main.tf](../tests/terraform/modules/main.tf).
 It is also required to have standard AWS environment variables present: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
 
+** Please also when creating tf var resource_name, make sure that you do not have any instances from other automations with the same name to avoid deleting wrong resources
+
 Tests can be run per package with:
 ```bash
 go test -timeout=30m -v ./tests/terraform/$PACKAGE_NAME/...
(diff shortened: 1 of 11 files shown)


tests/terraform/README.md
```diff
@@ -29,6 +31,25 @@ $ docker run --name rke2-tf-creation-test -t -e AWS_ACCESS_KEY_ID=<YOUR_ACCESS_K
 $ docker run --name rke2-tf-upgrade-test -t -e AWS_ACCESS_KEY_ID=<YOUR_ACCESS_KEY> -e AWS_SECRET_ACCESS_KEY=<YOUR_SECRET_KEY> -v /path/to/aws/key.pem:/tmp/aws_key.pem rke2-tf sh -c "go test -timeout=45m -v ./tests/terraform/upgradecluster/... -upgradeVersion=v1.24.8+rke2r1"
 ```
 
+We can also run tests through the Makefile:
+```bash
+Args:
```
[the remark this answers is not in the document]
fmoral2: nice, will do that . tks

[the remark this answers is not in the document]
fmoral2: nope, they are not for all

tests/terraform/README.md
```diff
@@ -28,6 +30,34 @@ $ docker build . -f ./tests/terraform/scripts/Dockerfile.build -t rke2-tf
...
 $ docker run --name rke2-tf-upgrade-test -t -e AWS_ACCESS_KEY_ID=<YOUR_ACCESS_KEY> -e AWS_SECRET_ACCESS_KEY=<YOUR_SECRET_KEY> -v /path/to/aws/key.pem:/tmp/aws_key.pem rke2-tf sh -c "go test -timeout=45m -v ./tests/terraform/upgradecluster/... -upgradeVersion=v1.24.8+rke2r1"
 ```
+Test Flags:
+```
+- ${upgradeVersion} version to upgrade to
+```
+We can also run tests through the Makefile:
+```bash
+Args:
+*All args are optional and cab be used with `$make tf-tests-run` and `$make tf-tests-logs`
```
mdrahman-suse, line 40: typo `cab`

tests/terraform/scripts/delete_resources.sh
```diff
@@ -0,0 +1,78 @@
...
+  "Name=instance-state-name,Values=running" --query \
+  'Reservations[].Instances[].InstanceId' --output text) > /dev/null 2>&1
+
+#Get the list of load balancer ARNs
+LB_ARN_LIST=$(aws elbv2 describe-load-balancers \
+  --query "LoadBalancers[?starts_with(LoadBalancerName, '${NAME_PREFIX}') && Type=='network'].LoadBalancerArn" \
+  --output text)
+
+#Loop through the load balancer ARNs and delete the load balancers
+for LB_ARN in $LB_ARN_LIST; do
```
mdrahman-suse, line 21: Do we need to add a check for LB_ARN existing in LB_ARN_LIST or will the for loop is sufficient?

tests/terraform/scripts/delete_resources.sh
```diff
@@ -0,0 +1,78 @@
...
+  aws elbv2 delete-load-balancer --load-balancer-arn "$LB_ARN"
+done
+
+#Get the list of target group ARNs
+TG_ARN_LIST=$(aws elbv2 describe-target-groups \
+  --query "TargetGroups[?starts_with(TargetGroupName, '${NAME_PREFIX}') && Protocol=='TCP'].TargetGroupArn" \
+  --output text)
+
+#Loop through the target group ARNs and delete the target groups
+for TG_ARN in $TG_ARN_LIST; do
```
mdrahman-suse, line 32: same question as https://github.com/rancher/rke2/pull/4009/files#r1137878038

tests/terraform/README.md
```diff
@@ -28,6 +30,34 @@ $ docker build . -f ./tests/terraform/scripts/Dockerfile.build -t rke2-tf
...
+
+- ${NAME}     append any string to the end of image name
```
mdrahman-suse, line 42: Can we rename this variable to IMAGE_NAME so that its purpose is clear?

tests/terraform/README.md
```diff
@@ -17,6 +17,8 @@ See the [create cluster test](../tests/terraform/createcluster_test.go) as an ex
 Before running the tests, it's required to create a tfvars file in `./tests/terraform/modules/config/local.tfvars`. This should be filled in to match the desired variables, including those relevant for your AWS environment. All variables that are necessary can be seen in [main.tf](../tests/terraform/modules/main.tf).
 It is also required to have standard AWS environment variables present: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
 
+** Please also when creating tf var resource_name, make sure that you do not have any instances from other automations with the same name to avoid deleting wrong resources
```
mdrahman-suse, line 20: Can this check be automated as test pre-run?

fmoral2: we totally can...but the describe instances command that we should use for that will increase more runtime seconds and this script is already kind of slow because it deletes resources from different places like: load balancers , tgs, r53 and ec2. 

Do you think this tradeoff will worth it ? please let me know you thoughts

mdrahman-suse: I was thinking about just failing the test if any resource with that name exists so that the user can just delete them

fmoral2: @mdrahman-suse  OH ok , so: 

- For resources that cant be created with the same name,  terraform already do that it breaks the execution.

- For resources like ec2 that can have the same name i dont see it as a good idea fails the test just because of that.  Do you think it is worthy ?

- The warn on the makefile is just about try to avoid create using names that you already commonly use in other places like  i use to create as `fmoral` but in this automation to avoid deletion problems i create like `localMoral` or `fmoral_local` or something else that you use to alias your resources.

mdrahman-suse: Good enough, thanks for sharing the details

Issue #4130, opened 2023-04-18
Enable automated dependency management to rke2 repos


matttrach:
This is the rke2 counterpart to https://github.com/k3s-io/k3s/issues/7040.

This has a slightly smaller scope, it only covers laying the groundwork for updatecli, not including the configs to keep things up to date.

### In the following repos we want to
1. generate a GH workflow file which enables dependabot, pointing at any dockerfiles present in the repo.
2. generate a GH workflow file which facilitates updatecli, including automatic cleanup of generated branches
3. add updatecli config files, specifically the directory and general values file
4. add a test showing that updatecli is enabled and working properly

### Repos
- https://github.com/rancher/rke2
- https://github.com/rancher/image-build-base
- https://github.com/rancher/image-build-calico
- https://github.com/rancher/image-build-cni-plugins
- https://github.com/rancher/image-build-containerd
- https://github.com/rancher/image-build-coredns
- https://github.com/rancher/image-build-crictl
- https://github.com/rancher/image-build-dns-nodecache
- https://github.com/rancher/image-build-etcd
- https://github.com/rancher/image-build-flannel
- https://github.com/rancher/image-build-ib-sriov-cni
- https://github.com/rancher/image-build-k8s-metrics-server
- https://github.com/rancher/image-build-kubernetes
- https://github.com/rancher/image-build-multus
- https://github.com/rancher/image-build-rke2-cloud-provider
- https://github.com/rancher/image-build-runc
- https://github.com/rancher/image-build-sriov-cni
- https://github.com/rancher/image-build-sriov-network-device-plugin
- https://github.com/rancher/image-build-sriov-network-resources-injector
- https://github.com/rancher/image-build-sriov-operator
- https://github.com/rancher/image-build-whereabouts

matttrach:
Dependabot has now been added to all repos except rke2 itself, it is ready there, but waiting for thaw.

matttrach:
Once these PRs merge updatecli and dependabot will be added to all mentioned repos.
Starting on https://github.com/rancher/rke2/issues/4143

matttrach:
Waiting for review on https://github.com/rancher/rke2/issues/4143
Starting https://github.com/rancher/rke2/issues/4165

matttrach:
This is now complete, each repo will need individual investigation for adding updatecli configs to manage dependencies in unstructured formats.

PR #4133, opened 2023-04-19
Add dependabot

matttrach:
This adds basic dependabot functionality for Dockerfiles and github workflows.
This contributes to https://github.com/rancher/rke2/issues/4130

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+
+  # Maintain dependencies for Docker Images
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,9 @@ updates:
   # Maintain dependencies for Docker Images
   - package-ecosystem: "docker"
     directory: "/"
+    ignore:
+      - dependency-name: "rancher/hardened-build-base"
+      - dependency-name: "rancher/hardened-kubernetes"
     schedule:
       interval: "weekly"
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+
+  # Maintain dependencies for Docker Images
+  - package-ecosystem: "docker"
```
brandond, line 5: I'm curious what this will find; in general we're pretty intentional about bumping things in the Dockerfile.

brandond (commented):
I'm good with having it work on Actions but curious what it is going to try to do in the Dockerfile.

macedogm (approved):

.github/dependabot.yml
matttrach: It will generate PRs for all of them. Which lines should we ignore?

brandond: Probably should ignore these, as the golang and kubernetes versions need to be manually kept in sync, and only bumped as part of the patch release process
* https://github.com/rancher/rke2/blob/master/Dockerfile#L4
* https://github.com/rancher/rke2/blob/9ce3825ec005872a0873307844ba99a2bdd4ee19/Dockerfile#L141
* https://github.com/rancher/rke2/blob/9ce3825ec005872a0873307844ba99a2bdd4ee19/Dockerfile#L143

matttrach: I can [configure dependabot to ignore certain dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore)

matttrach: these new lines should have dependabot completely ignore those two dependencies in all Dockerfiles, as PRs are generated we can iterate on this and add more lines to ignore if we chose

brandond (approved):

.github/dependabot.yml
macedogm: Just a note/reminder that Updatecli can be used for such updates in case more control logic is needed around which images to update and how to bump them.

rbrtbnfgl (approved):

PR #4108, opened 2023-04-11
Bump to Kubernetes 1.27.1

brandond:
#### Proposed Changes ####

Bump images and golang versions for Kubernetes 1.27.

Still waiting on k3s updates to bump versions in go.mod

#### Types of Changes ####

Version bump
Upstream minor

#### Verification ####

Check Kubernetes version

#### Linked Issues ####

*

PR #4188, opened 2023-05-09
[Release-1.25]Bump golangci-lint for golang 1.20 compat and fix warnings

osodracnai:
#### Proposed Changes ####
Backporting from #4108


dereknola:
We don't bump golang versions in release branches, we just match with what version upstream uses. I'm not sure this PR is necessary, as we will never bump golang to 1.20 for this branch.

brandond:
@dereknola we should keep the linter up to date regardless of what golang version is in use, and the linter raised valid findings on the flag vars shadowing built-in function names. Since we changed the flags on the master branch, and we backport everything, we also need to backport this.

brandond (approved):

dereknola (approved):

PR #4226, opened 2023-05-12
Refactoring of Restart Cluster Server and Add the Certificate Rotation

est-suse:
#### Proposed Changes ####
Add rotation certificate check and refactor the restart check

Same as the PR https://github.com/k3s-io/k3s/pull/7097

#### Types of Changes ####
Test

#### Verification ####
Run the test

### DIFF
--- a/tests/e2e/validatecluster/validatecluster_test.go
+++ b/tests/e2e/validatecluster/validatecluster_test.go
@@ -201,25 +201,6 @@ var _ = Describe("Verify Basic Cluster Creation", Ordered, func() {
 	})
 
 	Context("Validate restart", func() {
-		It("Deletes daemonset", func() {
-			_, err := e2e.DeployWorkload("daemonset.yaml", kubeConfigFile)
-			Expect(err).NotTo(HaveOccurred(), "Daemonset manifest not deployed")
-			defer e2e.DeleteWorkload("daemonset.yaml", kubeConfigFile)
-			nodes, _ := e2e.ParseNodes(kubeConfigFile, false)
-
-			Eventually(func(g Gomega) {
-				pods, _ := e2e.ParsePods(kubeConfigFile, false)
-				count := e2e.CountOfStringInSlice("test-daemonset", pods)
-				g.Expect(len(nodes)).Should((Equal(count)), "Daemonset pod count does not match node count")
-				podsRunning := 0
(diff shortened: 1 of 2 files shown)


tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -133,7 +134,7 @@ var _ = Describe("Verify Basic Cluster Creation", Ordered, func() {
 		}, "240s", "5s").Should(ContainSubstring("test-loadbalancer"), "failed cmd: "+cmd)
 	})
 
-	It("Verifies Restart", func() {
+	/*It("Verifies Restart", func() {
```
brandond, line 137: delete it if you want it gone, don't just comment it out

est-suse: Yes thanks

est-suse:
> You will need to change the function for start /stop from k3s to rke2... also you should test this locally, Rke2 restarts much slower than k3s so the timeouts may need to be larger.

Done, Thanks, could you please take a look again?

tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -247,8 +199,132 @@ var _ = Describe("Verify Basic Cluster Creation", Ordered, func() {
...
+				podsRunningAr := 0
+				for _, pod := range pods {
+					if strings.Contains(pod.Name, "test-daemonset") && pod.Status == "Running" && pod.Ready == "1/1" {
+						podsRunningAr++
+					}
+				}
+				g.Expect(len(nodes)).Should((Equal(podsRunningAr)), "Daemonset pods are not running after the restart")
+			}, "1120s", "5s").Should(Succeed())
+		})
+	})
```
rancher-max, line 245: I don't know how this is working -- the first `It` function has the `defer` statement to remove the daemonset workload, but the next one then expects it to be present. I guess this is the same code in k3s and it is working there too, but how? @dereknola any ideas?

rancher-max (commented):
This LGTM as long as it works with the already-defined timeouts, which it seems it does since CI is passing.

brandond (approved):

dereknola (approved):

ShylajaDevadiga (approved):

PR #4256, opened 2023-05-17
Add el9 to install script

galal-hussein:
#### Proposed Changes ####
Add el9 to the install script

#### Linked Issues ####

- https://github.com/rancher/rke2-selinux/issues/36
- https://github.com/rancher/rke2/issues/4285

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -492,7 +492,7 @@ do_install_rpm() {
                     :
                     ;;
                 9) # We are currently using EL8 packages for EL9 as well
-                    maj_ver="8"
+                    maj_ver="9"
                     ;;
                 *) # In certain cases, like installing on Fedora, maj_ver will end up being something that is not 7 or 8
                     maj_ver="7"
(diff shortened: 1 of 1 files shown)


briandowns (approved):

install.sh
```diff
@@ -487,7 +487,7 @@ do_install_rpm() {
                     :
                     ;;
                 9) # We are currently using EL8 packages for EL9 as well
```
brandond, line 489:
```suggestion
                9)
```

brandond (changes requested):

install.sh
brandond: Actually we don't need this at all. Just take out the whole `9)` case and add 9 to the `7|8)` case up above.

rancher-max:
https://github.com/rancher/rke2/issues/4285

install.sh
```diff
@@ -564,6 +571,16 @@ EOF
     fi
 }
 
+check_available_upgrades() {
+    set +e
+    available_upgrades=$(yum -q --refresh list $1 --upgrades | tail -n 1 | awk '{print $2}')
```
brandond, line 576: Don't we need to use `${rpm_installer}` for this, not `yum`?

PR #4260, opened 2023-05-18
V1.26.5+rke2r1

galal-hussein:
#### Proposed Changes ####

Upgrade Kubernetes to V1.26.5+rke2r1


brandond:
You should also bump the golang/hardened-build-base versions to track with upstream.

brandond (approved):

briandowns (approved):

PR #4262, opened 2023-05-18
V1.24.14+rke2r1

galal-hussein:


brandond (commented):
You should also bump the golang/hardened-build-base versions to track with upstream.

brandond (approved):

Issue #4194, opened 2023-05-09
Resolve tier 1 image CVEs that are identified by security team


caroline-suse-rancher:
Refer to Jira ticket for more detail: https://jira.suse.com/browse/SURE-6236

matttrach:
from the JIRA issue:
Resolve all the applicable CVEs flagged by image-scanning in tier 1 images that belong to team/rke2-k3s:

rancher/k3s-upgrade
rancher/rke2-cloud-provider
rancher/rke2-runtime
rancher/rke2-upgrade
rancher/system-agent-installer-k3s
rancher/system-agent-installer-rke2
rancher/system-upgrade-controller
The tier 1 images can be checked in image-scanning's dashboard - https://github.com/rancherlabs/image-scanning/issues/2062. Based on the image name, the issues can be filtered in https://github.com/rancherlabs/image-scanning/issues?q=is%3Aopen by adding the labels "team/rke2-k3s" and "cve/release/v2.7-head".

We are not listing the exact issues here, because new versions of the above images might be flagged by image-scanning in the meantime.

Upstream CVEs are outside the scope of this work, giving that we can't control when they are patched.

Please consult with the Security team in case of questions.

matttrach:
https://github.com/rancherlabs/image-scanning/issues?q=is%3Aopen+label%3Ateam%2Frke2-k3s+label%3Acve%2Frelease%2Fv2.7-head

matttrach:
[Just RKE2 CVEs](https://github.com/rancherlabs/image-scanning/issues?q=is%3Aopen+label%3Ateam%2Frke2-k3s+label%3Acve%2Frelease%2Fv2.7-head+label%3Acve%2Frke2All)

matttrach:
[Not Mirrored](https://github.com/rancherlabs/image-scanning/issues?q=is%3Aopen+label%3Ateam%2Frke2-k3s+label%3Acve%2Frelease%2Fv2.7-head+label%3Acve%2Frke2All+-label%3Acve%2Fimage%2Fmirrored+sort%3Arelevance-desc+): 86
[Only Mirrors](https://github.com/rancherlabs/image-scanning/issues?q=is%3Aopen+label%3Ateam%2Frke2-k3s+label%3Acve%2Frelease%2Fv2.7-head+label%3Acve%2Frke2All+label%3Acve%2Fimage%2Fmirrored+sort%3Arelevance-desc+): 91

matttrach:
[...]

PR #4365, opened 2023-06-14
Add new image_xref dev doc

matttrach:
The goal is to document all of the images we use, how they are built, what they are based on, what binaries are in them and how we build those binaries. 
This is a compilation of the investigation from https://github.com/rancher/rke2/issues/4194.


briandowns:
I agree. Updating the existing would probably be better and easier to read.

matttrach:
I  took your concerns into consideration and tried to clarify and separate the data as best as I could. I think this research is valuable to developers working on Rke2 in its current format and we should accept it into the codebase. I minimally altered the image_sources.md to reflect a new viewpoint on the data as well as clarify the goals and uses of the other documents. I don't have a lot more time to spend on this at the moment, so I hope we can accept this into the codebase as it is and iterate further in the future.

Issue #4411, opened 2023-07-05
Not possible to read calico/felix logs in rke2/windows


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Node(s) CPU architecture, OS, and Version: 

OS: Windows

Cluster Configuration:

Any server Linux and at least one windows agent

**Describe the bug:**

The current code is not providing a meaningful way of checking the output of the calico or felix executing when doing rke2-windows with calico cni plugin. The code uses:
```
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
```
however, that does not work and in the rke2 service logs, we see nothing. In any case, that would be a non-ideal solution because those logs can be verbose. It is best if we had them in one file. According to calico, that should happen by default or can be configured with env vars (e.g.  FELIX_LOGFILEPATH). However, it is not working in Windows

**Steps To Reproduce:**

- Installed RKE2: 

Deploy a windows node and try to find `felix.log`, it does not exist. Also check the logs of the rke2 service and there is nothing related to calico or felix

**Expected behavior:**

Be able to read logs in order to debug in case problems arise

**Actual behavior:**

Impossible to know what is going on

**Additional context / logs:**

ShylajaDevadiga:
### Validated on master branch using commit id 33caf61cf1fc71cca522eaeac1d3541b5f3c417c
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 22.04
Windows Server 2019

*Cluster Configuration:*
One linux server, one windows agent node

  *Config.yaml:*
```
$ cat /etc/rancher/rke2/config.yaml 
write-kubeconfig-mode: 644
cni: calico
node-external-ip: <IP>
token: <TOKEN>

```
 ## Validation steps
1. Copy config.yaml
2. Install rke2 server on linux node and agent on windows node
3. Check felix.log under c:\var\log dir

## Validation results:
Ubuntu node
```
$ curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_COMMIT=33caf61cf1fc71cca522eaeac1d3541b5f3c417c sh -

$ sudo systemctl enable rke2-server
$ sudo systemctl start rke2-server
```
Windows node
```
Enable-WindowsOptionalFeature -Online -FeatureName containers –All
Invoke-WebRequest -Uri https://raw.githubusercontent.com/rancher/rke2/master/install.ps1 -Outfile install.ps1
[...]

PR #4412, opened 2023-07-05
Add log files for felix and calico in rke2-windows

manuelbuil:
#### Proposed Changes ####

This PR dumps felix and calico outputs into different log files. These log files are under `C:/var/log/` which is a directory created by rke2 and removed when uninstalling.

As explained in the issue, there is currently no way to check the logs of calico or felix.

Apart from this, I am changing:
```
	if err := cmd.Run(); err != nil {
		logrus.Errorf("Felix exited: %v", err)
}
```

because when felix process fails ==>  `err = nil` and we never get that log. This PR will allow us to see the message "Felix exited" in the rke2 service logs and then we could investigate what happened by looking at the felix logs under  `C:/var/log/felix.log`

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy rke2 with a windows node and try to find felix logs (e.g. the described default felix.log in the docs) or by checking the rke2 service logs. You will find nothing.

When using this PR, you will see the logs in `C:/var/log/felix.log`. If rke2 is uninstall, that file is removed.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/4411

### DIFF
--- a/pkg/windows/calico.go
+++ b/pkg/windows/calico.go
@@ -125,6 +125,7 @@ const (
 	CalicoSystemNamespace  = "calico-system"
 	CalicoChart            = "rke2-calico"
 	calicoNode             = "calico-node"
+	calicoLogPath          = "C:\\var\\log\\"
 )
 
 // Setup creates the basic configuration required by the CNI.
@@ -373,6 +374,12 @@ func (c *Calico) overrideCalicoConfigByHelm(restConfig *rest.Config) error {
 }
 
 func startFelix(ctx context.Context, config *CalicoConfig) {
+	outputFile, err := os.Create(calicoLogPath + "felix.log")
+	if err != nil {
+		logrus.Fatalf("error creating felix.log: %v", err)
+		return
+	}
+	defer outputFile.Close()
 	specificEnvs := []string{
 		fmt.Sprintf("FELIX_FELIXHOSTNAME=%s", config.Hostname),
 		fmt.Sprintf("FELIX_VXLANVNI=%s", config.Felix.Vxlanvni),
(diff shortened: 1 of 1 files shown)


pkg/windows/calico.go
```diff
@@ -373,6 +374,11 @@ func (c *Calico) overrideCalicoConfigByHelm(restConfig *rest.Config) error {
 }
 
 func startFelix(ctx context.Context, config *CalicoConfig) {
+	outputFile, err := os.Create(calicoLogPath + "felix.log")
+	if err != nil {
+		logrus.Errorf("error creating felix.log: %v", err)
```
briandowns, line 379: Should we also return here?

manuelbuil: I am undecided here, so I would really appreciate your opinion. This is what I think: If the file does not get created, then we will get no logs. However, the process could still run successfully, as logging is not an essential part of the execution, so perhaps by just warning the user that there is a problem when creating the logging file is enough. This is actually what the code is currently doing. Are there any best practices which describe what to do in this case? Perhaps we could follow those

brandond: Is there any reason why this can't return an error, as `startCalico() ` does? This needs to either return an error which can be handled by the caller, or it needs so `Fatalf` to avoid panicing for the reason Brian mentioned.

manuelbuil: thanks for the feedback. I think it makes sense

brandond: yes. if it's a goroutine, then it shouldn't return, and should `Fatalf` on an unrecoverable error.

brandond (approved):

rbrtbnfgl (approved):

Issue #4485, opened 2023-07-19
Add test to avoid problems with chart images


manuelbuil:
**Is your feature request related to a problem? Please describe.**

From time to time we, humans, introduce image-related bugs into the rke2 code because:
* we forgot to sync helm-chart images with airgap images (e.g. https://github.com/rancher/rke2/issues/4480)
* we forgot to add "system-default-registry" bits into the chart (e.g. https://github.com/rancher/rke2/issues/3479)

I think it should be easy to verify with a simple test that a PR is not adding any bug related to these problems. Probably limit its run to PRs that update Dockerfile, which is the file changed to update charts.

**Describe the solution you'd like**

Test that would block the PR because it is adding a bug related to chart images

**Describe alternatives you've considered**

**Additional context**

PR #4615, opened 2023-08-10
Add chart validation tests

thomasferrandiz:
#### Proposed Changes ####

validate-charts runs as part of 'make validate' step and checks that all images used in packaged charts:
- use systemGlobalRegistry
- are present in script/build-images

#### Types of Changes ####

Tests

#### Linked Issues ####

https://github.com/rancher/rke2/issues/4485
#### User-Facing Change ####

```release-note

```

#### Further Comments ####

At the moment, the new test fails because some charts don't use `systemGlobalRegistry` at all.
I'm not sure what would be the best way to move forward:
- make the test informative only for now
- fix the problematic charts in another PR before merging this one

Also I don;t know if the `busybox` image should be an exception or not.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN set -x && \
     py3-pip \
     pigz \
     tar \
-    yq
+    yq \
+    helm
 
 RUN if [ "${ARCH}" = "amd64" ]; then \
     	apk --no-cache add mingw-w64-gcc; \
@@ -105,21 +106,7 @@ ARG KUBERNETES_VERSION=""
 ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
 RUN echo ${CACHEBUST}>/dev/null
-RUN CHART_VERSION="1.14.200"                  CHART_FILE=/charts/rke2-cilium.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="v3.26.1-build2023100902"   CHART_FILE=/charts/rke2-canal.yaml          CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="v3.26.101"                 CHART_FILE=/charts/rke2-calico.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="v3.26.101"                 CHART_FILE=/charts/rke2-calico-crd.yaml     CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="1.24.007"                  CHART_FILE=/charts/rke2-coredns.yaml        CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="4.8.200"                   CHART_FILE=/charts/rke2-ingress-nginx.yaml  CHART_BOOTSTRAP=false  /charts/build-chart.sh
-RUN CHART_VERSION="2.11.100-build2023051511"  CHART_FILE=/charts/rke2-metrics-server.yaml CHART_BOOTSTRAP=false  /charts/build-chart.sh
-RUN CHART_VERSION="v4.0.2-build2023081101"    CHART_FILE=/charts/rke2-multus.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="1.5.100"                   CHART_FILE=/charts/rancher-vsphere-cpi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="3.0.1-rancher101"          CHART_FILE=/charts/rancher-vsphere-csi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="0.2.200"                   CHART_FILE=/charts/harvester-cloud-provider.yaml CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="0.1.1600"                  CHART_FILE=/charts/harvester-csi-driver.yaml     CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller-crd.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.302"                   CHART_FILE=/charts/rke2-snapshot-validation-webhook.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
+RUN /charts/build-charts.sh
 RUN rm -vf /charts/*.sh /charts/*.md
 
 # rke2-runtime image
--- a/charts/build-charts.sh
+++ b/charts/build-charts.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -eux -o pipefail
+
+while read version filename bootstrap; do
+    CHART_VERSION=$version CHART_FILE=$filename CHART_BOOTSTRAP=$bootstrap /charts/build-chart.sh
+done <<< $(yq e '.charts[] | [.version, .filename, .bootstrap] | join(" ")' /charts/chart_versions.yaml)
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -89,10 +89,10 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-harvester.txt
     ${REGISTRY}/rancher/harvester-cloud-provider:v0.2.0
     ${REGISTRY}/rancher/mirrored-kube-vip-kube-vip-iptables:v0.6.0
     ${REGISTRY}/rancher/harvester-csi-driver:v0.1.5
-    ${REGISTRY}/rancher/longhornio-csi-node-driver-registrar:v2.3.0
-    ${REGISTRY}/rancher/longhornio-csi-resizer:v1.2.0
-    ${REGISTRY}/rancher/longhornio-csi-provisioner:v2.1.2
-    ${REGISTRY}/rancher/longhornio-csi-attacher:v3.2.1
+    ${REGISTRY}/rancher/mirrored-longhornio-csi-node-driver-registrar:v2.3.0
+    ${REGISTRY}/rancher/mirrored-longhornio-csi-resizer:v1.2.0
+    ${REGISTRY}/rancher/mirrored-longhornio-csi-provisioner:v2.1.2
+    ${REGISTRY}/rancher/mirrored-longhornio-csi-attacher:v3.2.1
 EOF
 fi
 # Continue to provide a legacy airgap archive set with the default CNI images
--- a/scripts/validate-charts
+++ b/scripts/validate-charts
@@ -1,0 +1,190 @@
+#!/usr/bin/env bash
+set -e
+source ./scripts/version.sh
+set +x
+
+info() {
+    echo '[INFO] ' "$@"
+}
+
+warn() {
+    echo '[WARN] ' "$@" >&2
+}
+
+error() {
+    echo '[ERROR] ' "$@" >&2
+}
+
+fatal() {
+    echo '[ERROR] ' "$@" >&2
+    exit 1
+}
+
+cleanup() {
+  exit_code=$?
+  trap - EXIT INT
+  rm -rf  /tmp/tmp.*.tar.gz
+  exit ${exit_code}
+}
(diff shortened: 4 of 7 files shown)


brandond:
> fix the problematic charts in another PR before merging this one

I would probably fix the charts in rke2-charts, then have two commits in this PR:
* first commit to add the test
* second commit to update the charts and make the test pass

Dockerfile
```diff
@@ -105,21 +106,12 @@ ARG KUBERNETES_VERSION=""
...
-RUN CHART_VERSION="v4.0.2-build2023081100"    CHART_FILE=/charts/rke2-multus.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="1.5.100"                   CHART_FILE=/charts/rancher-vsphere-cpi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="3.0.1-rancher101"          CHART_FILE=/charts/rancher-vsphere-csi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="0.2.200"                   CHART_FILE=/charts/harvester-cloud-provider.yaml CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="0.1.1600"                  CHART_FILE=/charts/harvester-csi-driver.yaml     CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller-crd.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.300"                   CHART_FILE=/charts/rke2-snapshot-validation-webhook.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
+COPY chart_versions.csv /charts/chart_versions.csv
```
brandond, line 109: If you moved this file to charts/ in the repo, it would be covered by the copy on line 106 and you wouldn't have to do it separately.

Plus then the file wouldn't clutter the base of the repo.

chart_versions.csv
```diff
@@ -0,0 +1,15 @@
+1.14.000,/charts/rke2-cilium.yaml,true
```
brandond, line 1: CSV is kind of a step backwards in terms of readability and maintainability. I get that it makes it easy to consume from bash and so on, but would you consider using yaml or perhaps just tab/whitespace delimiting instead, so that it's easy for us (the maintainers) to manage?

We regularly bump chart versions and I'd like to make that as easy and mistake-proof as possible.

brandond (changes requested):

Dockerfile
```diff
@@ -105,21 +106,15 @@ ARG KUBERNETES_VERSION=""
...
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller-crd.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.300"                   CHART_FILE=/charts/rke2-snapshot-validation-webhook.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
+RUN readarray charts < <(yq e -o=j -I=0 '.charts[]' charts/chart_versions.yaml ) \
+    for chart in "${charts[@]}"; do \
+        version=$(echo "$chart" | yq e '.version' -) \
+        filename=$(echo "$chart" | yq e '.filename' -) \
+        bootstrap=$(echo "$chart" | yq e '.bootstrap' -) \
+        CHART_VERSION=$version CHART_FILE=$filename CHART_BOOTSTRAP=$bootstrap /charts/build-chart.sh; \
+    done
```
brandond, line 115: Maybe move this logic into the build-chart.sh script, or another script that calls it? No need to put complicated loops and yq expressions inline in the Dockerfile.

scripts/build-images
```diff
@@ -61,6 +61,13 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-calico.txt
     ${REGISTRY}/rancher/mirrored-calico-apiserver:v3.26.1
 EOF
 
+xargs -n1 -t docker image pull --quiet << EOF > build/images-harvester-csi.txt
```
brandond, line 64: These are already in `images-harvester.txt` down below; why are we adding another list that includes only the harvester CSI without the CPI?

chart_versions.csv
thomasferrandiz: done

Dockerfile
thomasferrandiz: done

scripts/build-images
thomasferrandiz: I added them because the test was failing without them.
But in fact the problem was that the images in `images-harvester.txt` were not using the `mirrored-` prefix so I corrected that instead.

Dockerfile
thomasferrandiz: done

Dockerfile
```diff
@@ -105,21 +106,9 @@ ARG KUBERNETES_VERSION=""
...
-RUN CHART_VERSION="1.5.100"                   CHART_FILE=/charts/rancher-vsphere-cpi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="3.0.1-rancher101"          CHART_FILE=/charts/rancher-vsphere-csi.yaml CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="0.2.200"                   CHART_FILE=/charts/harvester-cloud-provider.yaml CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="0.1.1600"                  CHART_FILE=/charts/harvester-csi-driver.yaml     CHART_BOOTSTRAP=true /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.202"                   CHART_FILE=/charts/rke2-snapshot-controller-crd.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
-RUN CHART_VERSION="1.7.300"                   CHART_FILE=/charts/rke2-snapshot-validation-webhook.yaml CHART_BOOTSTRAP=false /charts/build-chart.sh
+RUN /charts/build-charts.sh
+
+
```
brandond, line 111: nit
```suggestion
```

brandond (approved):

scripts/build-images
brandond: hey look, you're finding bugs already!

charts/build-charts.sh
```diff
@@ -0,0 +1,12 @@
...
+set -eux -o pipefail
+
+readarray charts < <(yq e -o=j -I=0 '.charts[]' charts/chart_versions.yaml )
+for chart in "${charts[@]}"; do
+    version=$(echo "$chart" | yq e '.version' -)
+    filename=$(echo "$chart" | yq e '.filename' -)
+    bootstrap=$(echo "$chart" | yq e '.bootstrap' -)
+
+    CHART_VERSION=$version CHART_FILE=$filename CHART_BOOTSTRAP=$bootstrap /charts/build-chart.sh
+done
```
brandond, line 12:
```suggestion
while read version filename bootstrap; do
    CHART_VERSION=$version CHART_FILE=$filename CHART_BOOTSTRAP=$bootstrap /charts/build-chart.sh
done <<< $(yq e '.charts[] | [.version, .filename, .bootstrap] | join(" ")' charts/chart_versions.yaml)
```

manuelbuil (approved):
validate-charts is run as part of the CI for PRs, right?

scripts/validate-charts
```diff
@@ -0,0 +1,138 @@
...
+
+failed=0
+
+if [ ${#NO_SYSTEM_REGISTRY[@]} -ge 1 ]; then
+    failed=1
+    for chart in "${!NO_SYSTEM_REGISTRY[@]}"
+    do
+        error "Images not using systemGlobalRegistry in chart '$chart': ${NO_SYSTEM_REGISTRY[$chart]}"
+    done
+    error "Please use systemGlobalRegistry for above images"
```
brandond, line 124:
```suggestion
        error "Images not using global.systemDefaultRegistry in chart '$chart': ${NO_SYSTEM_REGISTRY[$chart]}"
    done
    error "Please use global.systemDefaultRegistry for above images"
```

scripts/validate-charts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    ' $yaml_tmp
+}
+
+check_airgap() {
+    chart_version=$1
+    chart_name=$2
+    chart_tmp=$3
+
+    if [[ $chart_name == 'rancher-vsphere-cpi' ]]; then
```
brandond, line 68: is this chart untestable? if so, why?

scripts/validate-charts
thomasferrandiz: done

scripts/validate-charts
thomasferrandiz: the chart doesn't work well with `helm template` because of the `versionOverrides` in its `values.yaml`.
Since there is no actual k8s cluster to determine the version of the image to use, `helm template` uses the `.Capabilities.APIVersion` of the version of k8s that helm was built against so it chooses an image that doesn't match the one we need in the airgap file.

To workaround this we would need manually install the right version of Helm which would be different for the main branch and each release branch and would need to be updated for every new k8s version. Since it's only for this chart, I found it simpler to skip the test.

See https://github.com/rancher/vsphere-charts/blob/7c1f2d4ee52548bcc8f7aee0e986b338a229ac8a/charts/rancher-vsphere-cpi/values.yaml#L28

brandond: You don't need to install different versions of Helm, you just need to use the `--kube-version` flag to specify the Kubernetes version you want it to use when templating. You could probably just pass that the `$VERSION` environment variable to make sure the chart handles it properly.

brandond (approved):

brandond:
I believe the test failure is valid, and that chart doesn't support 1.28 yet?

thomasferrandiz:
> I believe the test failure is valid, and that chart doesn't support 1.28 yet.
> 
> https://github.com/rancher/rke2-charts/blob/main/charts/rancher-vsphere-csi/rancher-vsphere-csi/3.0.1-rancher101/values.yaml#L142
> 
> This will be an ongoing problem whenever upstream support lags on a new minor.

I made a PR to update the upstream chart: https://github.com/rancher/vsphere-charts/pull/62
but I have no idea of how to test it as I don't know the charts.

In any case, the problem will happen for every new minor release so I guess it would still be better if the test doesn't fail because of this chart.

brandond:
Yeah, but we don't want to ignore problems on minors that ARE supported.

Some of the charts have an annotation that we could probably use to decide whether to fail or warn, base on whether or not the chart is expected to support the current Kubernetes version: `  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.28.0-0'`

https://github.com/rancher/rke2-charts/blob/main/charts/rancher-vsphere-csi/rancher-vsphere-csi/3.0.1-rancher101/Chart.yaml#L4

scripts/validate-charts
```diff
@@ -0,0 +1,189 @@
...
+upper_bound_patch="${upper_bound##*.}"
+upper_bound_dash="${upper_bound_patch##*-}"
+upper_bound_patch="${upper_bound_patch%-*}"
+
+if [ "$lower_bound_major" -le "$kube_version_major"  ] && \
+    [ "$kube_version_major" -le "$upper_bound_major" ] && \
+    [ "$lower_bound_minor" -le "$kube_version_minor"  ] && \
+    [ "$kube_version_minor" -le "$upper_bound_minor" ] && \
+    [ "$lower_bound_patch" -le "$kube_version_patch"  ] && \
+    [ "$kube_version_patch" -le "$upper_bound_patch" ]; then
```
brandond, line 99: this is a little gnarly but I think it'll work...

scripts/validate-charts
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env bash
+set -ex
```
brandond, line 2: I would recommend running this without `set -x`, it is contributing a TON to the verbosity and length of the CI output.

```suggestion
set -e
```

brandond (changes requested):

brandond (approved):

scripts/validate-charts
thomasferrandiz: yes it was quite unreadable.
I added
`set +x` because `versions.sh` also sets `set -x`

rbrtbnfgl (approved):

PR #4766, opened 2023-09-14
Add ADR for new KDM branching strategy

rancher-max:
#### Proposed Changes ####

Proposal to change the way we branch in KDM to allow for k3s/rke2 to have its own branches for rancher versions.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/006-kdm-strategy.md
+++ b/docs/adrs/006-kdm-strategy.md
@@ -1,0 +1,31 @@
+# 6. KDM Branching Strategy for RKE2
+
+Date: 2023-09-14
+
+## Status
+
+New
+
+## Context
+
+Rancher Manager uses [KDM](https://github.com/rancher/kontainer-driver-metadata/) for its integration of RKE2. This manages the following behaviors for each RKE2 version:
+- Minimum Rancher Manager version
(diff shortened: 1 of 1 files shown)


matttrach (approved):
This sort of breaks my brain, but I can't think of a better way without asking Rancher to change how they ingress this data.
It generally feels like this sort of branching strategy can cause merge pain, but that seems less painful than our current strategy.

docs/adrs/006-kdm-strategy.md
```diff
@@ -0,0 +1,31 @@
...
+
+Date: 2023-09-14
+
+## Status
+
+New
+
+## Context
+
+Rancher Manager uses [KDM](https://github.com/rancher/kontainer-driver-metadata/) for its integration of RKE2. This manages the following behaviors for each RKE2 version:
```
brandond, line 11: Should the ADR go here, in KDM, or ecm-distro-tools? This feels like an overarching issue where we want to observe parity for both K3s and RKE2, and the change were discussing doesn't really affect the distro - so I don't feel like it belongs in this repo.

brandond (changes requested):

Issue #4775, opened 2023-09-19
Static pods do not (re)start after fix issue 3725


xhejtman:
**Environmental Info:**
RKE2 Version:  1.26.9-rke2r1

https://github.com/rancher/rke2/issues/3725

Node(s) CPU architecture, OS, and Version:  Linux kub-b1.priv.cerit-sc.cz 6.2.0-32-generic #32~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 18 10:40:13 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration: 5 servers, 30 agents

**Describe the bug:**
it seems that after resolving issue https://github.com/rancher/rke2/issues/3725, static pods (kube-proxy) do not restart after upgrade (from 1.26.7 in my case). It may be connected with static cpu topology manager.

config part:
```
kubelet-arg:
  - "v=5"
  - 'config=/var/lib/rancher/rke2/agent/etc/kubelet.config'
  - "system-reserved=memory=8Gi"
  - "reserved-cpus=64-127"
  - "max-pods=160"
  - "kube-api-qps=100"
  - "kube-api-burst=100"
  - "cpu-manager-policy=static"
  - "topology-manager-policy=best-effort"
```

syslog complains about but it may be unrelated:
`kubepods-besteffort.slice: Failed to set 'cpuset.cpus' attribute on '/kubepods.slice/kubepods-besteffort.slice' to '': No space left on device`

I removed `/var/lib/kubelet/cpu_manager_state`, deleted kube-proxy pod, restarted rke2-agent and it is working now. On some nodes, I had to repeat it several times otherwise kube-proxy was in pending state and kubelet complained that kube-proxy cannot start now.

similar happened for control plane, where restarting rke2-server resolved all static pods. 

I added: `AllowedCPUs=63,127` to rke2-agent/server systemd unit to limit rke2, kubelet and containerd to those 2 CPUs only but it does not seem to be the cause of this issue.

brandond:
The issue you linked never prevented the static pods from actually being changed, it just made them appear unchanged if you queried the mirror pods from the apiserver. 

It is documented upstream that you have to manually clean up the state files if you change the policy on a node after the kubelet has already started; this is not a defect in rke2.

xhejtman:
I did not change the state, it was all the time static. Actually, if the state is not cleaned, kubelet won't start. I believe there is some issue after 1.26.7.

[...]

PR #4784, opened 2023-09-22
Handle restart attempts in static pod manifest checks

brandond:
#### Proposed Changes ####

Fixes issue where pod restart attempts may obscure currently running pods from being found.

This switches to using upstream functions to determine the current pod sandbox, to ensure that rke2 is in sync with the kubelet.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/4775

#### User-Facing Change ####

```release-note
Fixed an issue where static pod startup checks may return false positives in the case of pod restarts
```

### DIFF
--- a/pkg/rke2/spw.go
+++ b/pkg/rke2/spw.go
@@ -17,6 +17,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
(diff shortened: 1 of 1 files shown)


matttrach (approved):
lgtm

dereknola (approved):

brandond:
Changed to using upstream `runtimeutil.PodSandboxChanged` to determine the current pod sandbox, instead of maintaining duplicate logic here.

Have requested re-review.

PR #4934, opened 2023-10-19
fix: Add adr for supporting air gapped rpm installs

matttrach:
#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####

Adr requesting support for air gapped RPM installations

### DIFF
--- a/docs/adrs/006-airgap-rpm-support.md
+++ b/docs/adrs/006-airgap-rpm-support.md
@@ -1,0 +1,55 @@
+# RPM Airgap Support
+
+We should better support RPM installation in air gapped environments
+
+## Established
+
+2023-10-20
+
+## Revisit by
+
+2024-10-01
+
+## Subject
+
+1. When releasing, we bundle RPMs into groups by OS, copy them to a directory, use createrepo to generate local repo information, tarball the directory, and add it as an artifact on the release.
+2. Given install method 'rpm' and variable 'INSTALL_RKE2_ARTIFACT_PATH', when run, the install.sh script looks for a local rpm installation at the given path and installs rke2 using that repo.
+3. Given variable 'INSTALL_CUSTOM_RPM_SITE', when run, the install.sh script uses the value of that variable when creating repo files.
+   1. specifically the `rpm_site` value in the install.sh is based on this variable
+
+## Status
+
+Requesting Feedback
+
+## Context
+
+Users who are concerned with security often deploy in air gapped environments.
+These users also often want selinux enforcing.
+The rke2 selinux policies rely on RPM installation.
+There is a significant use case for RPM installation in air gapped environments.
+This use case closely aligns with goals for rke2.
+
+Strengths:
+
+- RPM installation in air gapped environments will be as simple as tar installation in air gapped environments.
+- This improves our ability to test and enables higher quality support for this use case.
+- When well documented, bundling all dependencies and building within a specific operating system can improve reliability.
+
+Weaknesses:
+
+* RPM bundling requires another step in the release process
+* RPM bundles have the potential to include dependencies that are not supported in a particular version of an operating system
+  * this is only when RPMs have external dependencies, which ours currently don't
+
+Threats involved in not doing process:
+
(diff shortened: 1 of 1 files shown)


docs/adrs/006-airgap-rpm-support.md
```diff
@@ -0,0 +1,55 @@
...
+2023-10-20
+
+## Revisit by
+
+2024-10-01
+
+## Subject
+
+1. When releasing, we bundle RPMs into groups by OS, copy them to a directory, use createrepo to generate local repo information, tarball the directory, and add it as an artifact on the release.
+2. Given install method 'rpm' and variable 'INSTALL_RKE2_ARTIFACT_PATH', when run, the install.sh script looks for a local rpm installation at the given path and installs rke2 using that repo.
```
brandond, line 16: Does this work? I didn't even know that you could point yum at a local repo-dir instead of a http/https server.

What are the advantages of shipping a full snapshot of repo files, over just shipping the 4 RPMs (rke2-common, rke2-agent, rke2-server, and rke2-selinux)?

matttrach: Our selinux policies require rpm installation, right?
The use case is to enable air gapped installation on selinux enforcing environments.

Local yum repositories are possible: https://rpmdeb.com/devops-articles/how-to-create-local-yum-repository/

matttrach: createrepo is not a core linux utility, and is only necessary to generate the rpm repo files, running this and bundling it up for the user enables a much simpler approach, just like bundling our other images into an archive. 

Another way to think of this approach is just like the tar installation, but for rpms.

matttrach: Thank you for the context on rancher-system-agent and system-agent-installer-rke2, I will add that to the context part of this doc! 
It appears rancher-system-agent is just a program that watches for plans from something else, it would be unaffected by this change.

The system-agent-installer-rke2 appears to be a plan for the rancher-system-agent which downloads the install script and runs it with install method = 'tar'. I am not sure how this affects rpm installs, but since the RPMs install the binaries in similar places as the docs for the manual install, I would assume this would still work. I am sure that we test this upgrade path already and that the directory forces the proper selinux tags on the binaries, the change proposed in this ADR won't affect that at all.

matttrach: I have found several tools which bundle all dependencies for rke2 so that it can be copied and installed locally on a server with no internet, but most have their own installation method. I think it would be better if we provided support for this use case. We supply almost all our dependencies currently in various repos, we just need to bundle them up by operating system and enable our installer to use them.
This ADR targets the installer changes.

matttrach:
selinux policies depend on container-selinux which has different versions for each operating system

matttrach:
I need to follow up internally to see if container-selinux has dependencies that we are not aware of, or if there is a possibility of choosing a single OS to target the bundle (rhel-9 STIG), or if there is a coupling between our selinux policies and the version of container-selinux that would have this make more sense.

matttrach:
here is a link to the dependencies for the rke2 policy: 
https://github.com/rancher/rke2-selinux/blob/master/policy/centos7/rke2-selinux.spec#L36C1-L48C34

docs/adrs/006-airgap-rpm-support.md
localleon: @brandond we run a Rancher Prime environment with about 23 different airgapped clusters. All clusters are installed via the RPM methode (because we are RHEL & Ansible heavy) and imported into Rancher.  The system-upgrade-controller of Rancher currently doesn't handle upgrades of a RPM installation for us very well e.g. because it will just replace the binary and leave the rpm packages in a broken state (rpm version 1.27 and rke2 binary version 1.29). Dependencies on rke2-selinux should be handled appropriatly with the controller too.  If anything could enable the use of the system upgrade controller for RPM installations, this would greatly enhance our environment! 

Back to the ADR topic, i personally don't see the need for integrating the RPM installation into the install script. If you are opting for an RPM install you will already have the repo infrastructure to deliver said RPM to the host inplace. Then a simple `dnf install` will be sufficient for most environment

PR #4959, opened 2023-10-25
Revert "[Release-1.26] Bump ingress-nginx to v1.9.3"

dereknola:
Reverts rancher/rke2#4957. Only to turn around and reapply because golangci-lint failed.


dereknola:
Codespell failed, but since this is explicitly a revert PR I am going to ignore that. I can fix it when I reapply the reverted PR again.

cwayne18 (approved):

briandowns (approved):

PR #5095, opened 2023-12-04
remove s390x steps since the runners are disabled

thomasferrandiz:
#### User-Facing Change ####

```release-note

```


brooksn:
@thomasferrandiz you may also want to remove the s390x image manifest here: https://github.com/rancher/rke2/blob/master/manifest-runtime.tmpl#L15-L18

We're also releasing to release-1.25, release-1.26 and to release-1.27.

PR #5145, opened 2023-12-20
Add adr about rke2-flannel

manuelbuil:
#### Proposed Changes ####

ADR. I am proposing a new feature

#### Types of Changes ####

New feature proposal

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5215

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/006-add-flannel-support.md
+++ b/docs/adrs/006-add-flannel-support.md
@@ -1,0 +1,34 @@
+# Flannel support in rke2
+
+## Established
+## Revisit by
+## Status
+
+Accepted
+
+## Context
+
+Currently RKE2-windows users can only deploy with one supported CNI plugin: Calico. In the last weeks, we have had several users complaining about a TCP Reset issue that breaks their applications. For example, applications that rely on a stable connectivity towards external nodes, like Jenkins Server, are impossible to operate if TCP Resets appears.
+
+The TCP Reset issue is known by Tigera (the company behind Calico) and documented [here](https://docs.tigera.io/calico/latest/getting-started/kubernetes/windows-calico/limitations#pod-to-pod-connections-are-dropped-with-tcp-reset-packets)
+As described in that doc, the issue should only appear when using network policies. Customers state that when using network policies the problem is constant but even without network policies, the TCP Resets appear from time to time with no clear trigger for it.
+
+It is hard to understand where are those TCP Resets coming from because the code creating the virtual network infrastructure in Windows is closed sourced and the documentation is very poor and not really explaining the details, so we can only especulate. I think the problem is that windows creates a TCP Proxy in their VMSwitches when using VFP (Virtual Filtering Platform) and whenever a rule, affecting the pod, of VFP changes, the TCP Proxy is recreated and thus all TCP connections are reset. It is probably part of the design of VFP and thus complicated to really work around it. Calico uses VFP to do the network policies.
+
+We have been communicating all this to Tigera and Microsoft but so far, the resolution to the problem does not seem to be close at hand. It feels like a complicated problem created by the VFP design.
+
+Moreover, some of these users were previous users of RKE1-windows with flannel and they were happy with it. THey are asking to have the possibility to continue with flannel in RKE2.
+
+## Proposal
+
+Include flannel as a CNI plugin alternative for RKE2. We know that it has limitations and it is really simple but it seems it could be enough for Windows users that are feeling the pain of the Calico TCP Resets
+
+### Strength
+
+* We can offer a plan B for customers that can't afford getting TCP Resets
+* We have an alternative while Microsoft and Tigera fix the problem
+* We are maintainers of flannel and could support it easily
+
+### Weakness
+* Yet another cni plugin to support
+* Flannel is very limited. We should document it very well to avoid disappointment on non-knowledgeable customers


docs/adrs/006-add-flannel-support-in-windows.md
```diff
@@ -0,0 +1,35 @@
...
+### Strength
+
+* We can offer a plan B for customers that can't afford getting TCP Resets
+* We have an alternative while Microsoft and Tigera fix the problem
+* We are maintainers of flannel and could support it easily
+
+### Weakness
+* Yet another cni plugin to support
+* Flannel is very limited. We should document it very well to avoid disappointment on non-knowledgeable customers
+* How to make it only available for rke2-Windows?
```
brandond, line 35: Is it even possible to run different CNIs on different nodes? If we support flannel for Windows, my understanding is that we've need to support it for the Linux nodes as well.

docs/adrs/006-add-flannel-support-in-windows.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Strength
+
+* We can offer a plan B for customers that can't afford getting TCP Resets
+* We have an alternative while Microsoft and Tigera fix the problem
+* We are maintainers of flannel and could support it easily
+
+### Weakness
+* Yet another cni plugin to support
+* Flannel is very limited. We should document it very well to avoid disappointment on non-knowledgeable customers
```
brandond, line 34: Would we support a network policy controller in this configuration, or would clusters configured this way lack netpol support?

docs/adrs/006-add-flannel-support-in-windows.md
rbrtbnfgl: Yes this is the idea to support flannel as CNI of RKE2 for both linux and windows.

docs/adrs/006-add-flannel-support-in-windows.md
rbrtbnfgl: It doesn't support any network policy. The only CNI with network policy and windows support is calico that has the TCP reset issue.

docs/adrs/006-add-flannel-support-in-windows.md
brandond: Should we remove this as a bullet point then?

manuelbuil: yeah true, I was thinking on not supporting it on "all-linux" nodes. But it does not make sense if we anyway need to support it in linux. I'll remove it

docs/adrs/006-add-flannel-support-in-windows.md
manuelbuil: No network policies. Customers know about this and still they prefer that compared to Calico

dereknola (approved):

brandond (approved):

matttrach (commented):
While I agree that this makes a lot of sense, I would like to hear what @briandowns and @rancher-max has to say about it since it would expand our support matrix and possibly change our security assumptions.

rancher-max (approved):
I am fine with this. I do agree with you point about documentation -- it should be extremely thorough. I also wonder if we have to use an "rke2-flannel" directly or can we double dip from k3s and pull that in? Or is that what this would effectively be, just packaged as a helm chart instead of embedded?

brandond:
Ideally it would be a helm chart, same as the other CNIs.

manuelbuil:
> I am fine with this. I do agree with you point about documentation -- it should be extremely thorough. I also wonder if we have to use an "rke2-flannel" directly or can we double dip from k3s and pull that in? Or is that what this would effectively be, just packaged as a helm chart instead of embedded?

Helm chart (in linux)! In Windows, I still need to figure it out but probably it will follow Calico's way... AFAIK, Windows does not support pods with access to the root network namespace, which is a requirement for CNI plugin pods. If that is still the case, we can't use helm chart in Windows and, as in Calico, we will need to execute flannel binary (flanneld.exe) from within rke2.exe

manuelbuil:
> While I agree that this makes a lot of sense, I would like to hear what @briandowns and @rancher-max has to say about it since it would expand our support matrix and possibly change our security assumptions.

Thanks Matt. What security assumptions? I haven't thought about those

brandond:
> we can't use helm chart in Windows and, as in Calico, we will need to execute flannel binary (flanneld.exe) from within rke2.exe

Yeah, as long as we set it up like Calico, where the Linux nodes use the chart, and the Windows nodes key off the charts presence to run the binary, I think we should be fine.

manuelbuil:
> > we can't use helm chart in Windows and, as in Calico, we will need to execute flannel binary (flanneld.exe) from within rke2.exe
> 
> Yeah, as long as we set it up like Calico, where the Linux nodes use the chart, and the Windows nodes key off the charts presence to run the binary, I think we should be fine.

that's the idea

matttrach:
I was thinking there is probably some security assumptions in the use of network policies. Without network policies won't pods be basically open for connections?

manuelbuil:
> I was thinking there is probably some security assumptions in the use of network policies. Without network policies won't pods be basically open for connections?

Inside the cluster, there will be no limitation for connections to the pods. Connectivity from the outside to the pod is still limited. In any case, we will document that network policies are not available in flannel, users will be aware of this

manuelbuil:
As nobody seems against it, I'll create an Issue and cover all required PRs there

matttrach (approved):

Issue #3917, opened 2023-02-17
DaemonSet for ContainerNetworking DHCP CNI Plugin


AJMansfield:
### The Problem

When setting up an RKE2 cluster to use Multus, it's not clear what the appropriate way is to set up and configure the DHCP daemon needed to allow the ContainerNetworking DHCP IPAM plugin to function.

Though there's ways of getting this daemon to run using DaemonSets or systemd units from other projects, the fact that the binary for this daemon (`/opt/cni/dhcp`) is distributed with RKE2 Multus suggests that it ought to also be run without needing additional steps much more complicated than those for running Multus in the first place.

### The Solution I Want

I'd like to be able to add `--cni=multus-dhcp` as another RKE2 server argument, similar to specifying `--cni=multus` for getting Multus set up.

(Or equivalently from the server `config.yaml`:`
```diff
 # /etc/rancher/rke2/config.yaml
 cni:
 - multus
+- multus-dhcp
 - canal
``` 

On starting, the server would use this to place/install the appropriate manifest at `/var/lib/rancher/rke2/server/manifests/rke2-multus-dhcp.yaml`, and from that install a `rke2-multus-dhcp` Addon and create the `rke2-multus-dhcp-ds` DaemonSet to run the plugin daemon.

### The Alternative Solutions I Already Have

The solution I'm using for now, is to add a copy of [the k8snetworkingwg reference-deployment dhcp-daemonset.yaml](https://github.com/k8snetworkplumbingwg/reference-deployment/tree/master/multus-dhcp/dhcp-daemonset.yaml) to the server manifest folder myself. The DHCP plugin is perfectly functional when set up this way; the only plausible issue with it is the third-party dependency it introduces, something I will eventually need to resolve.

Before I found the DaemonSet method above, I also had it working using systemd to run the daemon directly in the host. The plugin authors have [pre-made systemd unit files for this](https://github.com/containernetworking/plugins/tree/main/plugins/ipam/dhcp/systemd) which work perfectly, and this is a superior solution in the sense that it only starts the daemon on demand (via systemd socket activation). But the daemon is already very lightweight, so the scalability disadvantage of this method led me to switch to using a DaemonSet.

AJMansfield:
> I do know that many plugins are already built in to multus

[...]

PR #5146, opened 2023-12-20
update multus chart to add optional dhcp daemonset

thomasferrandiz:
Linked issue: https://github.com/rancher/rke2/issues/3917

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -20,7 +20,7 @@ charts:
   - version: 2.11.100-build2023051511
     filename: /charts/rke2-metrics-server.yaml
     bootstrap: false
-  - version: v4.0.2-build2023081103
+  - version: v4.0.2-build2023081106
     filename: /charts/rke2-multus.yaml
     bootstrap: true
   - version: 1.5.100
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -83,6 +83,7 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-multus.txt
     ${REGISTRY}/rancher/hardened-sriov-network-resources-injector:v1.5-build20231009
     ${REGISTRY}/rancher/hardened-sriov-network-webhook:v1.2.0-build20230607
     ${REGISTRY}/rancher/hardened-whereabouts:v0.6.2-build20231009
+    ${REGISTRY}/rancher/mirrored-library-busybox:1.36.1
 EOF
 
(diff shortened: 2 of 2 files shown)


charts/chart_versions.yaml
```diff
@@ -20,7 +20,7 @@ charts:
   - version: 2.11.100-build2023051511
     filename: /charts/rke2-metrics-server.yaml
     bootstrap: false
-  - version: v4.0.2-build2023081103
+  - version: v4.0.2-build2023081104
```
brandond, line 23: Are there any new images that come with the changes to the chart, or does the DHCP daemonset use an existing image?

thomasferrandiz: no new images, it's using busybox

manuelbuil: I guess we should include busybox in the list of images for air-gap

brandond: It looks like we for some reason specifically excluded the busybox image from the chart image validation scripts. If we're going to use busybox, we probably shouldn't exclude it from the list.

thomasferrandiz: I think I excluded busybox from the validation script because it's used in several charts but is never added in `build-images` so I assumed this was done on purpose.
Also several charts use the `latest` busybox instead of set version.

scripts/build-images
```diff
@@ -83,6 +83,7 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-multus.txt
     ${REGISTRY}/rancher/hardened-sriov-network-resources-injector:v1.5-build20231009
     ${REGISTRY}/rancher/hardened-sriov-network-webhook:v1.2.0-build20230607
     ${REGISTRY}/rancher/hardened-whereabouts:v0.6.2-build20231009
+    ${REGISTRY}/library/busybox:latest
```
brandond, line 86: We can't use the `latest` tag, there are different semantics for pulling when that is used, and it is not suitable for airgap use. We need to use a fixed tag. It sounds like this might require another change to the chart though?

brandond (changes requested):

charts/chart_versions.yaml
thomasferrandiz: busybox added to `build-images`

scripts/build-images
thomasferrandiz: I updated the chart to use the mirrored busybox from rancher

brandond (approved):

manuelbuil (approved):

Issue #5136, opened 2023-12-14
 rke2-images-cilium.linux-amd64.* missing and Airgap Tarball is failing


eifelmicha:
**Environmental Info:**
RKE2 Version: v1.28.4+rke2r1

Node(s) CPU architecture, OS, and Version: Linux k8s-master-1 6.1.0-15-arm64 #1 SMP Debian 6.1.66-1 (2023-12-09) aarch64 GNU/Linux

Cluster Configuration: 3 Server with Cilium CNI

**Describe the bug:**
I would like to prefetch files in Packer to do an Tarball Installation of RKE2 and reduce startup times. According to: https://docs.rke2.io/install/airgap#tarball-method i have to download `rke2-images-core` and `rke2-images-cilium `, because of  `--cni=cilium`
However  rke2-images-cilium.linux-**amd64**.tar.zst  does exist, while  rke2-images-cilium.linux-**arm64**.tar.zst  does not.

**Steps To Reproduce:**
- Scripted Curl to download files, the `+` is escaped before:
```
ARCH=amd64
if [ "$(uname -m)" = "aarch64" ]; then ARCH=arm64; fi
curl -OLs https://github.com/rancher/rke2/releases/download/$RKE2_VERSION/rke2-images-core.linux-$ARCH.tar.zst
curl -OLs https://github.com/rancher/rke2/releases/download/$RKE2_VERSION/rke2-images-cilium.linux-$ARCH.tar.zst
curl -OLs https://github.com/rancher/rke2/releases/download/$RKE2_VERSION/sha256sum-$ARCH.txt
```

**Expected behavior:**
Successful download

**Actual behavior:**
`CHECKSUM_CILIUM_EXPECTED=$(grep "rke2-images-cilium.linux-$ARCH.tar.zst" "sha256sum-$ARCH.txt" | awk '{print $1}');` failed, because no file and checksum is present.

**Additional context / logs:**

- https://docs.rke2.io/install/airgap#tarball-method

brandond:
I believe that up until recently, some or all of the cilium images did not support arm64, so we disabled packaging it for that architecture.

https://github.com/rancher/rke2/blob/c1494f5de1d2f6ae26cbb7d8ec365344dc1209d8/scripts/build-images#L36-L37

@manuelbuil @thomasferrandiz can we revisit this, and see if we can package cilium airgap images for arm64?

brandond:
[...]

PR #5154, opened 2023-12-21
Enable arm64 based images for calico, multus and harvester

mgfritch:
Enables packaging of `arm64` images for calico, multus and harvester.
Also removes the deprecated `cilium-etcd-operator`

Linked issues:
* https://github.com/rancher/rke2/issues/5136
* https://github.com/rancher/rke2/issues/5260

### DIFF
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -33,12 +33,11 @@ xargs -n1 -t docker image pull --quiet << EOF > build/images-canal.txt
     ${REGISTRY}/rancher/hardened-flannel:v0.23.0-build20231109
 EOF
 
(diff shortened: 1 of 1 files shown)


brandond (approved):

thomasferrandiz (approved):

manuelbuil (approved):

brandond (changes requested):
nit: this is also enabling arm64 for calico, multus, vsphere, and harvester. Either make sure that those all exist, or keep them excluded from this arch.

Also, if you keep those included, update the PR title and description to make it clear that we're doing more than just cilium here.

brandond (approved):

PR #5172, opened 2024-01-02
update channels

briandowns:
#### User-Facing Change ####

```release-note

```


VestigeJ (approved):

cwayne18:
LGTM, should we add a 1.29 channel since its out or would we want that to be a separate PR

briandowns:
I'd prefer a separate pr for that change.

Issue #5156, opened 2023-12-21
Generated node local coredns config has extra newline and is invalid


bmhughes:
**Environmental Info:**
RKE2 Version: 
```
rke2 version v1.27.8+rke2r1 (77c9470934d7073341fb297aefa2dda0b97909c9)
go version go1.20.11 X:boringcrypto
```

Node(s) CPU architecture, OS, and Version: 

```
Linux 5.14.0-362.8.1.el9_3.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Nov 8 17:36:32 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

Cluster Configuration:

**Describe the bug:**

When enabling node local DNS and invalid configuration is generated with an additional newline after the `forward` option which is rejected as an invalid config by coredns.

Issue appears to be [here](https://github.com/rancher/rke2-charts/blob/c05387b26272b07b7ed95f8564e9a4d22ea65f55/packages/rke2-coredns/generated-changes/patch/templates/_helpers.tpl.patch#L38) as the addition of a `-` to strip in the newline fixes the problem, might be something do with how the `split` function returns as it does not occur with the static definition of `10.43.0.10` nor the previous behaviour in [4811c82](https://github.com/rancher/rke2-charts/blob/4811c82e4b1a0015b91b2ca2757d8db166959dd9/packages/rke2-coredns/generated-changes/patch/templates/_helpers.tpl.patch#L28)

**Steps To Reproduce:**

- Install RKE2
- Enabled node local DNS

**Expected behavior:**
It works

**Actual behavior:**
An invalid config is generated and coredns is stuck in `CrashLoopBackoff` until the `ConfigMap` is manuallly fixed

**Additional context / logs:**

brandond:
Is this on a dual-stack cluster? Are you overriding the ClusterDNS address? This would only appear to be a problem when `.Values.global.clusterDNS` contains multiple entries, which I don't believe it would by default even on a dual-stack cluster.

bmhughes:
Apologies, my mistake leaving that out. Yes, this is a dual-stack cluster and the  `.Values.global.clusterDN` has both the v4 and v6 address present.

brandond:
Ok. And you manually set it to both addresses? It works fine if you leave it as the default single-track value?

bmhughes:
[...]

PR #5190, opened 2024-01-03
Add e2e test for dnscache

manuelbuil:
#### Proposed Changes ####

Add an e2e test about dnscache to avoid problems like this one https://github.com/rancher/rke2/issues/5156.
Using dnscache is not something common, so we could introduce a regression and not notice it. We should have an e2e test

#### Types of Changes ####

New e2e test

#### Verification ####

go test -v -timeout=20m ./tests/e2e/dnscache/

Note that currently it fails because of: https://github.com/rancher/rke2/issues/5156

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5191
#### User-Facing Change ####

```release-note

```


dereknola (approved):
Is this getting merged after https://github.com/rancher/rke2/pull/5189 merges?

brandond (approved):

manuelbuil:
> Is this getting merged after #5189 merges?

Yes! Because otherwise it will fail :)

Issue #5247, opened 2024-01-11
In windows, kube-proxy does not log the error and calico logs are in the wrong path


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Any

Node(s) CPU architecture, OS, and Version: 

One Windows agent

Cluster Configuration:

1 windows agent

**Describe the bug:**

If something goes wrong in kube-proxy.exe process, by checking the windows system logs, you will see the following output:
```
kube-proxy exited: exit status 1
```
which is useless to understand what is going on.

The problem is that these lines don't apparently work well in windows:
https://github.com/rancher/rke2/blob/master/pkg/pebinaryexecutor/pebinary.go#L225C8-L226
and they don't provide anything in the logging windows system

Then, calico logs are under /var/log/. It makes more sense that they are under `C:\var\lib\rancher\rke2\agent\logs`

**Steps To Reproduce:**

- Installed RKE2: 

I don't exactly know how to produce a `kube-proxy.exe` error with the current rke2 binary. This is how it happened to me:

1 - Add on extra arg here https://github.com/rancher/rke2/blob/master/pkg/pebinaryexecutor/pebinary.go#L225C8-L226. An arg which is wrong. For example:
```
	extraArgs := map[string]string{
		"network-name": p.CNIConfig.CNICfg.OverlayNetName,
		"bind-address": p.CNIConfig.CNICfg.IP,
                "log-file": "C://var//log//dummy.log"
	}
```
2 - Build rke2.exe `make build-windows-binary`
3 - Execute rke2.exe in the node-agent as a service `C:\usr\local\bin\rke2.exe agent service --add` and `Start-Service -Name rke2`
4 - By checking the windows logs, you'll see `kube-proxy exited: exit status 1` and have no idea what is going on

Check that calico logs and kube-proxy logs are in `C:\var\lib\rancher\rke2\agent\logs`
```
PS C:\Users\Administrator> ls C:\var\lib\rancher\rke2\agent\logs

    Directory: C:\var\lib\rancher\rke2\agent\logs

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           1/12/2024  2:39 AM           8402 calico-node.log
-a---           1/12/2024  2:39 AM          76379 felix.log
-a---           1/12/2024  2:39 AM              0 kube-proxy.log
-a---           1/12/2024  2:21 AM       51485306 kubelet.log
```

**Expected behavior:**

To know what is going on so that you can fix it

[...]

PR #5248, opened 2024-01-11
[windows] Improve kube-proxy logging and move calico logs to a better path

manuelbuil:
#### Proposed Changes ####

Following on from what we do with Calico processes, use a file to dump the logs of kube-proxy in windows. The current Stdout and Stderr output is supposed to be piped to the Windows Logging System but it does not work. Using a file instead works.

This PR also moves Calico logs to the same place where kubelet logs are. That place is `C:\var\lib\rancher\rke2\agent\logs` which makes much more sense

#### Types of Changes ####

Bugfix

#### Verification ####

Check issue

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5247

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/pebinaryexecutor/pebinary.go
+++ b/pkg/pebinaryexecutor/pebinary.go
@@ -220,10 +220,11 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
 
 	logrus.Infof("Running RKE2 kube-proxy %s", args)
 	go func() {
+		outputFile := logging.GetLogger(filepath.Join(p.DataDir, "agent", "logs", "kube-proxy.log"), 50)
 		for {
 			cmd := exec.CommandContext(ctx, filepath.Join("c:\\", p.DataDir, "bin", "kube-proxy.exe"), args...)
-			cmd.Stdout = os.Stdout
-			cmd.Stderr = os.Stderr
+			cmd.Stdout = outputFile
+			cmd.Stderr = outputFile
 			err := cmd.Run()
 			logrus.Errorf("kube-proxy exited: %v", err)
 			time.Sleep(5 * time.Second)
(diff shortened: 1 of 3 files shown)


pkg/pebinaryexecutor/pebinary.go
```diff
@@ -67,6 +67,8 @@ const (
 	CNICalico = "calico"
 	CNICilium = "cilium"
 	CNICanal  = "canal"
+	LogPath   = "C:\\var\\log\\"
```
brandond, line 70: The standard path for linux is `/var/lib/rancher/rke2/agent/logs/kubelet.log`, why do we use `c:/var/log` on windows instead of `c:/var/lib/rancher/rke2/agent/logs` ?

manuelbuil: AH! I missed that already existing path! Thanks Brad! I'm actually going to move calico logs to there as well, I think it makes much more sense

thomasferrandiz (approved):

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -220,11 +220,16 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
 
 	logrus.Infof("Running RKE2 kube-proxy %s", args)
 	go func() {
+		outputFile, err := os.Create(filepath.Join(p.DataDir, "agent", "logs", "kube-proxy.log"))
```
brandond, line 223: I would recommend reusing or at least duplicating the code that we use to capture kubelet logs using lumberjack, which handles rotating and pruning logs. Otherwise this one log file may get very large.

https://github.com/rancher/rke2/blob/f2396b6ea0a97b568a46b0fd754bac042e4ce854/pkg/logging/logging.go#L38-L40

brandond (changes requested):

pkg/pebinaryexecutor/pebinary.go
manuelbuil: good idea! I was actually wondering if these logs could get too large. I was not aware of this lumberjack library

brandond (approved):

Issue #5275, opened 2024-01-15
Refactor windows CNI code to simplify the support of new CNI plugins


manuelbuil:
**Is your feature request related to a problem? Please describe.**

Related to https://github.com/rancher/rke2/issues/5215, we need to abstract how pebinary talks to the CNI plugin. Right now, as Calico is the only CNI plugin, there is a lot of Calico specific stuff

**Describe the solution you'd like**

* An interface that each CNI plugin can honor
* A common config struct
 Both objects could be used by pebinary without having to know details on the CNI plugin

**Describe alternatives you've considered**

**Additional context**

rancher-max:
A few issues around this are similar. The code went through a few iterations and required some changes tracked by this, but I'm going to close this one in favor of the Calico and Flannel Windows issues directly.

Issue #5215, opened 2024-01-08
Add flannel as CNI plugin


manuelbuil:
**Is your feature request related to a problem? Please describe.**

We currently don't include flannel as CNI plugin. Flannel could be a good addition because of mainly two reasons:
* It supports Windows, which gives an alternative to Windows users. Right now, only Calico is supported in rke2-windows
* It gives a direct migration path to RKE1 users that chose Flannel as the CNI plugin

Note that Flannel does not include network policy controller, which could be a big limitation for users

**Describe the solution you'd like**

Be able to deploy Flannel with rke2 as we do with other CNI plugins:
* In linux, by installing a helm chart included in repo rke2-charts
* In Windows, by executing flanneld as part of the rke2 binary. Same way of executing calico in Windows

**Describe alternatives you've considered**

**Additional context**

fmoral2:
## Validated on Version:
```console
-$  rke2 version v1.29.2+rke2r1

 
```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
SUSE Linux Enterprise Server 15 SP4   


*Cluster Configuration:*
1 server node 
1 agent node 


 

 
## Steps to validate the fix
1. Install rke2 with cni: flannel
2. Validate it works as expected
 
 

 
 
## **Reproduction Issue:**
<details>

``` 

 
 
```

 
## **Validation Results:**
<details>


``` 
      
  $ rke2 -v 
v1.29.2+rke2r1

 
09:17:07.542   .us-east-2.compute.internal   Ready   control-plane,etcd,master   56s   v1.29.2+rke2r1      SUSE Linux Enterprise Server 15 SP4   5.14.21-150400.22-default   containerd://1.7.11-k3s2


 
09:17:07.542  PASSED! Test: Validate Nodes
09:18:44.620  [38;5;10m•[0m
09:18:44.620  Cluster pods:
09:18:44.620   kube-system   cloud-controller-manager-0.us-east-2.compute.internal   1/1   Running     0     2m15s   0   0.us-east-2.compute.internal   <none>   <none>
09:18:44.620  kube-system   etcd-0.us-east-2.compute.internal                       1/1   Running     0     116s    0   0.us-east-2.compute.internal   <none>   <none>
09:18:44.620  kube-system   helm-install-rke2-coredns-tmg4z                                       0/1   Completed   0     116s    0   0.us-east-2.compute.internal   <none>   <none>
[...]

PR #5276, opened 2024-01-15
[Windows] Create a common CNI interface and config struct

manuelbuil:
#### Proposed Changes ####

As rke2-windows will have two CNI plugin options: calico and flannel, we should abstract the current code so that it gets easier and cleaner to include flannel. This PR tries to achieve that.

Changes:
* CniPlugin interface which "win.Calico" struct honors
* CNICommonConfig struct with properties that are shared across CNIplugins and can be important to configure Kubernetes components (e.g. Kube-proxy)
* PEBinary struct now uses CniPlugin interface instead of "win.Calico"
* A few name changes, which I think are clearer, e.g. CNI ==> CNIName, getCniType ==> getCniPluginName

This is part of the Flannel PR I am preparing but that PR is getting too large, hence I decided to split it into a couple of PRs to simplify reviewing and reduce the risk of introducing bugs

#### Types of Changes ####

New Feature

#### Verification ####

This is basically a refactoring of code. It does not introduce any new feature. Therefore, what would be great is a verification that rke2-windows-calico is still working and no regression was introduced by this PR

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5275
https://github.com/rancher/rke2/issues/5215

#### User-Facing Change ####

```release-note

```


pkg/windows/calico.go
```diff
@@ -60,11 +60,11 @@ users:
 	calicoConfigTemplate = template.Must(template.New("CalicoConfig").Funcs(replaceSlashWin).Parse(`{
   "name": "{{ .Name }}",
   "windows_use_single_network": true,
-  "cniVersion": "{{ .CNI.Version }}",
+  "cniVersion": "0.3.1",
```
thomasferrandiz, line 63: is it necessary to hard-code the CNI version?

manuelbuil: No, it is not. I don't know why I thought it was a good idea :). Let me add it as part of the config

rbrtbnfgl (approved):

dereknola (approved):

brandond (approved):

Issue #5335, opened 2024-01-30
Race condition when rke2-windows-calico removes and creates an HNS network


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

An HNS overlay network is always bound to a physical interface. When that network is created or deleted, the IPs of that interface "disappear" for some seconds. It's hard to say for how many seconds but according to my experience, it is something in between 6 and 20 seconds.

When the windows node is starting, we generate all the required config files for Calico CNI plugin and create the "External" Calico network, which is the one with the vxlan overlay. However, before creating that network, we delete all existing networks:

https://github.com/rancher/rke2/blob/master/pkg/windows/calico.go#L284

The `deleteAllNetworks()` function: https://github.com/rancher/rke2/blob/master/pkg/windows/utils.go#L99, applies first a "GET" query and then removes all the returned networks. Immediately afterwards, the code creates the network.  If the interface was not specified, it searches for the interface with the nodeIP: https://github.com/rancher/rke2/blob/master/pkg/windows/calico.go#L301. It is possible that when the search for the interface happens, the interface is still unavailable due to the deletion of the HNS overlay network.

As a consequence, calico fails to start and we get the error:
```
no interface has the ip: a.b.c.d
```

Internal reference:  SURE-7509

**Steps To Reproduce:**

- Installed RKE2: 

**Expected behavior:**

When we create the network we are 100% that the interfaces are ready

**Actual behavior:**

We are not checking that interfaces are ready and that can cause problems when creating the network

**Additional context / logs:**

ShylajaDevadiga:
### Validated using rke2 version v1.29.2-rc1+rke2r1
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 22.04
Windows Server 2022

Steps to Reproduce and Validate following the steps in the PR
```
1 - Deploy rke2 server with calico
2 - Deploy rke2-agent on windows
3 - Once everything is up, Stop-Service rke2 and C:\usr\local\bin\rke2.exe agent service --delete
4 - Verify that there is at least one HNS Network: get-hnsnetwork
5 - Start the rke2-agent on windows again with debug: true (remember to remove the node first or it will complain about password already there)

You should at least see the messages:

 Deleting network: XXXXXX before starting calico"
And

[...]

PR #5336, opened 2024-01-30
Avoid race condition when deleting HNS networks

manuelbuil:
#### Proposed Changes ####

Check that interfaces are back if the HNS network was deleted before moving to HNS Network creation

#### Types of Changes ####

Bugfix

#### Verification ####

This is a race condition that really depends on how fast your windows system is able to "recover" the interface after deleting the HNS Network. For me it was impossible to reproduce. However, a basic verification would be:

1 -  Deploy rke2 server with calico
2 - Deploy rke2-agent on windows
3 - Once everything is up, `Stop-Service rke2` and `C:\usr\local\bin\rke2.exe agent service --delete`
4 -  Verify that there is at least one HNS Network: `get-hnsnetwork`
5 - Start the rke2-agent on windows again with `debug: true` (remember to remove the node first or it will complain about password already there)

You should at least see the messages:
```
 Deleting network: XXXXXX before starting calico"
```
And
```
Calico is waiting for the interface with ip: XXXXXX to come back
```
And it should finally succeed to connect to the server

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5335

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/windows/calico.go
+++ b/pkg/windows/calico.go
@@ -295,7 +295,7 @@ func (c *Calico) Start(ctx context.Context) error {
 // generateCalicoNetworks creates the overlay networks for internode networking
 func (c *Calico) generateCalicoNetworks() error {
 	if err := deleteAllNetworks(); err != nil {
-		return err
+		return errors.Wrapf(err, "failed to delete all networks before bootstrapping calico")
 	}
 
 	// There are four ways to select the vxlan interface. In order of priority:
--- a/pkg/windows/utils.go
+++ b/pkg/windows/utils.go
@@ -19,6 +19,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	opv1 "github.com/tigera/operator/api/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // createHnsNetwork creates the network that will connect nodes and returns its managementIP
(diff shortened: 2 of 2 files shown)


pkg/windows/calico.go
```diff
@@ -283,7 +283,7 @@ func (c *Calico) Start(ctx context.Context) error {
 // generateCalicoNetworks creates the overlay networks for internode networking
 func (c *Calico) generateCalicoNetworks() error {
 	if err := deleteAllNetworks(); err != nil {
-		return err
+		return fmt.Errorf("error deleting all networks before bootstrapping calico: %v", err)
```
brandond, line 286: For wrapping errors, the convention in this file seems to be:
```suggestion
		return errors.Wrap(err, "failed to delete all networks before bootstrapping calico")
```

brandond (approved):

pkg/windows/utils.go
```diff
@@ -101,15 +102,35 @@ func deleteAllNetworks() error {
...
+			ips = append(ips, network.ManagementIP)
 			_, err = network.Delete()
 			if err != nil {
 				return err
 			}
 		}
 	}
 
+	// HNS overlay networks restart the physical interface when they are deleted. Wait until it comes back before returning
+	waitErr := wait.Poll(2000*time.Millisecond, 20*time.Second, func() (bool, error) {
```
thomasferrandiz, line 119: it seems `wait.Poll` is deprecated: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Poll

thomasferrandiz: also is it possible to use `2*time.Second` to be coherent with `20*time.Second`?

pkg/windows/calico.go
manuelbuil: There are both options present in this file... let me change all to errors.Wrap

pkg/windows/utils.go
manuelbuil: good point, that makes it more readable

manuelbuil: re: deprecation. We discussed about this with Brad in another PR https://github.com/k3s-io/k3s/pull/9076#discussion_r1434397387, I'll add a comment to the code

brandond: The replacement `[PollUntilContextTimeout](https://github.com/kubernetes/apimachinery/blob/v0.29.1/pkg/util/wait/poll.go#L45)` was only added in 1.27, so we should continue to use the deprecated functions until 1.26 is end of life. Once we are no longer maintaining 1.26, we can switch to the new 1.27+ functions and backport that change to all the active branches.

brandond: Created an issue to track: https://github.com/k3s-io/k3s/issues/9321

thomasferrandiz: Thanks for the explanation

thomasferrandiz (approved):

Issue #2204, opened 2021-12-02
[Windows] Stop-Service rke2 does not stops all rke2 services


mdrahman-suse:
Extension of (https://github.com/rancher/rke2/issues/1755) 

**Environmental Info:**
RKE2 Version: 

rke2.exe version v1.22.4-rc2+rke2r1 (79dc33a7f4329be00dad7c79d559fc456264a2b9)
go version go1.16.10b7

Node(s) CPU architecture, OS, and Version: 

Server: Linux ip-172-31-47-190 5.4.0-1009-aws #9-Ubuntu SMP Sun Apr 12 19:46:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Agent: Windows 10 (1809)

Cluster Configuration:

1 server, 1 agent

**Describe the bug:**

`Stop-Service rke2` does not stop all the rke2 services cleanly. One or two services remains in running state randomly
```
PS C:\Users\Administrator> Stop-Service rke2
PS C:\Users\Administrator> Get-Process kubelet,kube-proxy,containerd,rke2
Get-Process : Cannot find a process with the name "kubelet". Verify the process name and call the cmdlet again.
At line:1 char:1
+ Get-Process kubelet,kube-proxy,containerd,rke2
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (kubelet:String) [Get-Process], ProcessCommandException
    + FullyQualifiedErrorId : NoProcessFoundForGivenName,Microsoft.PowerShell.Commands.GetProcessCommand

Get-Process : Cannot find a process with the name "kube-proxy". Verify the process name and call the cmdlet again.
At line:1 char:1
+ Get-Process kubelet,kube-proxy,containerd,rke2
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (kube-proxy:String) [Get-Process], ProcessCommandException
    + FullyQualifiedErrorId : NoProcessFoundForGivenName,Microsoft.PowerShell.Commands.GetProcessCommand

Get-Process : Cannot find a process with the name "rke2". Verify the process name and call the cmdlet again.
At line:1 char:1
+ Get-Process kubelet,kube-proxy,containerd,rke2
[...]

PR #5419, opened 2024-02-09
Implement custom `containerd` behavior for Windows Agents, ensure supporting processes exit

HarrisonWAffel:
#### Proposed Changes ####

Update the `pebinary` and `staticpod` executors to conform to the updated k3s executor interface which enables greater control over how `containerd` and `docker` are managed. 

The `pebinary` executor partially reimplements the logic for starting `containerd` with one minor change: If the `containerd` process is killed for any reason, `os.Exit` will _not_ be invoked and instead `containerd` will be restarted after 5 seconds. This is in line with how we start and restart other core components within the windows agent, such as the kubelet. 

This change ensures that the rke2 agent process and the Windows service do not prematurely exit alongside `containerd`, giving enough for the [`signals.Context`](https://github.com/k3s-io/k3s/blob/fa11850563064d4beb68e29baf7526fe3de6bb5c/pkg/cli/agent/agent.go#L90) cancellation to fully propagate and stop all processes spawned from rke2/k3s. 

In order to better coordinate how the RKE2 windows service shuts down, additional logic has been added to monitor processes spawned by RKE2 for up to ten seconds. This ensures that the agent process does not exit until all other processes have been cleaned up, or the time limit is reached. 

The `docker` implementation for Windows agents has not been changed and the `pebinary` executor  simply invokes the k3s implementation for `docker`.

The linux `staticpod` executor invokes the existing implementation within k3s for `containerd` and `docker`, resulting in no behavior changes for linux agents. 

#### Types of Changes ####

Updates to the `pebinary` executor, `staticpod` executor, and service_windows

#### Verification ####

This can be verified by joining an RKE2 windows node to an existing cluster, and manually stopping the rke2 service using Powershell, ensuring that all rke2 processes stop at the same time. I have done this verification manually and saw no issues with starting, stopping, and deploying workloads using `containerd`. 

#### Testing ####

The existing end to end testing which covers mixed OS clusters (`tests/e2e/mixedos/mixedos_test.go`) already tests that the windows agent starts and behaves as expected. I'm happy to help expand the e2e tests to cover this specific scenario if we feel it's necessary.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/2204

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

Since this change uses the updated executor interface from k3s, I've bumped the k3s version to the latest commit on master ([`cfc3a124eed6`](https://github.com/k3s-io/k3s/commit/cfc3a124eed64ca5bd349bafea97afba56337fb0)) at the time of raising this PR.


pkg/windows/service_windows.go
```diff
@@ -79,5 +80,39 @@ func StartService() error {
 		}
 	}()
 
+	return true, nil
+}
+
+// MonitorProcessExit ensures that the kubelet, kube-proxy, calico-node, and containerd processes have stopped running.
```
brandond, line 86: This is an interesting way to do this. I'm not a big fan of inline powershell and shelling out to do things that we should be able to do from native golang.

Can we not use a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) when starting these processes through the PEBinaryExecutor, and just wait on that when shutting down?

dereknola:
I would love to see the E2E test expanded, since you already have a manual verification method.

pkg/windows/service_windows.go
HarrisonWAffel: A bit of face palm moment on my side -  a WaitGroup is a much better idea. After looking into it a bit, I found `apimachinery/pkg/util/wait` which offers wait groups that utilize contexts. Spent some time Friday and this morning testing the changes out and everything works as desired without any inlined PowerShell. 

Let me know what you think. Also, I wanted to get your opinion on if it would be a good idea to add a timeout for the `WaitGroup.Wait` call. I can't think of a situation where one of these commands would continue to run after the context cancelation, but I want to avoid a situation where we might wait forever for the WaitGroup to complete.

HarrisonWAffel:
@dereknola Is this something that could be done as a follow up PR? My understanding is that code freeze for Feb patches is EOD, and so far I've had trouble setting up the e2e test suite on an ubuntu VM

dereknola:
It can happen in another PR. Two things for that:
1) Open an issue to track adding the E2E test
2) Expand the current `Verification` section of this PR with more info/commands on how you verified the changes.

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -212,22 +223,110 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
...
+// Docker starts cri-dockerd as implemented in the k3s cridockerd package
+func (p *PEBinaryConfig) Docker(ctx context.Context, cfg *config.Node) error {
+	return cridockerd.Run(ctx, cfg)
+}
+
+// Containerd configures and starts containerd.
+func (p *PEBinaryConfig) Containerd(ctx context.Context, cfg *config.Node) error {
+	args := getContainerdArgs(cfg)
+	stdOut := io.Writer(os.Stdout)
+	stdErr := io.Writer(os.Stderr)
```
manuelbuil, line 257: Did you manage to get containerd logs in Windows when using this? I had a lot of trouble and was unable to do it, so by default I decided to dump logs in files (kube-proxy, etc)

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -212,22 +223,110 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
...
+				// Turn CONTAINERD_LOG_LEVEL variable into log-level flag
+				args = append(args, "--log-level", pair[1])
+			case strings.HasPrefix(pair[0], "CONTAINERD_"):
+				// Strip variables with CONTAINERD_ prefix before passing through
+				// This allows doing things like setting a proxy for image pulls by setting
+				// CONTAINERD_https_proxy=http://proxy.example.com:8080
+				pair[0] = strings.TrimPrefix(pair[0], "CONTAINERD_")
+				cenv = append(cenv, strings.Join(pair, "="))
+			default:
+				env = append(env, strings.Join(pair, "="))
```
manuelbuil, line 299: shouldn't it only check on `CONTAINERD_` envs?

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -212,22 +223,110 @@ func (p *PEBinaryConfig) KubeProxy(ctx context.Context, args []string) error {
...
+			}
+		}
+
+		for {
+			logrus.Infof("Running containerd %s", config.ArgString(args[1:]))
```
manuelbuil, line 304: It might worth logging the env variales too, that's what we do in Calico to help debugging. If we are picking all of them, probably it makes more sense to log it only in Debug=true scenarios

pkg/pebinaryexecutor/pebinary.go
HarrisonWAffel: Yes, I was able to find the logs in `C:\var\lib\rancher\rke2\agent\containerd\containerd.log`

pkg/pebinaryexecutor/pebinary.go
HarrisonWAffel: This is a mirror implementation of what k3s does, I'm not sure what the implications are if start ignoring specific environment variables

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -160,9 +165,9 @@ func (p *PEBinaryConfig) Kubelet(ctx context.Context, args []string) error {
 		cleanArgs = append(cleanArgs, arg)
 	}
 
-	logrus.Infof("Running RKE2 kubelet %v", cleanArgs)
-	go func() {
+	win.ProcessWaitGroup.StartWithContext(ctx, func(ctx context.Context) {
```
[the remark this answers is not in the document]
HarrisonWAffel: The initial implementation of `pebinary` had access to this context as well. iiuc, using a separate context ensures that the CNI plugin is restarted each time that the kubelet is restarted, which may happen before the global context is canceled (for example, if someone accidentally runs  `Stop-Process` for the kubelet). Since this PR does not focus on how the kubelet starts and stops the CNI plugins I opted to leave this unchanged.

pkg/pebinaryexecutor/pebinary.go
HarrisonWAffel: I believe I have addressed this

pkg/podexecutor/staticpod.go
```diff
@@ -614,6 +616,16 @@ func (s *StaticPodConfig) ETCD(ctx context.Context, args executor.ETCDConfig, ex
 	return staticpod.Run(s.ManifestsDir, spa)
 }
 
+// Containerd starts the k3s implementation of containerd
+func (s *StaticPodConfig) Containerd(ctx context.Context, config *daemonconfig.Node) error {
+	return containerd.Run(ctx, config)
+}
+
+// Docker starts the k3s implementation of cridockerd
```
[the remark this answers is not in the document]
brandond: I think this is correct, the comment is just calling out that all this does is wrap the k3s implementation.

pkg/pebinaryexecutor/pebinary.go
brandond: I would avoid logging env vars, using env vars to pass through secrets or credentials since they don't get logged or show in `ps` output, and that is a pattern that we want to continue to enable.

HarrisonWAffel: Ah thats a good point, I'll update this

dereknola (approved):

pkg/windows/service_windows.go
brandond: No I think this is good. If we're blocking on child process exit and want to be sure they get cleaned up, I think blocking shutdown indefinitely is appropriate.

brandond (approved):

pkg/cli/cmds/agent.go
```diff
@@ -88,8 +88,15 @@ func agentSubcommands() cli.Commands {
...
 	validateCloudProviderName(clx, Agent)
 	validateProfile(clx, Agent)
-	if err := windows.StartService(); err != nil {
+	isWinService, err := windows.StartService()
+	if err != nil {
 		return err
 	}
-	return rke2.Agent(clx, config)
+
+	err = rke2.Agent(clx, config)
```
briandowns, line 96: If this is in error state, what is the expected behavior? SHould it continue to process?

Issue #5415, opened 2024-02-09
Improve Calico in Windows


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

Calico was unable to create endpoints correctly because the config had the wrong name. It had "calico" instead of "Calico". As a consequence, pods were never starting.

Apart from this, it is a good practice to follow along what flannel does and check that the node exists in kube-api before starting Calico. This way, we can avoid potential problems.

**Steps To Reproduce:**

- Installed RKE2:
Install RKE2 with this PR https://github.com/rancher/rke2/pull/5322 and deploy with Calico CNI.

**Expected behavior:**

Pods start

**Actual behavior:**

Pods do not start

**Additional context / logs:**

ShylajaDevadiga:
### Validated using latest commit id on master branch 095d861faf4d311bacbb7f554b444432ae1b1232

## Environment Details
*Infrastructure*
Cloud EC2 instance

**Using cni: calico**
```
ubuntu@ip-172-31-11-57:~$ rke2 -v
rke2 version v1.29.2+dev.095d861f (095d861faf4d311bacbb7f554b444432ae1b1232)
go version go1.21.7 X:boringcrypto
ubuntu@ip-172-31-11-57:~$ cat /etc/rancher/rke2/config.yaml
write-kubeconfig-mode: "0644"
tls-san:
  - fake.fqdn.value
node-name: ip-172-31-11-57.us-east-2.compute.internal
cni: calico

node-external-ip: <REDACTED>
node-ip: 172.31.11.57
ubuntu@ip-172-31-11-57:~$ kubectl get nodes
NAME                                         STATUS   ROLES                       AGE    VERSION
ip-172-31-11-57.us-east-2.compute.internal   Ready    control-plane,etcd,master   106m   v1.29.2+rke2r1
ip-172-31-4-8.us-east-2.compute.internal     Ready    <none>                      103m   v1.29.2+rke2r1
ip-ac1f2610                                  Ready    <none>                      99m    v1.29.2

ubuntu@ip-172-31-11-57:~$ kubectl describe node ip-ac1f2610|grep Windows
  OS Image:                   Windows Server 2022 Datacenter

ubuntu@ip-172-31-11-57:~$ kubectl get pods -A
NAMESPACE         NAME                                                                  READY   STATUS      RESTARTS      AGE
[...]

PR #5421, opened 2024-02-12
Improve cni windows code

manuelbuil:
#### Proposed Changes ####

This PR does a few things:
1 - Fixes a bug in calico which prevented pods from starting (config should have "Calico" instead of "calico")
2 - Increases time out to 30s. I saw twice during my tests that sometimes 20s is not enough for the interfaces to come back in Windows
3 - Similarly to Flannel, before starting Calico, we check if the node is present in kube-api to avoid potential issues
4 - Instead of using an infinite loop to start calico, we use `wait.PollImmediateWithContext` with a time out of 5 minutes
5 - Flannel's `wait.PollImmediateWithContext` now return an error

#### Types of Changes ####

Bugfix

#### Verification ####

Just verify that rke2 still works with flannel and calico

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5415

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/windows/calico.go
+++ b/pkg/windows/calico.go
@@ -116,7 +116,8 @@ users:
 )
 
 type Calico struct {
-	CNICfg *CalicoConfig
+	CNICfg     *CalicoConfig
+	KubeClient *kubernetes.Clientset
 }
 
 const (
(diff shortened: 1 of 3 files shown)


thomasferrandiz (approved):

rbrtbnfgl (approved):

manuelbuil:
I have just tested this deploying with Calico and Flannel and it works :)

PR #5466, opened 2024-02-14
[release-1.27] Update k8s and Go

briandowns:
#### User-Facing Change ####

```release-note

```


tashima42 (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/rke2
 
-go 1.19
+go 1.21
+
+toolchain go1.22.0
```
brandond, line 5: golang REALLY wants to do this now, but don't let it. If your system golang is 1.22+ you now need export GOTOOLCHAIN in order to avoid having golang mangle your go.mod:

`GOTOOLCHAIN=go1.20.13 go mod tidy`

where `go1.20.13` is whatever version of golang we're going to use at build time.

brandond (changes requested):

galal-hussein (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/rke2
 
-go 1.19
+go 1.21
+
+toolchain go1.21.7
```
```suggestion
go 1.19
```

k3s is still on `go 1.19` for this branch.

Remove the `toolchain` line and then run the tidy with `GOTOOLCHAIN=$(. ./scripts/version.sh &>/dev/null; echo $VERSION_GOLANG) go mod tidy` and it will prevent go1.22 from adding this unwanted option. We're still building with golang <1.22 so there is no need for 1.22 specific changes on this branch.

brandond (changes requested):

go.mod
briandowns: 1.26 branch go.mod is 1.19. Not this one. 1.27 is 1.20 going to 1.21.

brandond (approved):

brandond:
Why does this one work without the tidy adding the toolchain entry when 1.28 and 1.29 do not?

PR #5467, opened 2024-02-14
[release-1.28] Update k8s and Go

briandowns:
#### User-Facing Change ####

```release-note

```


tashima42 (approved):

brandond (changes requested):

galal-hussein (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/rke2
 
-go 1.20
+go 1.21
+
+toolchain go1.22.0
```
```suggestion
go 1.21
```

brandond: Remove the `toolchain` line and then run the tidy with `GOTOOLCHAIN=$(. ./scripts/version.sh &>/dev/null; echo $VERSION_GOLANG) go mod tidy` and it will prevent go1.22 from adding this unwanted option. We're still building with 1.21 so there is no need for 1.22 specific changes on this branch.

brandond (changes requested):

brandond (approved):

go.mod
```suggestion
toolchain go1.21.7
```

brandond (changes requested):

brandond (approved):

PR #5468, opened 2024-02-14
[release-1.26] Update k8s and Go

briandowns:
#### User-Facing Change ####

```release-note

```


tashima42 (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/rke2
 
-go 1.19
+go 1.21
+
+toolchain go1.22.0
```
brandond, line 5: https://github.com/rancher/rke2/pull/5466#discussion_r1489957949

brandond (changes requested):

galal-hussein (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/rke2
 
-go 1.19
+go 1.21
+
+toolchain go1.21.7
```
```suggestion
go 1.19
```

k3s is still on `go 1.19` for this branch.

Remove the `toolchain` line and then run the tidy with `GOTOOLCHAIN=$(. ./scripts/version.sh &>/dev/null; echo $VERSION_GOLANG) go mod tidy` and it will prevent go1.22 from adding this unwanted option. We're still building with golang <1.22 so there is no need for 1.22 specific changes on this branch.

brandond (changes requested):

brandond (changes requested):
(ノಠ益ಠ)ノ彡┻━┻

Sorry for wasting your time.

I guess there's no getting around the toolchain changes with golang 1.21. I'm not sure how upstream is getting away with leaving their go.mod at older versions without having tidy change it. I guess starting from now we'll just have to keep the toolchain version in sync with everything else.

brandond (approved):

Issue #5315, opened 2024-01-23
Network Policy rke2-flannel-host-networking when cis-1.23 and calico


albundy83:
**Environmental Info:**
RKE2 Version:
```bash
v1.26.12+rke2r1
```

Node(s) CPU architecture, OS, and Version:
```bash
outscale@ip-10-0-1-10:~$ uname -a
Linux ip-10-0-1-10 5.15.0-91-generic #101-Ubuntu SMP Tue Nov 14 13:30:08 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

Cluster Configuration:
3 servers, 3 agents

Installed with lablabs ansible-role release [1.28.0](https://github.com/lablabs/ansible-role-rke2/releases/tag/1.28.0)

**Describe the bug:**
When I enable `cis-1.23` profile and use `calico` as cni, I have the following network policy that is created in namespace `kube-system` `kube-public` and `default`

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: rke2-flannel-host-networking
  namespace: kube-system
spec:
  podSelector: {}
  ingress:
    - {}
  policyTypes:
    - Ingress
  ```

**Steps To Reproduce:**

- Installed RKE2: 
```yaml
---
# The node type - server or agent
rke2_type: server

# Deploy the control plane in HA mode
rke2_ha_mode: true

# Install and configure Keepalived on Server nodes
# Can be disabled if you are using pre-configured Load Balancer
rke2_ha_mode_keepalived: false

# Install and configure kube-vip LB and VIP for cluster
# rke2_ha_mode_keepalived needs to be false
rke2_ha_mode_kubevip: false

# Kubernetes API and RKE2 registration IP address. The default Address is the IPv4 of the Server/Master node.
# In HA mode choose a static IP which will be set as VIP in keepalived.
# Or if the keepalived is disabled, use IP address of your LB.
rke2_api_ip: my-youpi.com

# optional option for RKE2 Server to listen on a private IP address on port 9345
# rke2_api_private_ip:

# optional option for kubevip IP subnet
# rke2_api_cidr: 24

# optional option for kubevip
# rke2_interface: eth0
# optional option for IPv4/IPv6 addresses to advertise for node
# rke2_bind_address: "{{ hostvars[inventory_hostname]['ansible_' + rke2_interface]['ipv4']['address'] }}"

# kubevip load balancer IP range
rke2_loadbalancer_ip_range: {}
[...]

PR #5500, opened 2024-02-21
Add new network policy for ingress controller webhook

brandond:
#### Proposed Changes ####

Add new network policy for ingress controller webhook.

The ingress-nginx chart added a validating webhook a while back, but we did not have a policy that allowed access to it. Access was being allowed by the overly broad network policy that we removed in https://github.com/rancher/rke2/pull/5318.

Unfortunately we can't just add this to the existing netpol, as we document that it is not updated once the namespace annotation has been set.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/5315#issuecomment-1957853824

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/rke2/np.go
+++ b/pkg/rke2/np.go
@@ -24,10 +24,12 @@ const (
 	namespaceAnnotationNetworkPolicy        = "np.rke2.io"
 	namespaceAnnotationNetworkDNSPolicy     = "np.rke2.io/dns"
 	namespaceAnnotationNetworkIngressPolicy = "np.rke2.io/ingress"
+	namespaceAnnotationNetworkWebhookPolicy = "np.rke2.io/ingress-webhook"
 
 	defaultNetworkPolicyName        = "default-network-policy"
 	defaultNetworkDNSPolicyName     = "default-network-dns-policy"
 	defaultNetworkIngressPolicyName = "default-network-ingress-policy"
+	defaultNetworkWebhookPolicyName = "default-network-ingress-webhook-policy"
 
 	defaultTimeout     = 30
 	cisAnnotationValue = "resolved"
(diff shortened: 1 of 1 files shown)


pkg/rke2/np.go
```diff
@@ -120,13 +122,44 @@ var networkIngressPolicy = v1.NetworkPolicy{
 					{
 						Protocol: &tcp,
 						Port: &intstr.IntOrString{
-							IntVal: int32(80),
+							StrVal: "http",
```
brandond, line 125: note: targeting by name is more correct, as the numeric ports can be set by chart values.

https://github.com/rancher/rke2-charts/blob/3a5a54512273934ab59711b44e969c9ae84c94fe/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.8.200/values.yaml#L36-L37

pkg/rke2/np.go
```diff
@@ -120,13 +122,44 @@ var networkIngressPolicy = v1.NetworkPolicy{
...
 						},
 					},
 					{
 						Protocol: &tcp,
 						Port: &intstr.IntOrString{
-							IntVal: int32(443),
+							StrVal: "https",
```
brandond, line 131: note: targeting by name is more correct, as the numeric ports can be set by chart values.

https://github.com/rancher/rke2-charts/blob/3a5a54512273934ab59711b44e969c9ae84c94fe/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.8.200/values.yaml#L36-L37

pkg/rke2/np.go
```diff
@@ -120,13 +122,44 @@ var networkIngressPolicy = v1.NetworkPolicy{
...
+		PolicyTypes: []v1.PolicyType{
+			v1.PolicyTypeIngress,
+		},
+		Ingress: []v1.NetworkPolicyIngressRule{
+			{
+				Ports: []v1.NetworkPolicyPort{
+					{
+						Protocol: &tcp,
+						Port: &intstr.IntOrString{
+							StrVal: "webhook",
```
brandond, line 162: note: targeting by name is more correct, as the numeric ports can be set by chart values.

https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.8.200/values.yaml#L600

dereknola (approved):

briandowns (approved):

Issue #5535, opened 2024-02-28
secrets-encrypt rotate-keys is not working since the metics server output is not as expected


aganesh-suse:
## Environment Details

  
*Infrastructure*
- [x] Cloud
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"

$ uname -m
x86_64
```
  
*Cluster Configuration:*

```
HA : 1 etcd , 2 cp, 1 agent node
```
  
*Config.yaml:*

ETCD server config: 
```
token: xxxx
disable-apiserver: true
disable-controller-manager: true
disable-scheduler: true
write-kubeconfig-mode: "0644"
secrets-encryption: true
node-external-ip: 1.1.1.1
debug: true
```

CP only node configs: 
```
token: xxxx
server: https://1.1.1.1:9345
disable-etcd: true
write-kubeconfig-mode: "0644"
secrets-encryption: true
node-external-ip: 1.2.3.4
debug: true
```
  
## Steps to reproduce: 

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/rke2 && sudo cp config.yaml /etc/rancher/rke2
```
2. Install RKE2
```
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION='v1.29.2-rc3+rke2r1' INSTALL_RKE2_TYPE='server' INSTALL_RKE2_METHOD=tar sh -
```
3. Start the RKE2 service
```
$ sudo systemctl enable --now rke2-server
or 
$ sudo systemctl enable --now rke2-agent
```
4. Verify Cluster Status: 
```
kubectl get nodes -o wide
kubectl get pods -A
```
5. Run secrets-encrypt rotate-keys:
```
sudo rke2 secrets-encrypt rotate-keys
```
6. Run metrics server command:
```
kubectl get --raw /metrics | grep apiserver_encryption
``` 
  
**Reproducing Results/Observations:**
- rke2 version used for replication:

```
$ rke2 -v
rke2 version v1.29.2-rc3+rke2r1 (08699dfffdf75a61a5e6064f9f8efe8ddae857fe)
go version go1.21.7 X:boringcrypto
```

```
$ kubectl get nodes
NAME               STATUS   ROLES                  AGE     VERSION
ip-172-31-16-219   Ready    etcd                   6m15s   v1.29.2+rke2r1
[...]

PR #5604, opened 2024-03-12
Enable apiserver to access updated encryption-config.json

dereknola:

#### Proposed Changes ####
- Mount server/cred directory. Allows apiserver to see hot reload for encryption-config.json when secrets-encryption controller updates the file.
- Improved comments

#### Linked Issues ####
https://github.com/rancher/rke2/issues/5535

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/podexecutor/staticpod.go
+++ b/pkg/podexecutor/staticpod.go
@@ -314,6 +314,10 @@ func (s *StaticPodConfig) APIServer(_ context.Context, etcdReady <-chan struct{}
 		dirs = append(dirs, filepath.Dir(auditLogFile))
 		excludeFiles = append(excludeFiles, auditLogFile)
 	}
+	// encryption config is refreshed by the secrets-encryption controller
+	// so we mount the directory to allow the pod to see the updates
+	dirs = append(dirs, filepath.Join(s.DataDir, "server/cred"))
+	excludeFiles = append(excludeFiles, filepath.Join(s.DataDir, "server/cred/encryption-config.json"))
 
 	return after(etcdReady, func() error {
 		return staticpod.Run(s.ManifestsDir, staticpod.Args{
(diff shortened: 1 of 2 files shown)


pkg/podexecutor/staticpod.go
```diff
@@ -314,6 +314,10 @@ func (s *StaticPodConfig) APIServer(_ context.Context, etcdReady <-chan struct{}
 		dirs = append(dirs, filepath.Dir(auditLogFile))
 		excludeFiles = append(excludeFiles, auditLogFile)
 	}
+	// encryption config is refreshed by the secrets-encryption controller
+	// so we mount the directory to allow the pod to see the updates
+	dirs = append(dirs, filepath.Join(s.DataDir, "server/cred"))
+	excludeFiles = append(excludeFiles, filepath.Join(s.DataDir, "server/cred/encryption-config.json"))
```
brandond, line 320: I think we already have a var for this path at `runtime.EncryptionConfig`. We should have access to runtime here?

pkg/staticpod/staticpod.go
```diff
@@ -103,6 +103,15 @@ func Run(dir string, args Args) error {
 		return err
 	}
 
+	// Check to make sure we aren't double mounting directories and the files in those directories
+	for _, dir := range args.Dirs {
+		for _, file := range files {
+			if strings.HasPrefix(file, dir) {
+				logrus.Warnf("file %s is being double mounted with directory %s", file, dir)
```
brandond, line 110: Nit:
```suggestion
				logrus.Warnf("Mount of file %s shadows content of mounted directory %s", file, dir)
```

brandond: In most cases this isn't going to be a warning that the user can actually do anything about. Instead of warning, do we maybe just want to remove the explicit file mount if it is covered by a directory?

Similarly, do we need to check for directories mounted under existing directories?

brandond (approved):

vitorsavian (approved):

PR #5690, opened 2024-04-08
Add extra information in install.ps1 to help users

manuelbuil:
#### Proposed Changes ####

[RKE2 docs ](https://docs.rke2.io/install/quickstart#3-configure-path) already explain that the $PATH should be set with `[System.Environment]::SetEnvironmentVariable` but the current installation.ps1 is not specifying that. Therefore, a user might set the PATH as explained in the script and still see problems because rke2 can't find containerd in the PATH. This PR adds some extra information

#### Types of Changes ####

Docs

#### Verification ####

No need 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -726,7 +726,7 @@ switch ($Method) {
         Install-AirgapTarball -CommitHash $Commit -InstallAgentImageDir $AgentImagesDir -TempAirgapTarball $TMP_AIRGAP_TARBALL -ExpectedImageAirgapChecksum $AIRGAP_CHECKSUM_EXPECTED -TempImageChecksums $TMP_AIRGAP_CHECKSUMS
         Test-TarballChecksum -Tarball $TMP_BINARY_TARBALL -ExpectedChecksum $BINARY_CHECKSUM_EXPECTED
         Expand-Tarball -InstallPath $TarPrefix -Tarball $TMP_BINARY_TARBALL
-        Write-InfoLog "install complete; you may want to run:  `$env:PATH+=`";$TarPrefix\bin;C:\var\lib\rancher\rke2\bin`""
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -726,7 +726,7 @@ switch ($Method) {
         Install-AirgapTarball -CommitHash $Commit -InstallAgentImageDir $AgentImagesDir -TempAirgapTarball $TMP_AIRGAP_TARBALL -ExpectedImageAirgapChecksum $AIRGAP_CHECKSUM_EXPECTED -TempImageChecksums $TMP_AIRGAP_CHECKSUMS
         Test-TarballChecksum -Tarball $TMP_BINARY_TARBALL -ExpectedChecksum $BINARY_CHECKSUM_EXPECTED
         Expand-Tarball -InstallPath $TarPrefix -Tarball $TMP_BINARY_TARBALL
-        Write-InfoLog "install complete; you may want to run:  `$env:PATH+=`";$TarPrefix\bin;C:\var\lib\rancher\rke2\bin`""
+        Write-InfoLog "install complete; you may want to run:  `$env:PATH+=`";$TarPrefix\bin;C:\var\lib\rancher\rke2\bin`" or use [System.Environment]::SetEnvironmentVariable if rke2 will run as a service"
```
brandond, line 729: Under what circumstances would rke2 not run as a service? I don't think anyone is out there running rke2 agents from the command prompt.

Do we need some changes to properly set up the environment for the service when installing it, so that the correct paths are in its path?

brandond (changes requested):

install.ps1
manuelbuil: You mean adding it as part of the RKE2 windows code? Probably a better approach and we forget about changing the PATH. Let me see how we could do this

manuelbuil: Found the culprit! https://github.com/rancher/rke2/pull/5698

Issue #5697, opened 2024-04-10
Wrong path setting in rke2-windows


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Any

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

One windows node

**Describe the bug:**

When executing rke2 in windows without setting the PATH as explained in the docs: https://docs.rke2.io/install/quickstart#3-configure-path, rke2 fails to start correctly because neither containerd nor kubelet can be found in the path

**Steps To Reproduce:**

- Installed RKE2: 

Deploy RKE2 in Windows without setting the PATH as explained in the docs: https://docs.rke2.io/install/quickstart#3-configure-path

**Expected behavior:**

RKE2 does not need any previous configuration of the PATH

**Actual behavior:**

RKE2 needs a previous configuration of the PATH

**Additional context / logs:**

mdrahman-suse:
# Validation on master branch with commit

## Environment
```
1 Linux server and 1 Windows server
```

Followed steps mentioned here: https://github.com/rancher/rke2/issues/5697#issue-2234846492
Need to setup a server node to actually see the error

## Replication
```
C:\usr\local\bin\rke2.exe --version
rke2.exe version v1.29.3+rke2r1 (1c82f7ed292c4ac172692bb82b13d20733909804)
go version go1.21.8
```
- Logs
```
...
time="2024-04-16T23:27:46Z" level=debug msg="Kubelet image credential provider bin directory check failed: CreateFile /var/lib/rancher/credentialprovider/bin: The system cannot find the path specified."
...
time="2024-04-16T23:28:01Z" level=info msg="Running containerd -c C:\\var\\lib\\rancher\\rke2\\agent\\etc\\containerd\\config.toml"
W0416 23:28:01.043738    3524 logging.go:59] [core] [Channel #1 SubChannel #2] grpc: addrConn.createTransport failed to connect to {Addr: "//./pipe/containerd-containerd", ServerName: "%2F%2F.%2Fpipe%2Fcontainerd-containerd", }. Err: connection error: desc = "transport: Error while dialing: open //./pipe/containerd-containerd: The system cannot find the file specified."
time="2024-04-16T23:28:01Z" level=error msg="containerd exited: exec: \"containerd\": executable file not found in %PATH%"
```
- Windows agent fails to join the cluster

## Validation
```
C:\usr\local\bin\rke2.exe --version
[...]

PR #5698, opened 2024-04-10
Fix Windows path setting

manuelbuil:
#### Proposed Changes ####

For some reason it was necessary to configure the PATH for RKE2 to work correctly in Windows, as explained in the docs: https://docs.rke2.io/install/quickstart#3-configure-path. The reason is that we were setting the PATH wrongly and using ":" instead of ";". One is typical for Unix systems and the other one is for Windows

#### Types of Changes ####

Bugfix

#### Verification ####

Run rke2-windows without configuring the PATH previously as set in the docs: https://docs.rke2.io/install/quickstart#3-configure-path

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5697

#### User-Facing Change ####

```release-note
Fix windows path setting
```

### DIFF
--- a/pkg/pebinaryexecutor/pebinary.go
+++ b/pkg/pebinaryexecutor/pebinary.go
@@ -100,7 +100,7 @@ func (p *PEBinaryConfig) Bootstrap(ctx context.Context, nodeConfig *config.Node,
 	if err != nil {
 		return err
 	}
-	if err := os.Setenv("PATH", execPath+":"+os.Getenv("PATH")); err != nil {
+	if err := os.Setenv("PATH", execPath+";"+os.Getenv("PATH")); err != nil {
 		return err
 	}
 


pkg/pebinaryexecutor/pebinary.go
```diff
@@ -100,7 +100,7 @@ func (p *PEBinaryConfig) Bootstrap(ctx context.Context, nodeConfig *config.Node,
 	if err != nil {
 		return err
 	}
-	if err := os.Setenv("PATH", execPath+":"+os.Getenv("PATH")); err != nil {
+	if err := os.Setenv("PATH", execPath+";"+os.Getenv("PATH")); err != nil {
```
brandond, line 103: FYI there is `os.PathListSeparator` if you want to avoid forgetting which is which on different platforms.

https://pkg.go.dev/os#pkg-constants

brandond (approved):

thomasferrandiz (approved):

dereknola (approved):

manuelbuil:
Backports tomorrow!

Issue #5702, opened 2024-04-11
RKE2-windows flannel not retrying


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

When the Windows node has only one interface and we set up rke2-flannel, we will lose connectivity to the node for some seconds. It varies between 5 to 25 seconds. That's how windows works when binding an hns overlay network to a physical interface. In some environments, if it takes too long, flanneld exits with an error because it can't contact kube-api anymore

JIRA: SURE-8128

**Steps To Reproduce:**

- Installed RKE2:
Not always reproducible. Deploy rke2 on a windows agent with only one interface and check the rke2 logs, you will see:
`Flanneld has an error:`
And we never retry.
If you can't reproduce it, just verify that flanneld is using v0.25.1:

```
C:\var\lib\rancher\rke2\bin\flanneld.exe --version
v0.25.1
```

**Expected behavior:**

As we do with Calico, we retry again because it could be fine that flanneld must be restarted

**Actual behavior:**

We never retry and flanned dies forever. The workaround is restarting the rke2 service even though the rest of components work correctly

**Additional context / logs:**

mdrahman-suse:
## Validation on master branch with commit 95e13dc6

Unable to replicate the issue, validated the flannel version in Windows node

### Validation
```
PS C:\Users\Administrator> C:\usr\local\bin\rke2.exe -v
rke2.exe version v1.29.3+dev.95e13dc6 (95e13dc62fdbda33de2c709f1149b0c361d920b9)
go version go1.21.8

PS C:\Users\Administrator> C:\var\lib\rancher\rke2\bin\flanneld.exe --version
v0.25.1
```

## Validation on release-1.28 branch with commit 04086ee5

Unable to replicate the issue, validated the flannel version in Windows node
### Validation
```
PS > C:\usr\local\bin\rke2.exe -v
rke2.exe version v1.28.8+dev.04086ee5 (04086ee5e83a876f6576418c7fa4ff9ea5e1caf7)
go version go1.21.8

PS > C:\var\lib\rancher\rke2\bin\flanneld.exe --version
v0.25.1
```

## Validation on release-1.27 branch with commit 191329a6

Unable to replicate the issue, validated the flannel version in Windows node
### Validation
```
PS > C:\usr\local\bin\rke2.exe -v
[...]

PR #5703, opened 2024-04-11
Retry flanneld if it fails to start

manuelbuil:
#### Proposed Changes ####

This PR does two things:
1 - Slightly changes one log message because it was raising alarms in our customers/support even though it was harmless: `Flanneld can't start because it can't find node` ==> `Checking if node %s is already registered before starting flanneld`
2 - Retrying executing flanneld if it failed. We try 5 times and give up if it always fails. We should retry at least 1 or 2 times, in the case of a windows node with only one interface as explained in the issue. 5 seems like a good number to give up 

#### Types of Changes ####

Bugfix

#### Verification ####

The new log is easy to verify. Just check the rke2 logs

For the second part, make sure everything is running. Then execute: `Stop-Process -Name flanneld`. Then verify that the rke2 logs include:
```
4/11/2024 1:05:10 PM  Flanneld has an error: exit status 0xffffffff. Check c:\var\lib\rancher\rke2\agent\logs\flanneld.log for extra information
4/11/2024 1:05:10 PM  Running flanneld with envs: [NODE_NAME=mbuil-windows PATH=C:\var\lib\rancher\rke2\data\v1.29.3-rke2r1-windows-amd64-dfa86afd0b9e\bin:C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\Window
                      sPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program
                      Files\dotnet\;C:\ProgramData\chocolatey\bin;;c:\var\lib\rancher\rke2\bin;c:\usr\local\bin;C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps] and args:
                      [--kubeconfig-file=c:\var\lib\rancher\rke2\agent\flannel.kubeconfig --ip-masq --kube-subnet-mgr --iptables-forward-rules=false --iface=10.1.1.14
                      --net-config-path=c:\var\lib\rancher\rke2\agent\flanneld-net-conf.json]
4/11/2024 1:05:10 PM  Flanneld exited. Retrying.
```

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5702

#### User-Facing Change ####

```release-note
Retry flanneld.exe if it fails
```

### DIFF
--- a/pkg/windows/flannel.go
+++ b/pkg/windows/flannel.go
@@ -258,9 +258,9 @@ func (f *Flannel) Start(ctx context.Context) error {
 
 	// Wait for the node to be registered in the cluster
 	if err := wait.PollImmediateWithContext(ctx, 3*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
+		logrus.Infof("Checking if node %s is already registered before starting flanneld", f.CNICfg.Hostname)
 		_, err := f.KubeClient.CoreV1().Nodes().Get(ctx, f.CNICfg.Hostname, metav1.GetOptions{})
 		if err != nil {
-			logrus.WithError(err).Warningf("Flanneld can't start because it can't find node, retrying %s", f.CNICfg.Hostname)
 			return false, nil
 		} else {
 			logrus.Infof("Node %s registered. Flanneld can start", f.CNICfg.Hostname)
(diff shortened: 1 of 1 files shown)


pkg/windows/flannel.go
```diff
@@ -293,15 +293,21 @@ func startFlannel(ctx context.Context, config *FlannelConfig, logPath string) {
...
-	logrus.Infof("Flanneld Envs: %s and args: %v", specificEnvs, args)
-	cmd := exec.CommandContext(ctx, "flanneld.exe", args...)
-	cmd.Env = append(specificEnvs)
-	cmd.Stdout = outputFile
-	cmd.Stderr = outputFile
-	if err := cmd.Run(); err != nil {
-		logrus.Errorf("Flanneld has an error: %v. Check %s for extra information", err, logPath)
+	// We retry running Flanneld 5 times before giving up
+	maxretries := 5
+	for i:=0; i < maxretries; i++ {
```
thomasferrandiz, line 298: In flannelm we use https://github.com/avast/retry-go/ for retries.
I don't know if it would be worth it her?

thomasferrandiz (approved):

rbrtbnfgl (approved):

pkg/windows/flannel.go
brandond: What specifically are we trying to fix by retrying here? All other prerequisites are polled for readiness before attempting to start the component. Other components that are started in a retry loop don't have a limit on them and are just retried indefinitely while RKE2 is up.

I feel like we should figure out what is causing flannel to fail initially, and poll for that to become come available before starting flannel, rather than artificially limiting the number of restarts.

If we do for some reason need to artificially limit the restarts, shouldn't we `logrus.Fatalf` after the for loop, rather than just leaving rke2 running without any further attempt to start the CNI?

brandond (changes requested):

pkg/windows/flannel.go
manuelbuil: The first time flannel runs, it creates the HNS network, which is an overlay flannel network bound with a physical interface. Because of how windows works, that interface then becomes unavailable for some time (according to my measurements from 10s - 25s). If it is closer to the 25s, flanneld dies because it failed to contact kube-api for too long. That's the case that we are trying to fix here

manuelbuil: One thing we could is: run flannel once and then, if it fails, check for the interface to come back. Once we are sure the interface is back, run flannel again. If it fails a second time, then do the fatal failure. What do you think?

brandond: Does flannel have an option to do a one-shot startup that will just create the HNS network if necessary, and then exit? We could do that, wait for the network to stabilize, and then start the "permanent" flannel that we expect to not crash?

manuelbuil: Unfortunately that option does not exist. I have an idea I could try tomorrow in Flannel, hopefully before code freeze

manuelbuil:
Trying this instead https://github.com/rancher/rke2/pull/5747, hopefully solving the problem. Closing for the moment

Issue #5778, opened 2024-04-15
Errors when uninstalling rke2 from windows agent


manuelbuil:
**Environmental Info:**
RKE2 Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

Mixed cluster with windows and linux nodes

**Describe the bug:**

When executing `.\rke2-uninstall.ps1` sometimes we see the error:
```
time="2024-04-15T14:49:31Z" level=error msg="unable to delete k8s.io" error="namespace \"k8s.io\" must be empty, but it still has blobs, snapshots on \"windows\" snapshotter: failed precondition"
ctr: unable to delete k8s.io: namespace "k8s.io" must be empty, but it still has blobs, snapshots on "windows" snapshotter: failed precondition
```
and after that rke2 uninstallation normally fails when trying to delete the directory /var because there are containerd files that were not properly deleted. Deleting those files afterwards is impossible because Windows complains that they are locked.

**Steps To Reproduce:**

- Installed RKE2:
Any RKE2 1.27, 1.28 or 1.29

**Expected behavior:**

rke2-uninstall.ps1 removes everything installed by rke2

**Actual behavior:**

rke2-uninstall.ps1 sometimes fails to remove everything installed by rke2

**Additional context / logs:**

mdrahman-suse:
# Validated on master branch with commit 9eae9199

## Environment
```
1 Linux server, 1 linux agent and 1 Windows agent (2019)
```
## Testing
- Install rke2 on linux server and add linux and windows agent
- Deploy workloads (Linux and Windows)
- Ensure the cluster is up and running
- Login to Windows node and run the uninstall script
- Ensure the scripts runs successfully without any errors `C:\usr\local\bin\rke2-uninstall.ps1`
NOTE: It is not easily replicable sometimes

## Replication
```
PS C:\Users\Administrator> C:\usr\local\bin\rke2.exe -v
rke2.exe version v1.30.1+rke2r1 (e7f87c6dd56fdd76a7dab58900aeea8946b2c008)
go version go1.22.2
```
```
PS C:\Users\Administrator> c:/usr/local/bin/rke2-uninstall.ps1
Beginning the uninstall process
docker.io/mbuilsuse/pstools:v0.2.0
docker.io/mbuilsuse/pstools@sha256:2f7abee3e5ecc3c672cb1054ac053dab4e149ad2e51ab3143c43106c9c6fe335
docker.io/phillipsj/pstools:v0.2.0
[...]

PR #5779, opened 2024-04-15
Improve rke2-uninstall.ps1 script

manuelbuil:
#### Proposed Changes ####

This PR tries to fix the current implementation of the rke2-uninstall.ps1 script. Often this script fails to remove everything from the node because by the time it tries to remove the containerd namespace, there are still resources in that namespace. As a consequence, we end up with files in /var/lib/rancher/agent/ which are impossible to remove because according to Windows, they are locked. To remove them, a restart normally works but not always.

This PR does two things:
1 - The rke2-uninstall.ps1 would create a lock file in the dataDir. Then it will kill kubelet. In the pebinary code, we won't restart kubelet if that lock file exists
2 - It retries removing the namespace for a few times, in case the previous trial failed. There might be situations where the script can't succeed (e.g. user forgot to drain). For those cases, it gives up after 30 seconds and continues deleting other stuff, as it does today

#### Types of Changes ####

Bugfix

#### Verification ####

Install rke2 on linux and windows. Deploy a pod in windows. Run the rke2-uninstall.ps1 script. Without this PR, half of the time the script will fail and it will be impossible to remove the directory /var/lib/rancher.

With this PR, the script works and /var/lib/rancher does not exist anymore

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5778
https://github.com/rancher/rancher/issues/45016

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/bundle/bin/rke2-uninstall.ps1
+++ b/bundle/bin/rke2-uninstall.ps1
@@ -272,6 +272,13 @@ function Remove-Containerd () {
     }
 
     if (ctr) {
+        # We create a lockfile to prevent rke2 service from starting kubelet again
+        Create-Lockfile
+        Stop-Process -Name "kubelet"
+        while (-Not(Get-Process -Name "kubelet").HasExited) {
+            Write-LogInfo "Waiting for kubelet process to stop"
+            Start-Sleep -s 5
+        }
         $namespaces = $(Find-Namespaces)
         if (-Not($namespaces)) {
             $ErrorActionPreference = 'SilentlyContinue'
@@ -292,10 +299,23 @@ function Remove-Containerd () {
             foreach ($image in $images) {
                 Remove-Image $ns $image
             }
-            Remove-Namespace $ns
-            # TODO
-            # clean pods with crictl
-            # $CONTAINER_RUNTIME_ENDPOINT = "npipe:\\.\\pipe\\containerd-containerd"
+        }
+
+        # Some resources in the namespace take a while to disappear. Try several times to remove the namespace and give up after 30s 
+        $endTime = (Get-Date).AddSeconds(30)
+        while ((Get-Date) -lt $endTime) {
+            $namespaces = $(Find-Namespaces)
+            if ($namespaces) {
+                foreach ($ns in $namespaces) {
+                    Remove-Namespace $ns
+                }
+            } else {
+                break
+            }
+            Start-Sleep -Seconds 5
+            if ((Get-Date) -ge $endTime) {
+                Write-Output "Unable to remove all namespaces"
+            }
         }
     }
     else {
@@ -343,6 +363,20 @@ function Remove-Container() {
 function Remove-Namespace() {
     $namespace = $args[0]
     Invoke-Ctr -cmd "namespace remove $namespace"
+}
+
+function Create-Lockfile() {
+    # We fetch ctr.exe path and place the lock there
+    $command = Get-Command ctr -ErrorAction SilentlyContinue
+    if ($command) {
+        $executablePath = $command.Source
+        $dataBinDirDirectory = Split-Path -Parent $executablePath
+        $lockFilePath = Join-Path -Path $dataBinDirDirectory -ChildPath "rke2-uninstall.lock"
+        New-Item -ItemType File -Path $lockFilePath -Force
+    } else {
+	# ctr should exist at this point but just in case we add this log
+        Write-Host "ctr.exe not found, container cleanup and RKE2 uninstallation is unlikely to succeed."
+    }
 }
 
 function Invoke-Rke2Uninstall () {
--- a/pkg/pebinaryexecutor/pebinary.go
+++ b/pkg/pebinaryexecutor/pebinary.go
@@ -165,6 +165,15 @@ func (p *PEBinaryConfig) Kubelet(ctx context.Context, args []string) error {
 		cleanArgs = append(cleanArgs, arg)
 	}
 
+	// It should never happen but just in case, we make sure the rke2-uninstall.lock does not exist before starting kubelet
+	lockFile := filepath.Join(p.DataDir, "bin", "rke2-uninstall.lock")
+	if _, err := os.Stat(lockFile); err == nil {
+		// If the file exists, delete it
+		if err := os.Remove(lockFile); err != nil {
+			logrus.Errorf("Failed to remove the %s file: %v", lockFile, err)
+		}
+	}
+
 	win.ProcessWaitGroup.StartWithContext(ctx, func(ctx context.Context) {
 		for {
 			logrus.Infof("Running RKE2 kubelet %v", cleanArgs)
@@ -184,6 +193,12 @@ func (p *PEBinaryConfig) Kubelet(ctx context.Context, args []string) error {
 				logrus.Errorf("Kubelet exited: %v", err)
 			}
 			cancel()
+
+			// If the rke2-uninstall.ps1 script created the lock file, we are removing rke2 and thus we don't restart kubelet
+			if _, err := os.Stat(lockFile); err == nil {
+				logrus.Infof("rke2-uninstall.lock exists. kubelet is not restarted")
+				return
+			}
 
 			select {
 			case <-ctx.Done():


HarrisonWAffel (approved):
lgtm, do you think we should add a blurb in the docs calling out the need to drain the node before running this script?

bundle/bin/rke2-uninstall.ps1
```diff
@@ -100,6 +100,7 @@ function Invoke-HNSRequest {
 
 # cleanup
 Write-Host "Beginning the uninstall process"
+Read-Host "Node must be drained and removed from the cluster before proceeding. Press Enter to continue."
```
brandond, line 103: We don't require this for k3s or rke2 on linux. Is this just to give us less stuff to clean up when uninstalling?

bundle/bin/rke2-uninstall.ps1
```diff
@@ -292,10 +293,23 @@ function Remove-Containerd () {
             foreach ($image in $images) {
                 Remove-Image $ns $image
             }
-            Remove-Namespace $ns
-            # TODO
-            # clean pods with crictl
-            # $CONTAINER_RUNTIME_ENDPOINT = "npipe:\\.\\pipe\\containerd-containerd"
+        }
+
+        # Some resources in the namespace take a while to disappear. Try several times to remove the namespace and give up after 30s
```
brandond, line 298: Why is it so critical that this be removed? There's only ever the one namespace, `k8s.io` - and it can't be removed until everything in it is gone. Deleting the namespace doesn't trigger deletion of the images and containers inside it, those need to be done manually but I don't see that we're doing that here. Do we need to retry removing the specific resources that are blocking its deletion, instead of just retrying the namespace delete and hoping the other things have gone away in the background?

manuelbuil: Probably my comment confused you. In the previous lines, we are removing tasks, containers and images, which should remove all resources. The problem is that by the time we start removing namespaces, some of those resources are still available, probably set for deletion, but available. That makes the deletion of the namespace fail. Then, we move on and uninstall containerd, rke2, kube-proxy... We are killing everything when those resources are still available and that ends up badly. We end up with containerd files (normally related to snapshots) remiaining in the file system and impossible to remove

bundle/bin/rke2-uninstall.ps1
manuelbuil: Based on my experience, if we don't do this, the script does not work correctly. The problem is that all resources of containerd need to be gone before stopping and uninstalling kubelet/containerd, otherwise, there are remaining files which are impossible to remove. If there are pods running in this node, it's impossible to clean everything because immediately after killing containerd tasks with this script, new tasks are recreated.

In linux, with the correct permissions (i.e. sudo), we can remove everything in `/var/lib/rancher`. However in Windows we can't if the files belong to a process that does not exist anymore (containerd). The workaround I found is rebooting the host so that Windows "forgets" about that process

bundle/bin/rke2-uninstall.ps1
manuelbuil: While being at the gym, I thought that perhaps we could try removing the 30s. countdown and just verify everything (tasks, containers, images and snapshots) is empty before removing the namespace. However, if for example, the user forgot to drain, containerd will never be empty, so we would get stuck forever. That's why I think we need some kind of upper limit

brandond: > if for example, the user forgot to drain, containerd will never be empty

Is that because the kubelet will still be running, trying to restart the containers?

It sounds like we need to kill the kubelet to prevent it from restarting things, but leave containerd running so we can use it to kill containers by hand. Is that correct? This would probably require some changes to the RKE2 golang code, to allow for a partial node shutdown that allows for stopping the kubelet but not containerd. I don't think that asking the user to drain the node is going to work - even with a drain, some pods remain running. It is also asking a lot of the user to have them drain before running killall. Killall should be able to clean things up even if the node wasn't prepared properly.

manuelbuil: I changed the behaviour a bit. Now the rke2-uninstall.ps1 creates a lock file in DataDir and then kills kubelet. If kubelet dies, pebinary checks if that lock file exists and if it does, it does not restart kubelet. The lock file gets removed when the script removes the DataDir, so it won't affect new rke2 installations

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -185,6 +185,13 @@ func (p *PEBinaryConfig) Kubelet(ctx context.Context, args []string) error {
 			}
 			cancel()
 
+			// If the rke2-uninstall.ps1 script created the lock file,we are removing rke2 and thus we don't restart kubelet
+			_, err := os.Stat(filepath.Join(p.DataDir, "agent", "rke2-uninstall.lock"))
+			if err == nil {
```
brandond, line 190: nit:
```suggestion
			if _, err := os.Stat(filepath.Join(p.DataDir, "agent", "rke2-uninstall.lock")); err == nil {
```

brandond: As an additional safety measure, should we remove the lock file before entering the `for` loop that runs/restarts the kubelet? That way it would just inhibit the kubelet from restarting while RKE2 is up (while it's in the for loop), if you restart the service with the lock file in place it will not break rke2.

manuelbuil: yes, let me add that check just in case

bundle/bin/rke2-uninstall.ps1
```diff
@@ -345,6 +365,17 @@ function Remove-Namespace() {
     Invoke-Ctr -cmd "namespace remove $namespace"
 }
 
+function Create-Lockfile() {
+    $fullPath = Join-Path -Path "C:\var\lib\rancher\rke2\agent\" -ChildPath "kubelet.kubeconfig"
+    if (-not ($fullPath)) {
+        # For cases where Datadir is not the default location (can take a minute to find)
+        $fullPath = Get-ChildItem -Path C:\ -Recurse -ErrorAction SilentlyContinue | Where-Object { $_.Name -eq $file } | Select-Object -ExpandProperty FullName
```
brandond, line 372: Doing a full disk search on the C: is frightening, lets not do this. We can't even guarantee that it's on this disk, if they change the data-dir they are VERY likely to put it on a different drive letter.

Can we perhaps look at the kubelet process to find its executable path, and determine the data-dir from that?

brandond (changes requested):

bundle/bin/rke2-uninstall.ps1
manuelbuil: good idea, actually we are checking a bit before if `ctr` command exists, we can fetch dataDir from there and place the lockFile in that directory

brandond (approved):

brandond (approved):
lgtm with warning nit

bundle/bin/rke2-uninstall.ps1
```diff
@@ -345,6 +365,20 @@ function Remove-Namespace() {
...
+    # We fetch ctr.exe path and place the lock there
+    $command = Get-Command ctr -ErrorAction SilentlyContinue
+    if ($command) {
+        $executablePath = $command.Source
+        $dataBinDirDirectory = Split-Path -Parent $executablePath
+        $lockFilePath = Join-Path -Path $directory -ChildPath "rke2-uninstall.lock"
+        New-Item -ItemType File -Path $lockFilePath -Force
+    } else {
+	# ctr should exist at this point but just in case we add this log
+        Write-Host "Command ctr not found. Aumatica uninstallation might fail"
```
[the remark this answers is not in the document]
manuelbuil: ugh! Thanks, I missed that

bundle/bin/rke2-uninstall.ps1
```diff
@@ -345,6 +365,20 @@ function Remove-Namespace() {
     Invoke-Ctr -cmd "namespace remove $namespace"
 }
 
+function Create-Lockfile() {
+    # We fetch ctr.exe path and place the lock there
+    $command = Get-Command ctr -ErrorAction SilentlyContinue
+    if ($command) {
+        $executablePath = $command.Source
+        $dataBinDirDirectory = Split-Path -Parent $executablePath
+        $lockFilePath = Join-Path -Path $directory -ChildPath "rke2-uninstall.lock"
```
HarrisonWAffel, line 374: Looking back over this with the new changes, I don't see `$directory` being set anywhere, looks like it should be `$dataBinDirDirectory`

manuelbuil: good catch! Let me retest this carefully again

manuelbuil:
restested and works :)

HarrisonWAffel (approved):

Issue #4486, opened 2023-07-19
Add calico-bgp test for windows


manuelbuil:
**Is your feature request related to a problem? Please describe.**

rke2-windows can now use calico-bgp with this new PR https://github.com/rancher/rke2/pull/4164

We should replicate the [mixedos e2e test](https://github.com/rancher/rke2/tree/master/tests/e2e/mixedos) but using calico-bgp and verify that traffic between linux and windows works. 

Warning! Some iptables magic would need to be applied to avoid packets being dropped 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

rancher-max:
Thank you for the updates to the tests! 🎉

PR #5859, opened 2024-04-29
Add mixedos BGP e2e test

manuelbuil:
#### Proposed Changes ####

Add a new e2e test that verifies BGP communication works in a mixedos env

#### Linked Issues ####

https://github.com/rancher/rke2/issues/4486

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/mixedosbgp/Vagrantfile
+++ b/tests/e2e/mixedosbgp/Vagrantfile
@@ -1,0 +1,144 @@
+ENV['VAGRANT_NO_PARALLEL'] = 'no'
+NODE_ROLES = (ENV['E2E_NODE_ROLES'] ||
+["server-0", "linux-agent-0", "windows-agent-0" ])
+NODE_BOXES = (ENV['E2E_NODE_BOXES'] ||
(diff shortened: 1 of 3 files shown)


thomasferrandiz (approved):

dereknola (changes requested):
you need to change the name of the test folder to `mixedosbgp`, it must match the test name prefix. Or change both to be `mixedos_bgp`... we don't use dashes for the testing names.

rbrtbnfgl:
Maybe you need to add `Restart-Service RemoteAccess` somewhere on the script after RKE2 started on windows because it could happen that the BGP configuration is not rightly taken by the service. i don't know if it has been fixed in Calico.

manuelbuil:
> Maybe you need to add `Restart-Service RemoteAccess` somewhere on the script after RKE2 started on windows because it could happen that the BGP configuration is not rightly taken by the service. i don't know if it has been fixed in Calico.

The test was working in my computer, maybe it is not needed anymore

dereknola (approved):

brandond (approved):

Issue #5863, opened 2024-04-29
Enable E2E testing with local RKE2 builds


dereknola:
**Is your feature request related to a problem? Please describe.**

We need to support building a local RKE2 image + binary and running the E2E tests with this. 

**Describe the solution you'd like**

A `-local` flag that will scp the appropriate artifacts on the E2E VMs during testing

**Describe alternatives you've considered**

**Additional context**

PR #5864, opened 2024-04-29
Add E2E as a GitHub Action

dereknola:
#### Proposed Changes ####
- Enable local builds of RKE2 to run as E2E tests
- Add the dnscache E2E test to PR runs (only test with less than 3 nodes, so it fits inside GHA runner)
- Combine E2E and Integration tests into single "Test Suite" Workflow, helps with the limit number of Rancher runner instances

#### Verification ####
CI should be green

#### Testing ####
Its all testing changes

#### Linked Issues ####
https://github.com/rancher/rke2/issues/5863

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -1,24 +1,19 @@
-name: Integration Test Coverage
+name: Test Suite
 on: 
   push:
     paths-ignore:
       - "**.md"
       - "channel.yaml"
       - "install.sh"
-      - "tests/**"
-      - "!tests/integration**"
       - ".github/**"
-      - "!.github/workflows/integration.yaml"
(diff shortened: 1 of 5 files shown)


manuelbuil (approved):
LGTM. This means we will run some of the e2e tests in each PR (so far only one), right?

dereknola:
We have a limited on the GH runners with only 4 vcpu and 16GB of RAM. this effectively limits us to only running `1 server` or `1 server + 1 agent` E2E tests. Most of the e2e tests are 3 or 5 VM tests.

As of right now, EIO has larger custom runners, but they don't have nested virtualization enabled on them, so we can't use those. Current guidance for anything VM related is to use the GH native runners.

galal-hussein (approved):

PR #5885, opened 2024-05-06
Add cilium no proxy e2e test

manuelbuil:
#### Proposed Changes ####

Adds a e2e test for cilium without kube-proxy and using BPF. It uses ubuntu2310 because we need a kernel version above 5.10.

This PR also moves a couple of functions to testutils and consequently modifies how other tests are calling those functions

#### Types of Changes ####

New test

#### Verification ####

go test -v -timeout=25m ./tests/e2e/ciliumnokp/

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/ciliumnokp/Vagrantfile
+++ b/tests/e2e/ciliumnokp/Vagrantfile
@@ -85,10 +85,7 @@ Vagrant.configure("2") do |config|
     NODE_BOXES = NODE_BOXES.split(" ", -1)
   end
 
-  # Must iterate on the index, vagrant does not understand iterating 
-  # over the node roles themselves
-  NODE_ROLES.length.times do |i|
-    name = NODE_ROLES[i]
+  NODE_ROLES.each_with_index do |name, i|
     config.vm.define name do |node|
       roles = name.split("-", -1)
       role_num = roles.pop.to_i
--- a/tests/e2e/ciliumnokp/Vagrantfile
+++ b/tests/e2e/ciliumnokp/Vagrantfile
@@ -65,7 +65,6 @@ def provision(vm, roles, role_num, node_num)
         node-ip: #{node_ip4},#{node_ip6}
         node-external-ip: #{node_ip4},#{node_ip6}
         token: vagrant-rke2
-        kubelet-arg: "--node-ip=0.0.0.0" # Workaround for https://github.com/kubernetes/kubernetes/issues/111695
       YAML
     end
   end
--- a/tests/e2e/ciliumnokp/Vagrantfile
+++ b/tests/e2e/ciliumnokp/Vagrantfile
@@ -25,12 +25,12 @@ def provision(vm, roles, role_num, node_num)
     :libvirt__ipv6_address => "#{NETWORK6_PREFIX}::1",
     :libvirt__ipv6_prefix => "64"
     
-  vagrant_defaults = File.exists?("./vagrantdefaults.rb") ? "./vagrantdefaults.rb" : "../vagrantdefaults.rb"
+  vagrant_defaults = File.exist?("./vagrantdefaults.rb") ? "./vagrantdefaults.rb" : "../vagrantdefaults.rb"
   load vagrant_defaults
   
   defaultOSConfigure(vm)
   
-  scripts_location = Dir.exists?("./scripts") ? "./scripts" : "../scripts" 
+  scripts_location = Dir.exist?("./scripts") ? "./scripts" : "../scripts" 
   vm.provision "IPv6 Setup", type: "shell", path: scripts_location + "/ipv6.sh", args: [node_ip4, node_ip6, node_ip6_gw, "cilium", vm.box]
   
   install_type = getInstallType(vm, RELEASE_VERSION, GITHUB_BRANCH)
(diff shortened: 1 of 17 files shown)


tests/e2e/scripts/cilium_nokubeproxy.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+ip4_addr=$1
+
+# Override default calico and specify the interface for windows-agent
+# by default, the windows-agent use a different interface name than the linux-agent
+mkdir -p /var/lib/rancher/rke2/server/manifests
```
dereknola, line 6: You mention the `windows-agent`, but this test doesn't have any windows VMs. Is the intention to expand this test later?

tests/e2e/ciliumnokp/ciliumnokp_test.go
```diff
@@ -0,0 +1,264 @@
...
+		if len(fields) > 2 {
+			objIPs = append(objIPs, objIP{name: fields[0], ipv4: fields[1], ipv6: fields[2]})
+		} else if len(fields) > 1 {
+			objIPs = append(objIPs, objIP{name: fields[0], ipv4: fields[1]})
+		} else {
+			objIPs = append(objIPs, objIP{name: fields[0]})
+		}
+	}
+	return objIPs, nil
+}
```
dereknola, line 58: Can you move these under `testutils.go`, we likely will use them again in other tests.

dereknola (changes requested):

tests/e2e/scripts/cilium_nokubeproxy.sh
manuelbuil: copy/paste typo

tests/e2e/ciliumnokp/ciliumnokp_test.go
manuelbuil: good idea! I'll also change the dualStack test which is using this function as well, so that all use the ones in testutils. This PR will increase a bit the scope but I think it is worth it

tests/e2e/ciliumnokp/Vagrantfile
```diff
@@ -0,0 +1,98 @@
...
+  # Must iterate on the index, vagrant does not understand iterating 
+  # over the node roles themselves
+  NODE_ROLES.length.times do |i|
+    name = NODE_ROLES[i]
+    config.vm.define name do |node|
+      roles = name.split("-", -1)
+      role_num = roles.pop.to_i
+      provision(node.vm, roles, role_num, i)
+    end
+  end
```
dereknola, line 97: Old way of doing this for loop
```suggestion
 NODE_ROLES.each_with_index do |name, i|
    config.vm.define name do |node|
      roles = name.split("-", -1)
      role_num = roles.pop.to_i
      provision(node.vm, roles, role_num, i)
    end
  end
```

dereknola (changes requested):

tests/e2e/ciliumnokp/Vagrantfile
manuelbuil: Thanks for this! I copied it from another test so there are more with this old loop. I'll change those too in another commit (for reviewing)

dereknola (approved):

tests/e2e/ciliumnokp/Vagrantfile
```diff
@@ -0,0 +1,95 @@
...
+    :libvirt__guest_ipv6 => "yes",
+    :libvirt__ipv6_address => "#{NETWORK6_PREFIX}::1",
+    :libvirt__ipv6_prefix => "64"
+    
+  vagrant_defaults = File.exists?("./vagrantdefaults.rb") ? "./vagrantdefaults.rb" : "../vagrantdefaults.rb"
+  load vagrant_defaults
+  
+  defaultOSConfigure(vm)
+  
+  scripts_location = Dir.exists?("./scripts") ? "./scripts" : "../scripts"
```
mgfritch, line 33:
```suggestion
  vagrant_defaults = File.exist?("./vagrantdefaults.rb") ? "./vagrantdefaults.rb" : "../vagrantdefaults.rb"
  load vagrant_defaults
  
  defaultOSConfigure(vm)
  
  scripts_location = Dir.exist?("./scripts") ? "./scripts" : "../scripts" 
```

mgfritch: Deprecated since ruby 2.1 and removed in ruby 3.2:
https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/

mgfritch (approved):

tests/e2e/ciliumnokp/Vagrantfile
manuelbuil: ah! Nice, thanks Mike! I'll prepare another commit replacing that in all Vagrantfiles

tests/e2e/ciliumnokp/Vagrantfile
```diff
@@ -0,0 +1,95 @@
...
+      rke2.env = %W[INSTALL_RKE2_TYPE=agent #{install_type}]
+      rke2.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
+      rke2.install_path = false
+      rke2.config = <<~YAML
+        write-kubeconfig-mode: '0644'
+        server: https://#{NETWORK4_PREFIX}.100:9345
+        node-ip: #{node_ip4},#{node_ip6}
+        node-external-ip: #{node_ip4},#{node_ip6}
+        token: vagrant-rke2
+        kubelet-arg: "--node-ip=0.0.0.0" # Workaround for https://github.com/kubernetes/kubernetes/issues/111695
```
rbrtbnfgl, line 68: I think that this is not needed anymore.

manuelbuil: ah! I had forgotten about this

rbrtbnfgl (approved):

mgfritch (approved):

PR #5886, opened 2024-05-06
Fix mixedosbgp e2e test

manuelbuil:
#### Proposed Changes ####

This PR improves the mixedosbgp e2e test. Main changes:
1 - Restarts the Windows RemoteAccess service after RKE2 is set-up. This is a known issue in Calico that appears often. Thanks @rbrtbnfgl 
2 - Waits for Windows node to be ready before continuing executing tests

#### Types of Changes ####

test e2e fix

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/testutils.go
+++ b/tests/e2e/testutils.go
@@ -323,6 +323,12 @@ func RunCmdOnNode(cmd string, nodename string) (string, error) {
 	return RunCommand(runcmd)
 }
 
+// RunCmdOnWindowsNode executes a command from within the given windows node
+func RunCmdOnWindowsNode(cmd string, nodename string) (string, error) {
+	runcmd := "vagrant ssh -c 'powershell.exe -Command \""+ cmd + "\"' " + nodename
+	return RunCommand(runcmd)
+}
+
 // RunCommand execute a command on the host
 func RunCommand(cmd string) (string, error) {
 	c := exec.Command("bash", "-c", cmd)
--- a/tests/e2e/mixedosbgp/mixedosbgp_test.go
+++ b/tests/e2e/mixedosbgp/mixedosbgp_test.go
@@ -40,7 +40,7 @@ func createMixedCluster(nodeOS string, serverCount, linuxAgentCount, windowsAgen
 	}
 	windowsAgentNames := []string{}
 	for i := 0; i < linuxAgentCount; i++ {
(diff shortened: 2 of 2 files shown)


ShylajaDevadiga (approved):

tests/e2e/testutils.go
```diff
@@ -323,6 +323,13 @@ func RunCmdOnNode(cmd string, nodename string) (string, error) {
 	return RunCommand(runcmd)
 }
 
+// RunCmdOnWindowsNode executes a command from within the given windows node
+func RunCmdOnWindowsNode(cmd string, nodename string) (string, error) {
+	communicator := "ssh"
+	runcmd := "vagrant " + communicator + " -c 'powershell.exe -Command \""+ cmd + "\"' " + nodename
+	return RunCommand(runcmd)
+}
+
```
dereknola, line 332: We currently override the default communicator for Windows VMs and set it to `winssh` at https://github.com/rancher/rke2/blob/master/tests/e2e/vagrantdefaults.rb#L9

Did you want to change the default there? Then we don't need this function.

dereknola (changes requested):

tests/e2e/testutils.go
manuelbuil: I was not aware of that, thanks!

The reason for this function is that I can't use the current [RunCmdOnNode](https://github.com/rancher/rke2/blob/master/tests/e2e/testutils.go#L319-L324) for Windows because of how picky powerhsell is with quotes. The command I need to run is:
vagrant winssh -c 'powershell.exe -Command "Restart-Service RemoteAccess"' windows-agent-0`. By creating this function I have the quotes correctly set and ready for any other command we might need to run in Windows in the future

dereknola: Okay, makes sense. But I don't think we need to define the communicator, since it will always be ssh
```suggestion
// RunCmdOnWindowsNode executes a command from within the given windows node
func RunCmdOnWindowsNode(cmd string, nodename string) (string, error) {
	runcmd := "vagrant ssh -c 'powershell.exe -Command \""+ cmd + "\"' " + nodename
	return RunCommand(runcmd)
}

```

dereknola (changes requested):

dereknola (approved):

Issue #5928, opened 2024-05-20
Support Traefik as ingress for RKE2


caroline-suse-rancher:
Some users would like the option to use Traefik as ingress. We do offer this on k3s, but it is now being asked for on RKE2. This would be an addition to our current supported ingress of nginx. 

Related Jira ticket is SURE-8403

manuelbuil:
It would be good to understand the reasons why somebody might choose one over the other to add it in the docs

mdrahman-suse:
# Validated on RC v1.30.3-rc5+rke2r1

## Testing
Scenarios with flag `ingress-controller: traefik` / `ingress-controller: traefik,ingress-nginx`
- HA cluster 
- New cluster
- Upgrade cluster 
- Air-gapped cluster
- CIS enabled cluster 
- Rancher deployed on cluster 
- Dual-stack cluster
- Cluster with Windows as agent

Issues found: 
- Airgap: https://github.com/rancher/rke2/issues/6438
- CIS enabled: https://github.com/rancher/rke2/issues/6449

AaronDewes:
I could not find the `ingress-controller` setting in the documentation, apart from the [ADR](https://github.com/rancher/rke2/blob/5b8d8ceeee6af2fec66afc0e9eeb0f7ded236bda/docs/adrs/008-traefik-ingress.md?plain=1#L16). Will this setting be added to the documentation with available options?

ceejatec:
@AaronDewes and anyone else who wanders this way: the way to enable this feature is to add the line

    ingress-controller: traefik

to `/etc/rancher/rke2/config.yaml` when first installing RKE2. There may be other ways as well, but I can confirm this works with RKE2 1.31.5+rke2r1. The cluster comes up with traefik configured as an ingress class, and no nginx-related pods or resources.

AaronDewes:
Thanks for mentioning it.

I understood it in general, but at the time when I wrote this comment, the only official information was this GitHub issue, while I would expect to see it featured in the official docs. Not sure if it's still not mentioned, but I guess I could try to create a docs PR myself if it isn't.

brandond:
Yeah, we do need to add this to the docs.

cupracer:
Hi, I'm trying to use this new `ingress-controller` option via Terraform as shown here (shortened version):

```
resource "rancher2_cluster_v2" "create_cluster" {
  name               = "dummy"
  kubernetes_version = "v1.31.7+rke2r1"

  rke_config {
[...]

PR #5943, opened 2024-05-21
Add multiple ingress controller support + traefik

brandond:
Waiting on:
* [ ] https://github.com/rancher/rke2-charts/pull/487
---

#### Proposed Changes ####

* Add ADR for multiple ingress controller support; initially with traefik as the only additional packaged ingress controller.
* Add CLI flag to select ingress controllers; multiple may be chosen, and `--ingress-controller=none` will disable all of them, same as `--cni=none`.
* Update charts

**Note:** while multiple ingress controllers can be selected via the CLI flag, and the first one will be set as the default ingress class, additional user-provided configuration will be necessary to allow both ingress controllers to run in parallel , due to both charts defaulting to deploying a DaemonSet that binds to ports 80 and 443. **For this reason, while allowed by the CLI flag, we will not be *supporting* parallel installation of multiple ingress controllers.**

If someone did want to do it on a cluster with a LoadBalancer controller capable of binding multiple LoadBalancer services on the same ports, something like this would work to reconfigure both controllers to use Deployment+LoadBalancer instead of Daemonset+HostPort:

```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rke2-ingress-nginx
  namespace: kube-system
spec:
  valuesContent: |-
    controller:
      kind: Deployment
      hostPort:
        enabled: false
      service:
        enabled: true
        type: LoadBalancer
---
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rke2-traefik
  namespace: kube-system
spec:
  valuesContent: |-
    deployment:
      kind: Deployment
    service:
      enabled: true
      type: LoadBalancer
    ports:
      web:
        hostPort: null
      websecure:
        hostPort: null
```

#### Types of Changes ####

ADR; packaged component

#### Verification ####

*  See linked issue and ADR
* 1. Start rke2 with one or more ingress controllers selected by --ingress-controller flag
   2. Note that selected ingress controller charts are installed and ingressClass resources created.
   3. Note that the first selected ingressClass has annotation `ingressclass.kubernetes.io/is-default-class: "true"`; others are false or absent.
   4. Note that Ingress resources are picked up by the default ingress controller if the ingressClassName is not set, and the selected ingress controller if ingressClassName is set to one of the installed controller names.

```console
brandond@dev01:~$ kubectl get node -o 'custom-columns=NAME:.metadata.name,NODE-ARGS:metadata.annotations.rke2\.io/node-args'
NAME            NODE-ARGS
rke2-server-1   ["server","--ingress-controller","traefik,ingress-nginx"]

brandond@dev01:~$ kubectl get ingressclass -o 'custom-columns=NAME:.metadata.name,CONTROLLER:.spec.controller,DEFAULT:.metadata.annotations.ingressclass\.kubernetes\.io/is-default-class'
NAME      CONTROLLER                      DEFAULT
nginx     k8s.io/ingress-nginx            <none>
traefik   traefik.io/ingress-controller   true
```

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/5928

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/008-traefik-ingress.md
+++ b/docs/adrs/008-traefik-ingress.md
@@ -1,0 +1,29 @@
+# Support for Alternative Ingress Controllers
+
+Date: 2024-05-21
+
+## Status
+
+Accepted
+
+## Context
+
+RKE2 currently supports only a single ingress controller, ingress-nginx.
+It has been requested RKE2 support alternative ingress controllers, similar to how RKE2 supports multiple CNIs. 
+
+## Decision
+
+* A new --ingress-controller flag will be added; the default will be only `ingress-nginx` to preserve current behavior.
+* All selected ingress controllers will be deployed to the cluster.
+* The first selected ingress controller will be set as the default, via the `ingressclass.kubernetes.io/is-default-class` annotation
+  on the IngressClass resource.
+* Any packaged ingress controllers not listed in the flag value will be disabled, similar to how inactive packaged CNIs are handled.
+* RKE2 will package Traefik's HelmChart as a supported ingress controller, deploying as a Daemonset + ClusterIP Service
+  for parity with the `ingress-nginx` default configuration due to RKE2's lack of a default LoadBalancer controller.
+* RKE2 will use mirrored upstream Traefik images; custom-rebuilt hardened-traefik images will not be provided or supported.
+
+## Consequences
+
+* We will add an additional packaged component and CLI flag for ingress controller selection.
+* We will need to track updates to Traefik and the Traefik chart.
+* QA will need additional resources to test the new ingress controllers.
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -1,6 +1,7 @@
 package cmds
 
 import (
+	"errors"
 	"strings"
 
 	"github.com/k3s-io/k3s/pkg/cli/cmds"
@@ -9,6 +10,7 @@ import (
 	"github.com/rancher/wrangler/v3/pkg/slice"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


pkg/cli/cmds/server.go
```diff
@@ -164,47 +156,82 @@ func ServerRun(clx *cli.Context) error {
...
-		clx.Set("disable", "rke2-multus")
-	case 2:
-		if cnis[0] == "multus" {
-			cnis = cnis[1:]
-		} else {
-			logrus.Fatal("invalid values provided for --cni flag: may only provide multiple values if multus is the first value")
+		return values, nil
+	})
+}
+
```
manuelbuil, line 195: Could we add a go doc comment for the future? Something like "disableExceptSelected disables rke2 non-default components unless they are enabled in the configuration"

pkg/cli/cmds/server.go
```diff
@@ -164,47 +156,82 @@ func ServerRun(clx *cli.Context) error {
...
+	if v, err := validateFunc(values); err != nil {
+		logrus.Fatalf("Failed to validate --%s flag: %v", flag.Name, err)
+	} else {
+		flag.Value = &v
+	}
+
+	// prepare a list of items to disable, based on all valid components.
+	// we have to use an intermediate set because the flag interface
+	// doesn't allow us to remove flag values once added.
+	disabledCharts := sets.Set[string]{}
```
manuelbuil, line 217: I wonder why we need a set instead of a slice. If I understand correclty, the "values" are always slices that we are hardcoding with rke2 components in "pkg/rke2/rke2.go", e.g. CNIItems

docs/adrs/008-traefik-ingress.md
```diff
@@ -0,0 +1,29 @@
...
+## Decision
+
+* A new --ingress-controller flag will be added that accepts a list of values; the default will be only `ingress-nginx` to preserve current behavior.
+* All listed ingress controllers will be deployed to the cluster.
+* The first listed ingress controller will be set as the default, via the `ingressclass.kubernetes.io/is-default-class` annotation
+  on the IngressClass resource.
+* Any packaged ingress controllers not listed in the flag value will be disabled, similar to how inactive packaged CNIs are handled.
+* RKE2 will package Traefik's HelmChart as a supported ingress controller, deploying as a Daemonset + ClusterIP Service
+  for parity with the `ingress-nginx` default configuration due to RKE2's lack of a default LoadBalancer controller.
+* RKE2 will use mirrored upstream Traefik images; custom-rebuilt hardened-traefik images will not be provided or supported.
```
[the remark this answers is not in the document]
brandond: as discussed in review, we only need to have hardened options - in this case we have a hardened core, a singled hardened CNI, and a single hardened ingress controller.

pkg/pebinaryexecutor/pebinary.go
```diff
@@ -105,7 +106,7 @@ func (p *PEBinaryConfig) Bootstrap(ctx context.Context, nodeConfig *config.Node,
 	}
 
 	if p.IsServer {
-		return bootstrap.UpdateManifests(p.Resolver, nodeConfig, cfg)
+		return bootstrap.UpdateManifests(p.Resolver, p.IngressController, nodeConfig, cfg)
```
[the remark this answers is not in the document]
brandond: I wired it up here in case we ever do support windows server nodes. I don't think it needs to be an error, it's just not a code path that is reachable at the moment.

pkg/cli/cmds/server.go
brandond: I made it a Set so that I can easily  entries from it by value, without having to find their indexes in the slice and delete them. It'd be much more boilerplate code than just the `Insert()` and `Delete()` functions that sets provide. I could use a map but then I'd need to extract the keys at the end - and these generic sets are implemented as a `map[T]struct{}` under the hood anyway.

pkg/cli/cmds/server.go
brandond: done

pkg/cli/cmds/server.go
manuelbuil: aaah ok, I was thinking on the property that no copies are possible in sets by default and that did not make sense here. Thanks

manuelbuil (approved):

dereknola (approved):

briandowns (approved):

PR #6098, opened 2024-06-04
Improve rke2-uninstall.ps1

manuelbuil:
#### Proposed Changes ####

After a QA iteration, we realized that the previous [PR](https://github.com/rancher/rke2/pull/5779) fixing the rke2 uninstallation is still not capable of removing all containerd resources in all environments. Main detected problems:

1 - Time out is too small for certain environments
2 - If we try removing the containerd namespace when there are resources there is an error message which could be confusing for some users

These PR does the following:
1 - Increases the timeout to 3 minutes
2 - It does not try to remove the containerd namespace until all the snapshots are gone. Note that the typical resources in a containerd namespace are tasks, containers, images and snapshots. We are removing the first three and once those three are removed, snapshots should be empty.
3 - If timeout is reached, we try to remove the snapshots directly. If the namespace can't still be removed, we warn the user that the uninstall script might leave some files behind in `/var/lib/rancher/rke2`
4 - It adds some extra logs and comments to clarify what's going on

#### Types of Changes ####

Bugfix

#### Verification ####

Install rke2 on linux and windows. Deploy a pod in windows. Run the rke2-uninstall.ps1 script. Without this PR, half of the time the script will fail and it will be impossible to remove the directory /var/lib/rancher.

With this PR, the script works and /var/lib/rancher does not exist anymore

#### Linked Issues ####

https://github.com/rancher/rke2/issues/5778

#### User-Facing Change ####

```release-note

```


rbrtbnfgl (approved):

bundle/bin/rke2-uninstall.ps1
```diff
@@ -301,20 +301,43 @@ function Remove-Containerd () {
...
         }
 
-        # Some resources in the namespace take a while to disappear. Try several times to remove the namespace and give up after 30s 
-        $endTime = (Get-Date).AddSeconds(30)
-        while ((Get-Date) -lt $endTime) {
+        # Resources in the namespace take a while to disappear. Snapshots are always the last to go.
+        # For 240s, snapshots are checked and namespace won't be removed until snapshots are gone.
+        # If timeout is reached, we will try removing snapshots directly. If that does not work, we stop trying and user is warned
+        $endTime = (Get-Date).AddSeconds(240)
+        $forceRemove = $false
```
mdrahman-suse, line 308: Suggestion to add here: `Write-Output "Tasks, containers, images and snapshots will be deleted. Timeout 240s`

bundle/bin/rke2-uninstall.ps1
```diff
@@ -301,20 +301,43 @@ function Remove-Containerd () {
...
+        while ($true) {
             $namespaces = $(Find-Namespaces)
             if ($namespaces) {
                 foreach ($ns in $namespaces) {
-                    Remove-Namespace $ns
+                    $snapshots = $(Find-Snapshots $ns)
+                    if ($snapshots) {
+                        Write-Output "Tasks, containers, images and snapshots still being deleted. Retrying (timeout 240s)"
```
mdrahman-suse, line 315: We can just say `Retrying resource deletion, waiting for 10s...` and remove timeout from here

bundle/bin/rke2-uninstall.ps1
manuelbuil: good idea

briandowns (approved):

PR #6142, opened 2024-06-06
Run unit tests on self-hosted runners if avaliable

dereknola:

#### Proposed Changes ####
- If running on rancher/rke2, use self-hosted runners for unit tests

#### User-Facing Change ####

```release-note

```


dereknola:
Shelf hosted runners have issues around download performance for go modules. Closing until a better solution is available.

PR #6251, opened 2024-06-25
[Bug] Pass install_type as a string in the mixedos e2e test

manuelbuil:
#### Proposed Changes ####

rke2.env is read by the [provisioner.rb](https://github.com/rancher/vagrant-rke2/blob/main/lib/vagrant-rke2/provisioner.rb#L126-L131) and if it is passed as an array, it adds an extra `-`, as a consequence, powershell is executing:
```
./install.ps1 --Commit (Get-Content -TotalCount 1 ./rke2_commits.txt)
```
i.e. the flag Commit has 2 dash symbols and that is wrong. As a consequence, powershell ignores it and passes the value as the first positional flag, which is Method. Then it fails because:
```
Cannot validate argument on parameter 'Method'. The argument "f9aa70bc03be5818efd9016a8d9de2ddadf45703" does not belong to the set "choco,tar" specified by the ValidateSet
```

In the end the e2e test fails because when it tries to check that `rke2.exe --version` is correct, it can't find rke2.exe: https://github.com/rancher/vagrant-rke2/blob/main/lib/vagrant-rke2/provisioner.rb#L162-L166

This PR passes the install_type parameter as a string. We write the dash symbol in Vagrantfile but vagrant-rke2 does not add an additional dash symbol

Apart from that, this PR adds a linux worker node to the mixedos test because otherwise tests/e2e/resource_files/pod_client.yaml can't run as it requires linux and 2 nodes

#### Types of Changes ####

Bugfix

#### Verification ####

Run the mixedos or mixedosbgp test and verify that rke2 is installed

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/mixedos/Vagrantfile
+++ b/tests/e2e/mixedos/Vagrantfile
@@ -95,7 +95,7 @@ def provision(vm, role, role_num, node_num)
     # end
     # If using libvirt, use virt-viewer for GUI after bring up the node
     vm.provision :rke2, run: 'once' do |rke2|
-      rke2.env = [install_type]
+      rke2.env = install_type
       rke2.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
       # rke2.skip_start = true
       rke2.config = <<~YAML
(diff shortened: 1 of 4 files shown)


dereknola (approved):

dereknola (approved):

brandond (approved):

manuelbuil:
As agreed with @rancher-max , I can merge this to fix CI even if we are still on freeze :)

PR #6274, opened 2024-07-02
Update Vagrantfile of a few e2e tests

manuelbuil:
#### Proposed Changes ####

Remove the bind-address arg which is not needed and could be problematic if we use `kubectl exec ...` like in the cilium e2e test due to https://github.com/k3s-io/k3s/issues/10444

Besides, `kubelet-arg: "--node-ip=0.0.0.0"` is not needed anymore. That was fixed upstream

#### Types of Changes ####

Update e2e tests

#### User-Facing Change ####

```release-note

```


dereknola (approved):

brandond (approved):

manuelbuil:
Merging with Brooks' and Max' blessing

PR #6280, opened 2024-07-05
Refactor run_tests.sh script

manuelbuil:
#### Proposed Changes ####

This PR adds some extra logging in the latest_commit.sh script. If there was an error because of the API exceeding the limits, it was impossible to notice

Moreover, it refactors run_tests.sh script trying to make it simpler. Changes:

1 - Removes all the parameters that can be passed except for nodeOS, which is important for the results generation. Some of them depend on the test itself and should not be changed (servercount, agentcount). Some are not read by any test (db or hardened)

2 - Removes `rke2_version` and `rke2_channel` variables and references as the former was always empty and the latter was always `commit`, which is anyway the default: https://github.com/rancher/rke2/blob/master/tests/e2e/vagrantdefaults.rb#L19-L22

3 - Creates a loop that goes over an array with the different e2e tests available instead of calling each one in a different line

4 - Uses a 2h timeout for all e2e tests. In the end, the important timeout is the one happening in each Test Case (in the Eventually instruction)

5 - Slightly changes the "VM gargabe collection". It was detected that sometimes, VMs are not really cleaned up. That should always happen as we run the tests with the `-ci`. We suspect it's a problem related to the timeout, that's why we increased it to 2h as previously explained. Just in case, before starting the e2e test, we destroy the VMs related to that test. At the beginning of the script, we also destroy any VM which might be in an invalid state with `vagrant global-status --prune`

6 - The log file includes now the date in the name instead of the OS

7 - Upgrades all images to Ubuntu2310

Before running each test, it makes sure that VMs are destroyed. 

#### Types of Changes ####

Code refactor

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/scripts/run_tests.sh
+++ b/tests/e2e/scripts/run_tests.sh
@@ -1,22 +1,25 @@
 #!/bin/bash
-nodeOS=${1:-"generic/ubuntu2004"}
-servercount=${2:-3}
-agentcount=${3:-1}
-db=${4:-"etcd"}
-hardened=${5:-""}
-rke2_version=${rke2_version}
-rke2_channel=${rke2_channel:-"commit"}
+# Usage: ./run_tests.sh
+# This script runs all the rke2 e2e tests and generates a report with the log
+# The generated log is placed in createreport/rke2_${date}.log
+#
+# This script must be run inside the rke2 directory where the tests exist
+#
+# Example:
+#   To run the script with default settings:
+#     ./run_tests.sh
+#
+set -x
 
-E2E_EXTERNAL_DB=$db && export E2E_EXTERNAL_DB
+# tests to run
+tests=("ciliumnokp" "dnscache" "dualstack" "mixedos" "mixedosbgp" "multus" "secretsencryption" "splitserver" "upgradecluster" "validatecluster")
+date=$(date +%Y%m%d)
+
 E2E_REGISTRY=true && export E2E_REGISTRY
 
-cd
-cd rke2 && git pull --rebase origin master
+git pull --rebase origin master
 /usr/local/go/bin/go mod tidy
-
 cd tests/e2e
-OS=$(echo "$nodeOS"|cut -d'/' -f2)
-echo "$OS"
 
 # create directory to store reports if it does not exists
 if [ ! -d createreport ]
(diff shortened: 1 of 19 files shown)


thomasferrandiz (approved):

dereknola (approved):

tests/e2e/scripts/run_tests.sh
```diff
@@ -1,62 +1,60 @@
...
-	E2E_HARDENED="$hardened" /usr/local/go/bin/go test -v splitserver/splitserver_test.go -nodeOS="$nodeOS" -timeout=30m -json -ci |tee -a createreport/rke2_"$OS".log
+        echo "RUNNING ${tests[$i]} TEST"
+        /usr/local/go/bin/go test -v ${tests[$i]}_test.go -timeout=1h -json -ci |tee -a ../createreport/rke2_${date}.log
+        
+	popd
+    done
 }
 
-ls createreport/rke2_"$OS".log 2>/dev/null && rm createreport/rke2_"$OS".log
+ls createreport/rke2_${date}.log 2>/dev/null && rm createreport/rke2_${date}.log
```
ShylajaDevadiga, line 49: Can we leave it as OS to avoid cleanup of s3 bucket and preferably one sharable link to results for different OS?

tests/e2e/scripts/run_tests.sh
```diff
@@ -1,62 +1,60 @@
...
-	echo 'RUNNING DUALSTACK VALIDATION TEST'
-	E2E_HARDENED="$hardened" /usr/local/go/bin/go test -v dualstack/dualstack_test.go -nodeOS="$nodeOS" -serverCount=1 -agentCount=1  -timeout=30m -json -ci |tee -a createreport/rke2_"$OS".log
 
-	echo 'RUNNING CLUSTER VALIDATION TEST'
-	E2E_REGISTRY=true E2E_HARDENED="$hardened" /usr/local/go/bin/go test -v validatecluster/validatecluster_test.go -nodeOS="$nodeOS" -serverCount=$((servercount)) -agentCount=$((agentcount))  -timeout=30m -json -ci |tee -a createreport/rke2_"$OS".log
+    count=$(( count + 1 ))
 
-	echo 'RUNNING MIXEDOS TEST'
-	/usr/local/go/bin/go test -v mixedos/mixedos_test.go -nodeOS="$nodeOS" -serverCount=$((servercount)) -timeout=1h -json -ci |tee -a  createreport/rke2_"$OS".log
+    for i in ${!tests[@]}; do
```
ShylajaDevadiga, line 38: Some tests have just server nodes and no agents. Some have HA and some single server single agent. Can we default then in Vagrantfile?

manuelbuil: Yes, vagrantfile defaults the number of servers and agents per test. Same for the OS

tests/e2e/scripts/run_tests.sh
manuelbuil: ok, done! I moved all tests to Ubuntu2310 (same like in k3s)

dereknola (approved):

ShylajaDevadiga (approved):

Issue #6312, opened 2024-07-12
Install.sh on does not work on aarch64 RPM systems


TomWis97:
**Environmental Info:**
RKE2 Version: v1.28.11+rke2r1

Node(s) CPU architecture, OS, and Version: 
In this case RHEL 9, but this occures on every aarch64 machine.
`Linux ip-x-x-x-x.eu-west-3.compute.internal 5.14.0-427.20.1.el9_4.aarch64 #1 SMP PREEMPT_DYNAMIC Thu May 23 21:07:05 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux`

Cluster Configuration:
(Not relevant.) 1 Server.

**Describe the bug:**
When running the installation script from https://get.rke2.io on a aarch64 machine with an RPM-based package manager, installation fails.  
The script places a repository config, referring to the x86_64 repository. Obviously, packages from that repo cannot be installed on aarch64 machines. So the script fails, leaving an incorrect Yum configuration.
Also, ARM64 RPM packages may not be available, but I think that's out of scope for this bug.

**Steps To Reproduce:**

- Have a aarch64 machine with a RPM-based distribution.
- Run `curl https://get.rke2.io | sudo sh`.

**Expected behavior:**
- If aarch64 RPMs are available, the script should refer to the repository containing the correct architecture.
- If aarch64 RPMs are _not_ available: fall back to .tar installation method.

**Actual behavior:**
- Yum repository configuration referring to incorrect architecture is placed.
- Installation fails.

**Additional context / logs:**

N/A.

brandond:
Duplicate of https://github.com/rancher/rke2/issues/5135 which I have installed.

Issue #5135, opened 2023-12-14
RKE2 Installation Fails on Rocky Linux with aarch64 Architecture


khushalchandak17:
<!-- Thanks for helping us to improve RKE2! We welcome all bug reports. Please fill out each area of the template so we can better help you. Comments like this will be hidden when you post but you can delete them if you wish. -->

**Environmental Info:**
RKE2 Version: 
- rke2-v1.27.3 and above
- rke2-v1.28

Node(s) CPU architecture, OS, and Version: 
<!-- Provide the output from "uname -a" on the node(s) -->

First verified on the latest Rocky Linux 9.3 after I saw that this OS is validated [here](https://github.com/rancher/rke2/issues/5101):
~~~
# uname -a
Linux localhost.localdomain 5.14.0-362.8.1.el9_3.aarch64 #1 SMP PREEMPT_DYNAMIC Thu Nov 9 05:21:32 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
~~~


Second verified on the supported OS Rocky Linux 8.7:
~~~
# cat /etc/os-release
NAME="Rocky Linux"
VERSION="8.7 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.7 (Green Obsidian)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-8"
ROCKY_SUPPORT_PRODUCT_VERSION="8.7"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
[root@ip-10-0-16-58 rocky]# uname -m
aarch64
~~~

Cluster Configuration:
<!-- Provide some basic information on the cluster configuration. For example, "3 servers, 2 agents". -->
Facing an issue during RKE2 installation. Sharing two scenarios as follows:

Scenario-1 On Validated Rocky Linux 9.3:
~~~
[root@localhost verify]# uname -a
[...]

PR #6313, opened 2024-07-13
Install.sh: Only use RPM installation method on AMD64

TomWis97:
#### Proposed Changes ####

As described in issue #5135, the `install.sh` script fails on a RPM-based distribution on aarch64 systems. As the RPM method only supports AMD64, I propose we check if the architecture is indeed AMD64 and continue with the tar installation method if that's not the case.

We should clarify in the documentation that the RPM installation is at the moment only supported on x86_64.

#### Types of Changes ####
This is a bugfix, or, more like a workaround until the RPM method further matures.

#### Verification ####
1. Run the script on a x64_64 system with a RPM-based distro. Observe the RPM installation method is chosen.
2. Run the script on a aarch64 system with a RPM-based distro. Observe the tar installation method is chosen.

#### Testing ####
I believe this is not required for such a small change.

#### Linked Issues ####
#6312 and #5135 

#### User-Facing Change ####

```release-note
Installation script only uses the RPM installation method on x86_64-based systems.
```

#### Further Comments ####
None.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -116,8 +116,8 @@ setup_env() {
         INSTALL_RKE2_TYPE="${INSTALL_RKE2_EXEC:-server}"
     fi
 
-    # --- use rpm install method if available by default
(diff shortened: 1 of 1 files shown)


brandond:
Fixed the "right way" by:
* https://github.com/rancher/rke2/pull/6927

PR #6387, opened 2024-07-22
Add missing package windows step in release

galal-hussein:
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,7 +63,7 @@ jobs:
         DOCKER_USERNAME: ${{ env.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ env.DOCKER_PASSWORD }}
     
-    - name: Pacakge windows images
+    - name: Package windows images
       run: |
         GITHUB_ACTION_TAG=${{ github.ref_name }} dapper -f Dockerfile --target dapper make package-windows-images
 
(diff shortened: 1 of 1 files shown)


rafaelbreno (approved):

briandowns (approved):

.github/workflows/release.yml
```diff
@@ -63,6 +63,10 @@ jobs:
         DOCKER_USERNAME: ${{ env.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ env.DOCKER_PASSWORD }}
     
+    - name: Pacakge windows images
```
brandond, line 66: Spelling nit

```suggestion
    - name: Package windows images
```

Also, shouldn't this go above, either directly below the existing `Package Images` step, or perhaps just even another `run` line in that same step?

brandond (changes requested):

.github/workflows/release.yml
galal-hussein: I will fix the typo, however the order is actually what it used to be in drone
https://github.com/rancher/rke2/blob/v1.28.11-rc6%2Brke2r1/.drone.yml#L160C1-L172C1

brandond: ok!

We need to review the makefile when we get rid of dapper, I think we are rebuilding the same stuff many times.

brandond (approved):

PR #6483, opened 2024-08-06
Don't overwrite existing rke2-pss.yaml if one exists

ethanchowell:
#### Proposed Changes ####

Our current workflows involve building a base image with a lot of the configs baked in. I noticed recently after upgrading to 1.28 that even though we're putting a custom config at the default location `/etc/rancher/rke2/rke2-pss.yaml`, RKE2 is overwriting it no matter what. This PR adds a check similar to the check for an existing audit-policy.yaml to respect files at the default location,

#### Types of Changes ####

Bugfix

#### Verification ####

Create a default config at `/etc/rancher/rke2/rke2-pss.yaml`. A good sample is the [Rancher sample](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions). Start/restart rke2-server and you'll see the file overwritten.

### DIFF
--- a/pkg/rke2/psa.go
+++ b/pkg/rke2/psa.go
@@ -17,6 +17,11 @@ const (
 // CIS mode. For CIS mode, a default PSA configuration with enforcement for restricted will be applied
 // for non CIS mode, a default PSA configuration will be applied that has privileged restriction
 func setPSAs(cisMode bool) error {
+	if _, err := os.Stat(defaultPSAConfigFile); os.IsExist(err) {
+		logrus.Warnf("existing rke2-pss.yaml found at %s", defaultPSAConfigFile)
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


dereknola:
If you want your own PSS, it is expected that you will pass your own flag to a custom pss file, not use our default location:
```
#/etc/rancher/rke2/config.yaml
kube-apiserver-arg:
- "pod-security-admission-config-file=/etc/my/custom/pss.yaml"
```

brandond:
We also have a top-level `rke2 server` flag: https://docs.rke2.io/reference/server_config#security
> pod-security-admission-config-file | Path to the file that defines Pod Security Admission configuration | RKE2_POD_SECURITY_ADMISSION_CONFIG_FILE
> -- | -- | --

Issue #6705, opened 2024-09-03
Update SR-IOV chart to `v1.3.0`


mgfritch:
Backport https://github.com/suse-edge/charts/pull/145 into the `rancher/charts` chart repository

thomasferrandiz:
replaced by https://github.com/rancher/rke2/issues/6746

PR #6706, opened 2024-09-04
Update sriov images to `v1.3.0`

mgfritch:
Also updates node-feature-discovery to `0.15.6`
Issue: https://github.com/rancher/rke2/issues/6705

Chart releases:
- SUSE Edge: https://github.com/suse-edge/charts/pull/145
- Rancher: https://github.com/rancher/charts/pull/4394

### DIFF
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -83,14 +83,14 @@ fi
 xargs -n1 -t docker image pull --quiet << EOF > build/images-multus.txt
     ${REGISTRY}/rancher/hardened-multus-cni:v4.0.2-build20240612
     ${REGISTRY}/rancher/hardened-cni-plugins:v1.4.1-build20240430
-    ${REGISTRY}/rancher/hardened-node-feature-discovery:v0.15.4-build20240513
-    ${REGISTRY}/rancher/hardened-sriov-network-operator:v1.2.0-build20240327
-    ${REGISTRY}/rancher/hardened-sriov-network-config-daemon:v1.2.0-build20240327
-    ${REGISTRY}/rancher/hardened-sriov-network-device-plugin:v3.6.2-build20240327
-    ${REGISTRY}/rancher/hardened-sriov-cni:v2.7.0-build20240327
-    ${REGISTRY}/rancher/hardened-ib-sriov-cni:v1.0.3-build20240327
-    ${REGISTRY}/rancher/hardened-sriov-network-resources-injector:v1.5-build20240327
-    ${REGISTRY}/rancher/hardened-sriov-network-webhook:v1.2.0-build20240327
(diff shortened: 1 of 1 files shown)


thomasferrandiz (approved):

brandond (changes requested):
Where's the corresponding chart bump? Just bumping the images in the tarball doesn't change what the chart uses.

mgfritch:
> Where's the corresponding chart bump? Just bumping the images in the tarball doesn't change what the chart uses.

The SR-IOV chart is currently maintained in the SUSE edge repo and we've added a [deprecation warning](https://github.com/rancher/charts/pull/4025) for it in the rancher chart repo.

- SUSE edge: https://github.com/suse-edge/charts/pull/145
- Rancher Charts: https://github.com/rancher/charts/pull/4394

brandond:
OK but how do we ensure that the images included in the airgap tarball are the same ones referenced by the chart? Do the RKE2 docs need to cover what specific version of the SR-IOV chart its airgap images are valid with? What versions of the chart are these NEW images valid with? I see that https://docs.rke2.io/networking/multus_sriov#using-multus-with-sr-iov just points to rancher charts, but it doesn't tell you what version of the chart to deploy.

rbrtbnfgl:
I think we can't. SR-IOV chart is installed through helm and it's not associated to the specific RKE2 version. If installed with rancher is associated to the rancher version from what I know.

brandond:
I'm confused why we have SRIOV images in our Multus tarball, if they are no longer bundled with Multus or in any way tied to the Multus chart version.

I feel like we should either remove these from the multus tarball, or call out in the release notes what version of the SRIOV chart the images are meant to be used with.

rbrtbnfgl:
@thomasferrandiz has more knowledge of SRIOV

thomasferrandiz:
@brandond I think you're right and the images were added in the multus tarball because at first sriov was supposed to be bundled and installed with multus.
Since the sriov chart had to be moved in Rancher later, we just ended up in a situation where we hope that the version of the images in the airgap tarball matches the version used by the chart in the Rancher repo.

We could add the versions to the release notes to help users understand if they have the right versions.
Is it done only through the PR description?

brandond:
This needs to go deeper than just PR titles or descriptions. 

I think we should do one of two things:
1. - Remove the SRIOV images from the multus tarball, since multus and sriov are no longer tied together.
2. - Work with the QA and release teams to figure out how to best indicate which version of the sriov chart the multus airgap tarball is meant to be used with, and add this to the component version table in the release notes.
    - Add CI checks to ensure that the images are correct for the indicated chart version - like we have for other chart images.

thomasferrandiz:
As discussed, it is simpler to just remove the sriov images from the airgap tarball: https://github.com/rancher/rke2/pull/6747

Issue #6926, opened 2024-10-03
Can't Install RKE2 on ARM64 Architecture


ccravens:
**Environmental Info:**
RKE2 Version: 
v1.28.3~rke2r1

Node(s) CPU architecture, OS, and Version: 
Rocky 8.10 aarch64

Cluster Configuration:
3 masters 6 workers

**Describe the bug:**
I'm running the following command as root:
```bash
curl -sfL https://get.rke2.io/ | INSTALL_RKE2_TYPE="server" INSTALL_RKE2_VERSION="v1.28.3+rke2r1" sh -
```

However I get the following error:
```
[INFO]  using 1.28 series from channel stable
Rancher RKE2 Common (stable)                     21 kB/s | 2.9 kB     00:00    
Rancher RKE2 1.28 (stable)                       23 kB/s | 2.9 kB     00:00    
Error: 
 Problem: conflicting requests
  - package rke2-server-1.28.3~rke2r1-0.el8.x86_64 from rancher-rke2-1.28-stable does not have a compatible architecture
  - nothing provides rke2-common = 1.28.3~rke2r1-0.el8 needed by rke2-server-1.28.3~rke2r1-0.el8.x86_64 from rancher-rke2-1.28-stable
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

Even though in the https://get.rke2.io/ script, I see this:

```bash
# setup_arch set arch and suffix,
# fatal if architecture not supported.
setup_arch() {
    case ${ARCH:=$(uname -m)} in
    x86_64|amd64)
        ARCH=amd64
        SUFFIX=$(uname -s | tr '[:upper:]' '[:lower:]')-${ARCH}
        ;;
    aarch64|arm64)
        ARCH=arm64
        SUFFIX=$(uname -s | tr '[:upper:]' '[:lower:]')-${ARCH}
        ;;
    s390x)
        ARCH=s390x
        SUFFIX=$(uname -s | tr '[:upper:]' '[:lower:]')-${ARCH}
        ;;
    *)
        fatal "unsupported architecture ${ARCH}"
        ;;
    esac
}
```

Which returns the correct value of aarch64:

```bash
[root@localhost ~]# uname -m
aarch64
```

Why is it pulling x86_64 packages when installing on an aarch64 architecture?

**Steps To Reproduce:**
Run the following on an ARM64 build of Rocky 8.10:

```bash
[...]

PR #6927, opened 2024-10-03
Use correct architecture for yum repo instead of hardcoding x86_64 

ccravens:
#### Proposed Changes ####
I'm working on fixing the RPM install process for ARM64 `aarch64` architecture type.

Does not require a change to docs

#### Types of Changes ####
Minor bug fixes to enable RPM installs of `aarch64`

#### Verification ####
Run the following command on an ARM64 version of Rocky 8:
```
curl -sfL https://get.rke2.io/ | INSTALL_RKE2_TYPE="server" INSTALL_RKE2_VERSION="v1.28.14+rke2r1" sh -
```

#### Testing ####
Not covered by testing

#### Linked Issues ####
https://github.com/rancher/rke2/issues/6926

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####
Quick fix, not a major change

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -562,7 +562,7 @@ EOF
     cat <<-EOF >>"${repodir}/rancher-rke2.repo"
 [rancher-rke2-${rke2_majmin}-${rke2_rpm_channel}]
 name=Rancher RKE2 ${rke2_majmin} (${1})
-baseurl=https://${rpm_site}/rke2/${rke2_rpm_channel}/${rke2_majmin}/${rpm_site_infix}/x86_64
+baseurl=https://${rpm_site}/rke2/${rke2_rpm_channel}/${rke2_majmin}/${rpm_site_infix}/${ARCH}
 enabled=1
 gpgcheck=1
 repo_gpgcheck=0
--- a/install.sh
+++ b/install.sh
@@ -562,7 +562,7 @@ EOF
     cat <<-EOF >>"${repodir}/rancher-rke2.repo"
 [rancher-rke2-${rke2_majmin}-${rke2_rpm_channel}]
 name=Rancher RKE2 ${rke2_majmin} (${1})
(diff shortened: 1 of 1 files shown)


brandond (approved):
thanks!

install.sh
```diff
@@ -562,7 +562,7 @@ EOF
     cat <<-EOF >>"${repodir}/rancher-rke2.repo"
 [rancher-rke2-${rke2_majmin}-${rke2_rpm_channel}]
 name=Rancher RKE2 ${rke2_majmin} (${1})
-baseurl=https://${rpm_site}/rke2/${rke2_rpm_channel}/${rke2_majmin}/${rpm_site_infix}/x86_64
+baseurl=https://${rpm_site}/rke2/${rke2_rpm_channel}/${rke2_majmin}/${rpm_site_infix}/${ARCH}
```
brandond, line 565: Actually I think you want to use `uname -m` as the $ARCH is a docker/golang arch, not the machine type as used by RPM. Ref: 
https://github.com/rancher/rke2/blob/bf1bcfdcbe0d37b8a6f64c25b47b7e096e6efedb/install.sh#L164-L167
```suggestion
baseurl=https://${rpm_site}/rke2/${rke2_rpm_channel}/${rke2_majmin}/${rpm_site_infix}/$(uname -m)
```

brandond (changes requested):

cwayne18:
@vitorsavian is this still an issue after your recent PRs?

install.sh
ccravens: Yes `uname -m` is the way to go

brandond:
I suspect we didn't actually test installing RPMs using the install script on arm64. We only checked to see that they are built and published.

ccravens:
Tested the patch with `(uname -m)` on my Rocky 8 ARM64 and it seems to install OK

briandowns (approved):

Issue #4238, opened 2023-05-15
Add metadata labels to rke2 images


matttrach:
We would like the metadata for rke2 images to reflect the standard org.opencontainers.image.url and org.opencontainers.image.source information to easily trace image data.

brandond:
Ref: https://github.com/opencontainers/image-spec/blob/main/annotations.md

matttrach:
We would also like to include `org.opencontainers.image.base.name`

matttrach:
Example Implementation:
```
% docker inspect registry.suse.com/bci/bci-base:latest | grep "org.opencontainers.image"
                "org.opencontainers.image.created": "2022-12-11T11:44:12.117506595Z",
                "org.opencontainers.image.description": "Image for containers.",
                "org.opencontainers.image.source": "https://sources.suse.com/...",
                "org.opencontainers.image.title": "Base Container Image",
                "org.opencontainers.image.url": "https://www.suse.com/products/server/",
                "org.opencontainers.image.vendor": "SUSE LLC",
                "org.opencontainers.image.version": "1.1.1.1.1",
```

brandond:
Not sure we've got EVERYTHING here but with the rke2-runtime image I think we can at least have QA validate that the main image has the label.

ShylajaDevadiga:
### Validated using v1.31.1-rc1+rke2r2
## Environment Details
*Infrastructure*
Cloud EC2 instance
*Node(s) CPU architecture, OS, and Version:*
$ cat /etc/os-release 
```
PRETTY_NAME="Ubuntu 24.04 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04 LTS (Noble Numbat)"
```
*Cluster Configuration:*
Single node

*Config.yaml:*
```
cat /etc/rancher/rke2/config,yaml
write-kubeconfig-mode: "0644"
```
*Validation results:*
```
ubuntu@ip-172-31-11-1:~$ rke2 -v
rke2 version v1.31.1-rc1+rke2r2 (a1d5456aec21b761ac85f0348a1926290e48c38b)
go version go1.22.6 X:boringcrypto
ubuntu@ip-172-31-11-1:~$ sudo docker inspect docker.io/rancher/rke2-runtime:v1.31.1-rc1-rke2r2 |grep open
                "org.opencontainers.image.source": "https://github.com/rancher/rke2",
[...]

PR #6962, opened 2024-10-09
add org.opencontainers.image url and source labels to dockerfiles

sankalp-at-gh:
#### Proposed Changes ####

This PR adds the image metadata labels to rke2-runtime image

#### Types of Changes ####

enhancement

#### Verification ####

check rke2-runtime image labels

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/4238

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -149,6 +149,8 @@ COPY --from=runtime-collect / /
 
 FROM ubuntu:24.04 AS test
 ARG TARGETARCH
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/rancher/rke2-runtime"
+LABEL org.opencontainers.image.source="https://github.com/rancher/rke2"
 VOLUME /var/lib/rancher/rke2
 VOLUME /var/lib/kubelet
 VOLUME /var/lib/cni
--- a/Dockerfile
+++ b/Dockerfile
@@ -145,12 +145,12 @@ COPY --from=charts \
     /charts/
 
 FROM scratch AS runtime
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/rancher/rke2-runtime"
+LABEL org.opencontainers.image.source="https://github.com/rancher/rke2"
 COPY --from=runtime-collect / /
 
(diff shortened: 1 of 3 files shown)


sankalp-at-gh:
Hi @brandond and @matttrach, could you please help with review and approval of this PR.
Thanks a lot !

Dockerfile.docs
```diff
@@ -1,4 +1,6 @@
 FROM squidfunk/mkdocs-material
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/rancher/rke2-runtime"
+LABEL org.opencontainers.image.source="https://github.com/rancher/rke2"
```
brandond, line 3: This Dockerfile is no longer used; we were just using it to build the docs website before it was moved out to a different repo - to labeling it as published as this image is incorrect.

brandond (changes requested):

Dockerfile.docs
sankalp-at-gh: totally makes sense, thanks for the feedback !

brandond (approved):

Dockerfile
```diff
@@ -149,6 +149,8 @@ COPY --from=runtime-collect / /
 
 FROM ubuntu:24.04 AS test
 ARG TARGETARCH
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/rancher/rke2-runtime"
```
brandond, line 152: You've added these to the `test` target which is not published. I think you probably want to move them to the `runtime` step a couple lines up, which is the step that builds the actual published rke2-runtime linux image.

brandond (changes requested):
one more issue - I didn't notice that the labels were on the wrong step in the main Dockerfile.

Dockerfile
sankalp-at-gh: Thank you for point it out, it was not very obvious when I first looked at the dockerfile

brandond (approved):

vitorsavian (approved):

PR #6977, opened 2024-10-09
Bump klipper-helm to v0.9.3-build20241008

cwayne18:


brandond:
You can't just update the image. Helm-controller needs to be bumped in k3s first, and then pulled through to rke2. The updated helm-controller version will use the new image.

I was planning to bump k3s in rke2 later this week when backports are all in.

PR #6978, opened 2024-10-09
Bump some mirrored images

cwayne18:
bumpy bump

these are all already in the image mirror


brandond (changes requested):
The LH chart will need to be bumped to use these new images. Just changing the tag that we put in the tarball will not update the charts to us the new tags.

cwayne18:
ignore this im dumb

PR #7089, opened 2024-10-21
rke2-runtime signing and manifests

brooksn:
#### Proposed Changes ####

* Updates the release workflow to use the publish-image action to run image/manifest push scripts
* Pushes images directly to the prime registry
* Moves the windows image push into a separate step
* Adds SBOM to image manifests

#### User-Facing Change ####

```release-note

```


scripts/publish-image-runtime
```diff
@@ -5,11 +5,21 @@ cd $(dirname $0)/..
...
-docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
-set -x
-
-docker image push ${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-${GOOS}-${GOARCH}
-if [ "${GOARCH}" != "s390x" ] && [ "${GOARCH}" != "arm64" ]; then
-    docker image push ${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}-windows-amd64
-fi
+./scripts/setup-docker-builder
+
+DOCKER_BUILDKIT=${DOCKER_BUILDKIT:-1} docker image build ${IID_FILE_FLAG} \
```
brooksn, line 10: The publish-image action calls the make target, and sets IID_FILE_FLAG for use by cosign.

https://github.com/rancher/ecm-distro-tools/blob/master/actions/publish-image/action.yaml#L166-L174

scripts/setup-docker-builder
```diff
@@ -0,0 +1,11 @@
...
+set -ex
+
+source ./scripts/version.sh
+
+BUILDER=${PROG}
+
+docker buildx use ${BUILDER}|| \
+    docker buildx create --name=${BUILDER} --platform=linux/amd64,linux/arm64
+
+docker buildx use ${BUILDER}
```
dereknola, line 11: What is the reason for setting up this buildx builder vs using the generic setup-buildx action? Is it just to maintain everything inside `/script`?
```
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v3
```

brooksn: I think you're right, setup-buildx is a composite step being run already, and it should use a builder automatically.

dereknola (approved):

briandowns (approved):

Issue #6571, opened 2024-08-15
rke2 start fails after rke2-killall.sh execution


mynktl:
**Environmental Info:**
RKE2 Version: 

rke2 version v1.30.1+rke2r1 (e7f87c6dd56fdd76a7dab58900aeea8946b2c008)
go version go1.22.2 X:boringcrypto

Node(s) CPU architecture, OS, and Version: 

Linux server0 4.18.0-372.41.1.el8_6.x86_64 #1 SMP Thu Jan 5 13:56:06 EST 2023 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

1 server

**Describe the bug:**

For RKE2, We are using two mount points, one for etcd database and other for rke2.
structure of `/var/lib/rancher` directory is as below:
```
[root@server0 rancher]# tree /var/lib/rancher/ -L 3
/var/lib/rancher/
`-- rke2
    |-- agent
    |   |-- client-ca.crt
    |   |-- client-kubelet.crt
    |   |-- client-kubelet.key
    |   |-- client-kube-proxy.crt
    |   |-- client-kube-proxy.key
    |   |-- client-rke2-controller.crt
    |   |-- client-rke2-controller.key
    |   |-- containerd
    |   |-- etc
    |   |-- images
    |   |-- kubelet.kubeconfig
    |   |-- kubeproxy.kubeconfig
    |   |-- logs
    |   |-- pod-manifests
    |   |-- rke2controller.kubeconfig
    |   |-- server-ca.crt
    |   |-- serving-kubelet.crt
    |   `-- serving-kubelet.key
    |-- bin -> /var/lib/rancher/rke2/data/v1.30.1-rke2r1-c42b85364830/bin
    |-- data
    |   `-- v1.30.1-rke2r1-c42b85364830
    `-- server
        |-- agent-token -> /var/lib/rancher/rke2/server/token
        |-- cred
        |-- db
        |-- etc
        |-- manifests
        |-- node-token -> /var/lib/rancher/rke2/server/token
        |-- tls
        `-- token

16 directories, 16 files
```

and mount configuration is as below:
```
[root@server0 rancher]# lsblk
NAME                          MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sda                             8:0    0  128G  0 disk
|-sda1                          8:1    0  500M  0 part /boot
|-sda2                          8:2    0   63G  0 part
| |-rootvg-tmplv              253:11   0    2G  0 lvm  /tmp
[...]

PR #7134, opened 2024-10-24
Only clean up pod filesystems if umount succeeds

brandond:
#### Proposed Changes ####

Only clean up pod filesystems if umount succeeds.

If the unmount fails for some reason we shouldn't `rm -rf` the mountpoint, there might still be stuff under it.

#### Types of Changes ####

killall script; bugfix

#### Verification ####

See linked issue

#### Linked Issues ####
* https://github.com/rancher/rke2/issues/6571#issuecomment-2294339626

#### User-Facing Change ####

```release-note
```


brandond:
I will note that we do this much more concisely in k3s, although it requires `xargs` which we have so far avoided requiring for RKE2.

https://github.com/k3s-io/k3s/blob/v1.31.1%2Bk3s1/install.sh#L839-L845

manuelbuil (approved):

Issue #7193, opened 2024-11-06
[Release 1.31] Bump multus, cilium and flannel charts


manuelbuil:
Backport: https://github.com/rancher/rke2/issues/7090

mdrahman-suse:
# Validation on release-1.31 with commit a97905b2

### Cilium, CNI Plugins
```
[2024-11-06T22:27:44.242Z] rke2 version v1.31.2+dev.a97905b2 (a97905b2f7ba0a6098548b578adc1418c6fecaea)
[2024-11-06T22:27:44.242Z] go version go1.22.8 X:boringcrypto 
[2024-11-06T22:27:44.242Z] 
[2024-11-06T22:27:44.242Z] Command:
[2024-11-06T22:27:44.242Z] kubectl get node -o yaml  --kubeconfig=/tmp/mdrke2131cilvb_kubeconfig  | grep mirrored-cilium  -A1
[2024-11-06T22:27:44.242Z] ---------------------
[2024-11-06T22:27:44.242Z] Assertion:
[2024-11-06T22:27:44.242Z] v1.16.3
[2024-11-06T22:27:44.242Z] ----------------------
[2024-11-06T22:27:44.242Z] Matched with result:
[2024-11-06T22:27:44.242Z] - docker.io/rancher/mirrored-cilium-cilium@sha256:62d2a09bbef840a46099ac4c69421c90f84f28d018d479749049011329aa7f28
[2024-11-06T22:27:44.242Z]       - docker.io/rancher/mirrored-cilium-cilium:v1.16.3
[2024-11-06T22:27:44.242Z]       sizeBytes: 223180987
[2024-11-06T22:27:44.242Z] --
[2024-11-06T22:27:44.242Z]       - docker.io/rancher/mirrored-cilium-operator-generic@sha256:6e2925ef47a1c76e183c48f95d4ce0d34a1e5e848252f910476c3e11ce1ec94b
[2024-11-06T22:27:44.242Z]       - docker.io/rancher/mirrored-cilium-operator-generic:v1.16.3
[2024-11-06T22:27:44.242Z]       sizeBytes: 31096970
[2024-11-06T22:27:44.242Z] 
[2024-11-06T22:27:44.243Z] Command:
[2024-11-06T22:27:44.243Z] kubectl get node -o yaml  --kubeconfig=/tmp/mdrke2131cilvb_kubeconfig  | grep hardened-cni-plugins -A1
[2024-11-06T22:27:44.243Z] ---------------------
[...]

PR #7091, opened 2024-10-22
Bump multus, cilium and flannel version

manuelbuil:
#### Proposed Changes ####

Bump multus, cilium and flannel charts so that we use the latest version of cni-plugins.

Flannel version includes windows and linux

#### Types of Changes ####

Version bump

#### Verification ####

Deploy those CNIs and verify that things are still working:
* Cilium
* Multus (with whatever other CNI plugin)
* Flannel in both windows and linux

#### Linked Issues ####

https://github.com/rancher/rke2/issues/7090

#### User-Facing Change ####

```release-note

```

PR #7194, opened 2024-11-06
[Release 1.31] Bump multus, cilium and flannel charts

manuelbuil:
Issue: https://github.com/rancher/rke2/issues/7193
Backport: https://github.com/rancher/rke2/pull/7091


.github/workflows/test-suite.yaml
brandond, line 1: Why is this whole file deleted? In the original PR it was just modified I believe?

manuelbuil: yeah, that was wrong. I had a conflict when backporting and did something stupid. It should be fixed now

brandond (approved):

PR #7417, opened 2024-12-11
Bump dns-node-cache to 1.24.0

dereknola:
#### Proposed Changes ####
Upstream has released a new dns-node-cache (after 6 months) that fixes about 10 CVEs. 

#### Types of Changes ####
Dependency Bump

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -11,7 +11,7 @@ charts:
   - version: v3.29.100
     filename: /charts/rke2-calico-crd.yaml
     bootstrap: true
-  - version: 1.36.101
+  - version: 1.36.102
     filename: /charts/rke2-coredns.yaml
     bootstrap: true
   - version: 4.10.503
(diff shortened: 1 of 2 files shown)


brandond (approved):

cwayne18:
Error: ASICS] + echo '[ERROR] Failed airgap image check'

Is that just a flake? seems weird it'd pass on arm64 but fail on amd64

cwayne18 (approved):

cwayne18:
I think we need to bump the coredns chart as well maybe?

cwayne18:
I'm too dumb to know exactly where to bump it but it looks like it does need a chart change:  https://github.com/rancher/rke2-charts/blob/b23470de7aa96ba4cb281fe1e6f50c3758d08441/packages/rke2-coredns/generated-changes/patch/values.yaml.patch#L174

Issue #7422, opened 2024-12-12
kube-controller-manager pod fails to be created with kube-controller-manager-extra-mount flag


skanakal:
**Environmental Info:**
RKE2 Version: v1.31.3+rke2r1

Node(s) CPU architecture, OS, and Version: x86_64, Ubuntu, 22.04

Cluster Configuration:

**Describe the bug:**

`kube-controller-manager` pod fails to be created with kube-controller-manager-extra-mount flag

**Steps To Reproduce:**

- Installed RKE2: 
~~~
#/etc/rancher/rke2/config.yaml
      kube-controller-manager-extra-mount:
        - /tmp/example.txt:/tmp/example.txt
        - /var/lib/rancher/rke2/server/logs/:/var/lib/rancher/rke2/server/logs
~~~

**Expected behavior:**

```
  - hostPath:
      path: /tmp/example.txt
      type: File
    name: extra-mount-0
  - hostPath:
      path: /var/lib/rancher/rke2/server/logs/
      type: DirectoryOrCreate
    name: extra-mount-1
```

**Actual behavior:**

~~~
tail -f /var/lib/rancher/rke2/agent/pod-manifests/kube-controller-manager.yaml 
    name: file9
  - hostPath:
      path: /tmp/example.txt
      type: File
    name: extra-mount-0
  - hostPath:
      path: /var/lib/rancher/rke2/server/logs/
      type: File
    name: extra-mount-1
~~~

**Additional context / logs:**

https://jira.suse.com/browse/SURE-9493

brandond:
How the mount is handled depends on whether or not something exists at that path, and if so, what kind of file it is. If you want to set the mount type manually, instead of having RKE2 try to guess it for you, use the four-parameter form, where the 3rd parameter is the mode (rw/ro) and the 4th is the mount type.
```yaml
 kube-controller-manager-extra-mount:
- /var/lib/rancher/rke2/server/logs/:/var/lib/rancher/rke2/server/logs:rw:DirectoryOrCreate
```

https://github.com/rancher/rke2/pull/5621

ShylajaDevadiga:
Closing as Issue is validated on all backports.

PR #7423, opened 2024-12-12
fix-extramount-type 

skanakal:
#### Proposed Changes ####
reset the `sourceType` variable for each extra mount entry, ensuring it does not retain the value from a previous mount specification.
initialize `sourceType` to `v1.HostPathUnset` at the start of processing each mount

#### Types of Changes ####
bugfix

#### Testing ####

Mentioned in linked issue

#### Linked Issues ####
https://github.com/rancher/rke2/issues/7422

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/staticpod/staticpod.go
+++ b/pkg/staticpod/staticpod.go
@@ -338,7 +338,7 @@ func addVolumes(p *v1.Pod, src []string, volume typeVolume) {
 func addExtraMounts(p *v1.Pod, extraMounts []string) {
 
 	for i, rawMount := range extraMounts {
-		var sourceType v1.HostPathType = v1.HostPathUnset // Reset sourceType for each mount
+		var sourceType v1.HostPathType
 		mount := strings.Split(rawMount, ":")
 		var ro bool
 		switch len(mount) {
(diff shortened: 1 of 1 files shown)


brandond (approved):

pkg/staticpod/staticpod.go
```diff
@@ -336,9 +336,9 @@ func addVolumes(p *v1.Pod, src []string, volume typeVolume) {
 }
 
 func addExtraMounts(p *v1.Pod, extraMounts []string) {
-	var sourceType v1.HostPathType
 
 	for i, rawMount := range extraMounts {
+		var sourceType v1.HostPathType = v1.HostPathUnset // Reset sourceType for each mount
```
brandond, line 341: Technically you don't have to initialize it with a specific value, `v1.HostPathType` is just a `string` and `v1.HostPathUnset` is an empty string. So just moving the initialization inside the loop would be fine.
```suggestion
		var sourceType v1.HostPathType
```

skanakal: Thank you for reviewing it. I have updated it.

brandond (approved):

vitorsavian (approved):

PR #7696, opened 2025-02-06
update report

fmoral2:
#### Proposed Changes ####
-     change report to only upload log file to s3
    the processing and parsing will be done in https://github.com/rancher/distros-test-framework/pull/215

 

#### User-Facing Change ####

```release-note

```


brandond (changes requested):
appears to need a `go mod tidy` due to new imports from the s3upload binary.

Also, does this really belong here? Adding this adds a dependency on aws-sdk-go that I do not believe we currently have. Could this go in https://github.com/rancher/ecm-distro-tools?

fmoral2:
> appears to need a `go mod tidy` due to new imports from the s3upload binary.
> 
> Also, does this really belong here? This appears to add a dependency on aws-sdk-go that I do not believe we currently have. Could this go in https://github.com/rancher/ecm-distro-tools?

- The dependency was already there
- since we are uploading a report from e2e tests, i cant see why go to tools. But if you prefer i can delete the directory and just have a file on the e2e/uploads3.go

matttrach (approved):

ShylajaDevadiga (approved):

tests/e2e/scripts/run_tests.sh
```diff
@@ -58,29 +58,29 @@ count=0
...
 do
-        cp createreport/rke2_${OS}.log createreport/rke2_${OS}_${count}.log
+        cp report/rke2_${OS}.log report/rke2_${OS}_${count}.log
         run_tests
 done
 
-# Generate report and upload to s3 bucket
-cd createreport && /usr/local/go/bin/go run -v report-template-bindata.go generate_report.go -f rke2_${OS}.log
+# Upload to s3 bucket
+cd report && /usr/local/go/bin/go run -v s3upload.go -f rke2_${OS}.log
```
mdrahman-suse, line 86: nit: EOF

PR #7820, opened 2025-02-26
Improves Updatecli bumping of CPI and CSI charts

macedogm:
#### Proposed Changes ####

Improves Updatecli bumping of CPI and CSI charts. The current method isn't working, see https://github.com/rancher/rke2/pull/7809, because the charts' versioning are fully following the [semantic versioning spec](https://semver.org/#spec-item-2):

> 2. A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and **MUST NOT contain leading zeroes**.

Due to that, Updatecli is correctly ignoring and removing the versions that have `.000`, `.001` etc.

-> https://github.com/updatecli/updatecli/blob/v0.94.1/pkg/plugins/resources/helm/source.go#L43
-> https://github.com/updatecli/updatecli/blob/v0.94.1/pkg/plugins/utils/version/filter.go#L165
-> https://github.com/updatecli/updatecli/blob/v0.94.1/pkg/plugins/utils/version/semver.go#L27-L29

Note: this isn't an issue with Updatecli.

Instead of trying to change upstream or all the versioned charts to follow proper semver, we propose a simple script that gets the latest chart from the index file (supposing that we are always adding the current latest version from top to bottom (i.e., at position 0)).

#### User-Facing Change ####

None.

```release-note

```

### DIFF
--- a/updatecli/scripts/retrieve_chart_version.sh
+++ b/updatecli/scripts/retrieve_chart_version.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -eu
+
+fatal() {
+    echo '[ERROR] ' "$@" >&2
+    exit 1
+}
+
+CHART_INDEX_FILE_URL="https://rke2-charts.rancher.io/index.yaml"
+CHART_NAME="${1}"
+# Retrieves the first entry '[0]', because we expect that the versions are
+# already ordered from last (more recent) to older.
+CHART_VERSION=$(curl -sfL "${CHART_INDEX_FILE_URL}" | yq -r '.entries.'"${CHART_NAME}"'[0].version')
+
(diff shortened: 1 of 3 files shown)


dereknola (approved):

brandond (approved):

brandond:
The leading zeroes come from how the charts build script embeds the PackageVersion. This has been fixed in newer versions of the library but upgrading would require a complete restructuring of the repo and CI config in rke2-charts so we have put it off.

macedogm:
Thanks for the approvals! JFYI that I cannot merge.

caroline-suse-rancher:
Thanks, @macedogm!

Issue #7840, opened 2025-03-06
Bump coredns to v1.39.100


thomasferrandiz:

aganesh-suse:
Closing based on: https://github.com/rancher/rke2/issues/7857

PR #7841, opened 2025-03-06
Bump coredns to v1.39.100

thomasferrandiz:
Issue: https://github.com/rancher/rke2/issues/7840

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -11,7 +11,7 @@ charts:
   - version: v3.29.101
     filename: /charts/rke2-calico-crd.yaml
     bootstrap: true
-  - version: 1.36.102
+  - version: 1.39.100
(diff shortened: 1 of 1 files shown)


mgfritch (approved):

dereknola (approved):

charts/chart_versions.yaml
```diff
@@ -11,7 +11,7 @@ charts:
   - version: v3.29.101
     filename: /charts/rke2-calico-crd.yaml
     bootstrap: true
-  - version: 1.36.102
+  - version: 1.39.100
```
brandond, line 14: No new images, only chart changes?

thomasferrandiz: yes, we had actually bumped the image before the upstream chart

PR #8029, opened 2025-04-09
Allow all IP protocols in rke2-flannel-host-networking policy.

prelz:
This patch takes into account all CIDRs in node.Spec.PodCIDR and node.Spec.PodCIDRs to compose the rke2-flannel-host-networking ingress policy. On dual-stack IPv6/IPv4 clusters this allows traffic from e.g. dual-stack MetalLB speakers to reach dual-stack Cluster IPs over both IPv4 and IPv6.

#### Proposed Changes ####

On dual-stack IPv4/IPv6 clusters (e.g. with cluster-cidr: "10.42.0.0/16,fd00:42::/56" and
service-cidr: "10.43.0.0/16,fd00:43::/112" in /etc/rancher/rke2/config.yaml), Cluster IPs on
nodes are allocated out of the node CIDRs, that have this form:
PodCIDR:                      10.42.1.0/24
PodCIDRs:                     10.42.1.0/24,fd00:42:0:1::/64

The rke2-flannel-host-networking policy, however, only allows the IPv4 source
address from all cluster nodes:
```
# kubectl describe netpol rke2-flannel-host-networking
Name:         rke2-flannel-host-networking
Namespace:    default
Created on:   2025-04-03 15:02:14 +0200 CEST
Labels:       <none>
Annotations:  <none>
Spec:
  PodSelector:     <none> (Allowing the specific traffic to all pods in this namespace)
  Allowing ingress traffic:
    To Port: <any> (traffic allowed to all ports)
    From:
      IPBlock:
        CIDR: 10.42.0.0/32
        Except: 
    From:
      IPBlock:
        CIDR: 10.42.1.0/32
        Except: 
( ... etc. ...)
```

The effect is that all IPv6 traffic received on an IPv6 External IP (assigned e.g. by a MetalLB loadbalancer
and received on the cluster node where the MetalLB speaker is running) cannot be routed
to the Cluster IP where the load-balanced service is listening.

This patch adds the IPv6 origin addresses as well, in parallel with the existing IPv4 policy and
with the same semantics. Not having a clear picture of the PodCIDR/PodCIDRs relation (is the
prefix in PodCIDR guaranteed to be always included in PodCIDRs ?) I tried to keep the patch
totally backwards compatible by checking node.Spec.PodCIDR first, then other prefixes from 
node.Spec.PodCIDRs.

This shouldn't require changes to the docs - that don't enter into much detail on the dual-stack IPv6/IPv4 scenario.
#### Types of Changes ####

If dual-stack IPv6/IPv4 clusters on bare metals are a supported feature this is a bug fix, otherwise it goes
one step further in supporting the IPv6/IPv4 dual-stack scenario.
#### Verification ####

On dual-stack IPv4/IPv6 clusters (e.g. with cluster-cidr: "10.42.0.0/16,fd00:42::/56" and
service-cidr: "10.43.0.0/16,fd00:43::/112" in /etc/rancher/rke2/config.yaml), check that the rke2-flannel-host-networking policy includes both IPv4 and IPv6 addresses. For example:
```
# kubectl describe netpol rke2-flannel-host-networking
Name:         rke2-flannel-host-networking
Namespace:    default
Created on:   2025-04-03 15:02:14 +0200 CEST
Labels:       <none>
Annotations:  <none>
Spec:
  PodSelector:     <none> (Allowing the specific traffic to all pods in this namespace)
  Allowing ingress traffic:
    To Port: <any> (traffic allowed to all ports)
    From:
      IPBlock:
        CIDR: 10.42.0.0/32
        Except: 
    From:
      IPBlock:
        CIDR: 10.42.1.0/32
        Except: 
    From:
      IPBlock:
        CIDR: 10.42.2.0/32
        Except: 
    From:
      IPBlock:
        CIDR: 10.42.3.0/32
        Except: 
    From:
      IPBlock:
        CIDR: 10.42.5.0/32
        Except: 
    From:
      IPBlock:
        CIDR: fd00:42:0:1::/128
        Except: 
    From:
      IPBlock:
        CIDR: fd00:42:0:2::/128
        Except: 
    From:
      IPBlock:
        CIDR: fd00:42:0:3::/128
        Except: 
    From:
      IPBlock:
        CIDR: fd00:42:0:5::/128
        Except: 
    From:
      IPBlock:
        CIDR: fd00:42::/128
        Except: 
  Not affecting egress traffic
  Policy Types: Ingress
```

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cisnetworkpolicy/controller.go
+++ b/pkg/controllers/cisnetworkpolicy/controller.go
@@ -134,16 +134,44 @@ func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIn
 			logrus.Debugf("CISNetworkPolicyController: node=%v doesn't have flannel label, skipping", node.Name)
 			continue
 		}
+		var addrlen string
+		// For full backwards compatibility check first 'PodCIDR'
+		// and next 'PodCIDRs'
 		podCIDRFirstIP, _, err := net.ParseCIDR(node.Spec.PodCIDR)
 		if err != nil {
 			logrus.Debugf("CISNetworkPolicyController: node=%+v", node)
 			logrus.Errorf("CISNetworkPolicyController: couldn't parse PodCIDR(%v) for node %v err=%v", node.Spec.PodCIDR, node.Name, err)
-			continue
+		} else {
+			addrlen = "/32"
+			if podCIDRFirstIP.To4() == nil {
+				addrlen = "/128"
+			}
+			ipBlock := netv1.IPBlock{
+				CIDR: podCIDRFirstIP.String() + addrlen,
+			}
+			npIR.From = append(npIR.From, netv1.NetworkPolicyPeer{IPBlock: &ipBlock})
+                }
+
+		for _, cidr := range node.Spec.PodCIDRs {
+			podCIDRsFirstIP, _, err := net.ParseCIDR(cidr)
+			if err != nil {
+				logrus.Debugf("CISNetworkPolicyController: node=%+v", node)
+				logrus.Errorf("CISNetworkPolicyController: couldn't parse PodCIDRs(%v) for node %v err=%v", cidr, node.Name, err)
+				continue
+			}
+			if podCIDRsFirstIP.String() == podCIDRFirstIP.String() {
+				// Same as 'podCIDR' above
+				continue
+			}
+			addrlen = "/32"
+			if podCIDRsFirstIP.To4() == nil {
+				addrlen = "/128"
+			}
+			ipBlock := netv1.IPBlock{
+				CIDR: podCIDRsFirstIP.String() + addrlen,
+			}
+			npIR.From = append(npIR.From, netv1.NetworkPolicyPeer{IPBlock: &ipBlock})
 		}
-		ipBlock := netv1.IPBlock{
-			CIDR: podCIDRFirstIP.String() + "/32",
-		}
-		npIR.From = append(npIR.From, netv1.NetworkPolicyPeer{IPBlock: &ipBlock})
 	}
 
 	// sort ipblocks so it always appears in a certain order


pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -134,16 +134,44 @@ func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIn
...
+			if podCIDRFirstIP.To4() == nil {
+				addrlen = "/128"
+			}
+			ipBlock := netv1.IPBlock{
+				CIDR: podCIDRFirstIP.String() + addrlen,
+			}
+			npIR.From = append(npIR.From, netv1.NetworkPolicyPeer{IPBlock: &ipBlock})
+                }
+
+		for _, cidr := range node.Spec.PodCIDRs {
```
brandond, line 155: If PodCIDRs is guaranteed to exist and contain the legacy PodCIDR, why not just iterate over PodCIDRs and skip the special case for the legacy field?

Alternately, why not do something like
```golang
podCIDRS := node.Spec.PodCIDRs
if len(podCIDRs) == 0 {
  podCIDRs = []string{node.Spec.PodCIDR}
}
for _, cidr := range podCIDRs {
```

This would significantly cut down on the duplicate logic.

brandond (changes requested):

pkg/controllers/cisnetworkpolicy/controller.go
prelz: OK - thanks. I am not clear on the PodCIDRs/PodCIDR relationship, and could not navigate further up into the code to understand it better. I had left the patch as a pure addition to the original logic to make my purpose clearer. I committed the proposed change into the branch and am now testing it in our production environment.

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -134,16 +134,29 @@ func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIn
...
+		if len(podCIDRs) == 0 {
+			podCIDRs = []string{node.Spec.PodCIDR}
 		}
-		ipBlock := netv1.IPBlock{
-			CIDR: podCIDRFirstIP.String() + "/32",
+
+		for _, cidr := range podCIDRs {
+			podCIDRFirstIP, _, err := net.ParseCIDR(cidr)
+			if err != nil {
+				logrus.Debugf("CISNetworkPolicyController: node=%+v", node)
```
brandond, line 147: This should go outside the `range podCIDRs` loop

brandond (changes requested):
one nit, lgtm otherwise

pkg/controllers/cisnetworkpolicy/controller.go
prelz: The only way I can figure out to avoid the duplicate message (will there ever be more than two CIDRs?) without the side effect of swapping the order of log messages  (and as someone who started with punched cards and would have hoped to retire before having to code in Go...) is the one I just committed to the branch.

pkg/controllers/cisnetworkpolicy/controller.go
```diff
@@ -134,16 +134,33 @@ func (h *handler) generateHostNetworkPolicyIngressRule() (*netv1.NetworkPolicyIn
...
-			CIDR: podCIDRFirstIP.String() + "/32",
+
+		debug_logged := false
+		for _, cidr := range podCIDRs {
+			podCIDRFirstIP, _, err := net.ParseCIDR(cidr)
+			if err != nil {
+				if !debug_logged {
+					logrus.Debugf("CISNetworkPolicyController: node=%+v", node)
+					debug_logged = true
+				}
```
brandond, line 151: this is too much... just move it before the range block.

brandond (changes requested):

pkg/controllers/cisnetworkpolicy/controller.go
prelz: This debug message was not/never logged when no CIDR parse error occurs. This changes the
behaviour beyond what is needed for the patch at hand. Are you sure about that >

brandond: I guess you could just remove it entirely? We don't really need the whole node object dumped out if the CIDR is invalid; I think maybe this was added during development and never removed.

prelz: Ok - debug log line removed.

brandond (approved):

manuelbuil (approved):

rbrtbnfgl (approved):

PR #8130, opened 2025-05-02
Fix the vSphere chart version check to use version sort for the retrieved versions

macedogm:
In https://github.com/rancher/rke2/pull/7820 I wrongly assumed that the charts' versions were sorted from newest to oldest automatically in https://rke2-charts.rancher.io/index.yaml. That isn't the case, so we must sort them by natural version sort instead.

Fixes the wrong update PR https://github.com/rancher/rke2/pull/7809.


macedogm:
FYI @cwayne18

brandond (approved):

dereknola (approved):

Issue #8146, opened 2025-05-08
Windows: impossible to orchestrate the uninstall.ps1 script to not ask for confirmation


Najzer0:
**Environmental Info:**
RKE2 Version: 
1.31.7 and 1.31.8 - but should be the same for most recent ones as uninstall.ps1 did not receive changes for quite some time.

Cluster Configuration:
1 master
1 lin + 1 win worker

**Describe the bug:**
Not possible to orchestrate uninstall.ps1 to run silent/unattended.
Tried: $ConfirmationPreference = None/Low
Pipeing via `echo Y Y Y[...] | powershell.exe [...]uninstall.ps1`
adding -Confirm:$false to the script

**Steps To Reproduce:**
Call uninstall.ps1 script and see that you'll have to confirm quite alot stop-process/stop-service an new-item prompts.

**Expected behavior:**
Possible to run this without user interaction

**Actual behavior:**
requires multible confirmations

mdrahman-suse:
Validated on the release branch with commits

PR #8147, opened 2025-05-08
Windows: Allow for silent/non confirmation use of uninstall.ps1

Najzer0:
#### Proposed Changes ####

We encountered that it is impossible to orchestrate our cmdlines (pipe, settings) to auto-approve the stop-process, stop-service and create file cmds within the uninstall.ps1 for windows nodes.
Thus I tried to improve the script to offer an optional silent param to the script.
If none is given, it still acts & behaves like before.

#### Types of Changes ####

New feature: Confirm:$falset flag added
supresses confirmation within script while calling stop-process, stop-service and new-item

#### Verification ####

Using the uninstall script without silent flag still asks for confirmation.
No means from outside are able to overwrite this for us:
examples given: 
$ConfirmationPreference = none/low
pipeing "Y" via `echo Y Y Y [...] | uninstall.ps1`
adding -Confirm:$false

#### Testing ####

uninstall a windows node with `Confirm:$false` and it should not ask for confirmation. If you uninstall without passing any flag, it should work as today and ask for confirmation

#### Linked Issues ####
https://github.com/rancher/rke2/issues/8146

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####

Hope I am doing this right ;-)

### DIFF
--- a/bundle/bin/rke2-uninstall.ps1
+++ b/bundle/bin/rke2-uninstall.ps1
@@ -11,6 +11,8 @@
 .EXAMPLE 
     rke2-uninstall.ps1
     Uninstalls the RKE2 Windows service and cleans the RKE2 Windows Agent (Worker) Node
+    rke2-uninstall.ps1 -silent $true
+    Uninstalls the RKE2 Windows service and cleans the RKE2 Windows Agent (Worker) Node with suppressed confirmations
 #>
 
 Param(
--- a/bundle/bin/rke2-uninstall.ps1
+++ b/bundle/bin/rke2-uninstall.ps1
@@ -11,13 +11,13 @@
 .EXAMPLE 
     rke2-uninstall.ps1
     Uninstalls the RKE2 Windows service and cleans the RKE2 Windows Agent (Worker) Node
-    rke2-uninstall.ps1 -silent $true
-    Uninstalls the RKE2 Windows service and cleans the RKE2 Windows Agent (Worker) Node with suppressed confirmations
+    rke2-uninstall.ps1 -Confirm $false
+    Uninstalls the RKE2 Windows service and cleans the RKE2 Windows Agent (Worker) Node without asking for confirmations
 #>
 
 Param(
     [Parameter(Mandatory=$False)]
-    [bool]$silent = $False
+    [bool]$Confirm = $True
 )
 
 $ErrorActionPreference = 'Stop'
@@ -114,7 +114,7 @@ function Stop-Processes () {
         Write-LogInfo "Checking if $ProcessName process exists"
         if ((Get-Process -Name $ProcessName -ErrorAction SilentlyContinue)) {
             Write-LogInfo "$ProcessName process found, stopping now"
-            if ($silent) {
+            if ($Confirm -eq $false) {
                 Stop-Process -Name $ProcessName -Force -Confirm:$false
             } else {
                 Stop-Process -Name $ProcessName
@@ -140,7 +140,7 @@ function Invoke-CleanServices () {
         Write-LogInfo "Checking if $ServiceName service exists"
         if ((Get-Service -Name $ServiceName -ErrorAction SilentlyContinue)) {
             Write-LogInfo "$ServiceName service found, stopping now"
-            if ($silent) {
+            if ($Confirm -eq $false) {
                Stop-Service -Name $ServiceName -Force -Confirm:$false
             } else {
                Stop-Service -Name $ServiceName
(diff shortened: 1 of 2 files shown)


siprbaum:
Is there a reason why `Force` isn't the default? E.g. if the user said to uninstall, why ask again for stop-process and stop-service?

manuelbuil:
Two things:
1 - Args in powershell normally start with a capital letter
2 - I'd say in powershell the name "--Force" makes more sense. When I see "--Silent", I'd expect it to not provide any user interaction, but also in terms of logging, etc, which is not the case here. If you don't agree, that's fine, I don't think it is a hard requirement and "--Silent" is not that bad. I guess it's just my taste

Najzer0:
> Two things: 1 - Args in powershell normally start with a capital letter 2 - I'd say in powershell the name "--Force" makes more sense. When I see "--Silent", I'd expect it to not provide any user interaction, but also in terms of logging, etc, which is not the case here. If you don't agree, that's fine, I don't think it is a hard requirement and "--Silent" is not that bad. I guess it's just my taste

Agreed on the capital letter.
For the "silent" param name: I wanted to make clear that its "doing more" than the usual -Force (timing issue if no delay between answers) and in my corpo language we have "silent install" (no window, happening without user interaction)... what it actually does is "SupressConfirmation" of some sub-powershell commands. I do not have a preference in this regard and will gladly follow your "Force" or any other recommendation :-)

manuelbuil:
> > Two things: 1 - Args in powershell normally start with a capital letter 2 - I'd say in powershell the name "--Force" makes more sense. When I see "--Silent", I'd expect it to not provide any user interaction, but also in terms of logging, etc, which is not the case here. If you don't agree, that's fine, I don't think it is a hard requirement and "--Silent" is not that bad. I guess it's just my taste
> 
> Agreed on the capital letter. For the "silent" param name: I wanted to make clear that its "doing more" than the usual -Force (timing issue if no delay between answers) and in my corpo language we have "silent install" (no window, happening without user interaction)... what it actually does is "SupressConfirmation" of some sub-powershell commands. I do not have a preference in this regard and will gladly follow your "Force" or any other recommendation :-)

What about using the flag "-Confirm:$false" as well for the script, instead of -Silent? In the end that's what you are adding for the different cmdlets and it is easy to understand for a powershell user

Najzer0:
> What about using the flag "-Confirm:$false" as well for the script, instead of -Silent? In the end that's what you are adding for the different cmdlets and it is easy to understand for a powershell user

I wanted to keep the initial script behaviour the way it is and its just personal preference to have optional flags "true-ish". I can actually live with the idea/rework in favor of having Confirm:$true as the second (optional) param. I just wanted to give us a way to supress it and could not inject it any other way but fixing it with ifs (eg. ContinueAfterError scope).

From tommorow I'll be on 2 weeks vacation, but I intend to rework/write this in favor of Confirm:$boolean.

Thanks for your feedback, lets see how the corpo gods give me time tomorrow :-)

manuelbuil:
> > What about using the flag "-Confirm:$false" as well for the script, instead of -Silent? In the end that's what you are adding for the different cmdlets and it is easy to understand for a powershell user
> 
> I wanted to keep the initial script behaviour the way it is and its just personal preference to have optional flags "true-ish". I can actually live with the idea/rework in favor of having Confirm:$true as the second (optional) param. I just wanted to give us a way to supress it and could not inject it any other way but fixing it with ifs (eg. ContinueAfterError scope).
> 
> From tommorow I'll be on 2 weeks vacation, but I intend to rework/write this in favor of Confirm:$boolean.
> 
> Thanks for your feedback, lets see how the corpo gods give me time tomorrow :-)

Could we have the behaviour as `Confirm:$true` as the default if no flag is passed, so that it behaves like today. And only do differently if we add `Confirm:$false` (following your silent approach)? Thanks a lot for your time [Najzer0](https://github.com/Najzer0)

Najzer0:
> Could we have the behaviour as `Confirm:$true` as the default if no flag is passed, so that it behaves like today. And only do differently if we add `Confirm:$false` (following your silent approach)? Thanks a lot for your time [Najzer0](https://github.com/Najzer0)

reworked in this way. Thanks for your feedback & incorporated now.

dereknola (approved):

manuelbuil:
Great! Works perfectly! Before merging, could you re-write the description of the PR so that it reflects the new behavior please? Just to avoid confusion in the future :). Thanks

manuelbuil:
I just noticed the author is on PTO. I decided to update the description

PR #8287, opened 2025-05-18
Prevent accidental remove across filesystems

richardmoe:
#### Proposed Changes ####
Prevent accidental deletion of files on remote filesystems in case there are any lingering pods with remote  mounts when cleaning a node.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/8296

#### User-Facing Change ####

```release-note
NONE
```


brandond:
Can you open an issue briefly describing the issue you're resolving here, so that our QA team can reproduce it?

richardmoe:
We had a case where a node was removed from the cluster to be reinstalled. The node had been used to run jobs where the pods mount smb shares. We have experienced that smb shares have continued to stay mounted under `/var/lib/kubelet` after a job has completed. When rke2-uninstall.sh runs `rm -rf  /var/lib/kubelet` it will traverse into remote fileshares. 
The script could be improved to display a warning of that there are remote filesystems mounted and ask the user for confirmation before running the script.

brandond (approved):

dereknola (approved):

PR #8289, opened 2025-05-19
Respect k3s experimental agentless server option

glennpratt:
#### Proposed Changes ####

[The experimental k3s flag `--disable-agent`](https://docs.k3s.io/advanced#running-agentless-servers-experimental) is passed through already, however, items that contact containerd directly in `rke2 server` do not respect this flag.

Supporting this flag allows running rke2 server independently of local static pods. This might be useful for:

* Customizing the apiserver execution environment
* A vcluster distro like [k3s](https://github.com/loft-sh/vcluster/blob/main/pkg/k3s/k3s.go#L38)
* Running `rke2 server` as a pod for quick testing
    * this could be useful for testing [agent node bootstrap](https://github.com/rancher/cluster-api-provider-rke2) with a local kind Cluster API for example.

#### Types of Changes ####

Fix for hidden, unsupported feature.

#### Verification ####

I verified this change by applying the patch onto `v1.32.4-rke2r1` and running that build as a StatefulSet. Etcd and apiserver were not rebuilt, used the published images.

https://gist.github.com/glennpratt/6272c94db3093127a948a37c5a378a0e

```
Containers(rke2-bootstrap-system/rke2-server-0)[4]

NAME              READY     STATE         IMAGE                                                                        
rke2-server-init  true      Completed     docker.local/rke2:tilt-7412b621e3de84bd                                      
rke2-server       true      Running       docker.local/rke2:tilt-7412b621e3de84bd                                      
kube-apiserver    true      Running       index.docker.io/rancher/hardened-kubernetes:v1.32.4-rke2r1-build20250423     
etcd              true      Running       index.docker.io/rancher/hardened-etcd:v3.5.21-k3s1-build20250411  
```

#### Testing ####

This change is not currently covered by unit tests. Happy to add any tests desired, but I may need some assistance if it's beyond unit tests.

#### User-Facing Change ####

This is a hidden feature, undocumented in rke2 but documented in k3s.

```release-note
Improved support for `--disable-agent` (experimental, hidden server flag)
```

### DIFF
--- a/pkg/rke2/rke2.go
+++ b/pkg/rke2/rke2.go
@@ -88,6 +88,10 @@ const (
 )
 
 func Server(clx *cli.Context, cfg Config) error {
+	disableAgent := clx.Bool("disable-agent")
+	if disableAgent {
+		cmds.ServerConfig.DisableAgent = true
+	}
 	if err := setup(clx, cfg, true); err != nil {
 		return err
 	}
@@ -114,13 +118,26 @@ func Server(clx *cli.Context, cfg Config) error {
 		metav1.NamespaceDefault,
 		metav1.NamespacePublic,
 	}
+
 	dataDir := clx.String("data-dir")
+
+	if !disableAgent {
+		cmds.ServerConfig.StartupHooks = append(cmds.ServerConfig.StartupHooks,
+			reconcileStaticPods(cmds.AgentConfig.ContainerRuntimeEndpoint, dataDir),
+		)
+	}
 	cmds.ServerConfig.StartupHooks = append(cmds.ServerConfig.StartupHooks,
 		setNetworkPolicies(cisMode, defaultNamespaces),
 		setClusterRoles(),
 		restrictServiceAccounts(cisMode, defaultNamespaces),
 		setKubeProxyDisabled(),
-		cleanupStaticPodsOnSelfDelete(dataDir),
+	)
+	if !disableAgent {
+		cmds.ServerConfig.StartupHooks = append(cmds.ServerConfig.StartupHooks,
+			cleanupStaticPodsOnSelfDelete(dataDir),
+		)
+	}
+	cmds.ServerConfig.StartupHooks = append(cmds.ServerConfig.StartupHooks,
 		setRuntimes(),
 	)
 
@@ -190,7 +207,11 @@ func setup(clx *cli.Context, cfg Config, isServer bool) error {
 		}
 	}
 
-	return removeDisabledPods(dataDir, containerRuntimeEndpoint, disabledItems, clusterReset)
+	if !clx.Bool("disable-agent") {
+		return removeDisabledPods(dataDir, containerRuntimeEndpoint, disabledItems, clusterReset)
+	}
+
+	return nil
 }
 
 func ForceRestartFile(dataDir string) string {


brandond:
`--disable-agent` is not supported for RKE2 and will never work. RKE2 is built around deploying the datastore and control-plane components as static pods. Static pods do not function without a kubelet and container runtime, which is exactly what is excluded from startup when you use the `--disable-agent` flag.

brandond:
I guess I'm not even sure what you're trying to accomplish here. Do you want to start up just the RKE2 supervisor API (the bit that runs on 9345) without any of the control plane components running?

brandond:
Is this perhaps closer to what you had in mind?
https://github.com/rancher/rke2/compare/master...brandond:rke2:disable-agent

glennpratt:
Thanks for the quick response @brandond. Yes, having only the Supervisor API was the goal.

I think I want it for similar reasons I would in k3s and similar to vcluster. To have a Node-less control plane that uses an external etcd (e.g. kine), has no root privileges on the machine it's running on and starts quickly. Normal full VM RKE2 Agent nodes would join to this.

Granted, with this simple change, I have to orchestrate the apiserver myself. In my example, the flags could still be consumed from generated manifests, but the host bind mounts no longer make any sense.

[The example Gist](https://gist.github.com/glennpratt/6272c94db3093127a948a37c5a378a0e) has ephemeral etcd just to be a self-contained example. I will attach a full agent node to it and see how that goes. It at least runs without errors and responds to kubectl requests with the linked gist example. In testing with Tilt, it's also just nice how fast I can iterate with it.

brandond:
> To have a Node-less control plane that uses an external etcd (e.g. kine), has no root privileges on the machine it's running on and starts quickly.
> Granted, with this simple change, I have to orchestrate the apiserver myself.

If you want a functional control-plane with RKE2 you'll need the kubelet and containerd. Is there any particular reason you're not just using K3s?

brandond:
> What I want is part of an internal project to unify K8s control plane hosting internally and externally at my company. 

What exactly do you mean by this? What does this look like, in practice?

> Ideally, most parts of the RKE2 control plane (e.g. not DaemonSets) could run as unprivileged pods in an orchestration cluster that are invisible to the resulting user cluster. 

There are no daemonsets. The RKE2 control-plane is entirely comprised of static pods, using manifests created by the RKE2 supervisor process and executed by the kubelet.

While I would love to enable some variety of an agentless server, even in an unsupported capacity - we're unlikely to move forward with any approach that doesn't result in a working cluster. So, an agentless server using kine or external etcd, where RKE2 continues to manage the control-plane pods through containerd + kubelet, is something that we could enable. An agentless server where the supervisor comes up and does nothing until someone externally provides the control-plane components using some external automation, is not something we'd be interested in.

glennpratt:
> What does this look like, in practice?

K8s control planes run inside orchestration Kubernetes clusters as Pods. These orchestration clusters are managed by teams focused on k8s, the workloads are managed by people managing apps or creating products.

> There are no DaemonSets

I know, it was just an example, there are DaemonSets in a typical RKE2 control plane Node. These aren't part of our control planes, which I suppose is obvious.

> heterogenous clusters

This is not what we're doing - clusters are homogenous, it's just the control plane is Node-less, is provided a datastore and run as Pods with as minimal privileges, not VMs.

> custom executor that integrated with the host cluster's Kubernetes API

This is interesting and I could work towards this, but there are complexities like mounts that may be awkward to wire up unless it's pluggable. We treat these control planes as typical Kubernetes applications, they have no host mounts and things like certs and tokens are generated in an operator provided by mounting from Secrets where possible. Rotations are performed by rollout (Pod replacement).

It's also a little weird for scheduling, I'm not sure you can elegantly express that the rke2 Pod is going to create an apiserver Pod next to it - two containers fit more naturally there, but you would generate that in advance. I'd be happy to even use `rke2` in the manifest generation phase, but that would be in something like and operator or a Job.

glennpratt:
That does sound similar to what we're doing. Ours is implemented as a private, Pod-based [cluster-api control plane provider](https://cluster-api.sigs.k8s.io/user/concepts#control-plane). (I'm not sure Pod-based is anything more than an idea from cluster-api's perspective.) 

At a glance, something that is part of the "workload" cluster's steady operation or listening to remote traffic, e.g. the Supervisor API, I'd want isolated from an operator. I'd want it to have little or no access to the management apiserver and separate concerns about eviction / scaling.

brandond:
I've been poking at this a bit more and it's... more complicated than I initially anticipated. It is easy enough to get the supervisor to run the control plane components as deployments in a host cluster, I have that working. Other than the previously mentioned difficulties around managing the etcd cluster pods, the supervisor API is proving somewhat difficult to break out from the apiserver.

The control-plane pods normally run with host network, the components can always find eachother on the loopback address, and clients can rely on every apiserver IP also hosting the supervisor API on a different port. This will need to either be broken out on the k3s side, or the apiserver pods will need to proxy the supervisor port back to the supervisor pods.

There are also complications around the agent tunnels that the apiserver uses to connect up to the kubelet for `kubectl exec` and `kubectl logs`. Normally the agents directly connect to the supervisor port on all apiserver nodes so that the apiservers can dial them through the tunnel, but this won't work if the pods are all hidden behind a single endpoint. Only the single pod that the agent is connected to will be able to connect to the agent, unless we do some extra work to peer the remotedialer instances between pods. This will also require changes on the k3s side.

glennpratt:
I think I my own trials that's addressed by a couple things.

Things that expect to talk on loopback go in the same Pod.
Each member of an STS gets a host ports or its own IP visible to agents. So they can be addressed directly.

brandond:
If things are going to be properly abstracted, they need to work behind a LoadBalancer or Ingress. I wouldn't personally settle for anything that required agents to be able to connect directly to one or more pods.

PR #8439, opened 2025-06-18
Add ADR for multus thick support

thomasferrandiz:


docs/adrs/009-multus-thick-plugin.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Pros
+- allow the use of [multus-dynamic-networks-controller](https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller) by Harvester
+- Multus thick plugin provides better metrics
+
+### Cons
+- higher resource consumption than with thin plugin
+- more complex multus chart to support
+    - new optional components
+    - a different configuration is needed for each CNI we support (https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/thick-plugin.md#configure-deployment)
```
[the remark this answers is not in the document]
thomasferrandiz: With the multus thick plugin, the delegate CNI is called from inside the multus-daemon pod so the pod needs access to the part of the filesystem used by the specific CNI used in the cluster (e.g. `/run/flannel/subnet.env`) for flannel.

docs/adrs/009-multus-thick-plugin.md
```diff
@@ -0,0 +1,41 @@
...
+
+In addition, the thick plugin deployment should become the default method from upstream in the future.
+
+### Pros
+- allows the use of [multus-dynamic-networks-controller](https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller) by Harvester to implement [interface hotplugging](https://kubevirt.io/user-guide/network/hotplug_interfaces/) with Kubevirt
+- Multus thick plugin provides better metrics
+
+### Cons
+- higher resource consumption than with thin plugin
+- increased complexity to support the multus CRI shim
```
manuelbuil, line 23: Do you have an overview on what needs to happen to integrate it with containerd? If the containerd configuration needs to change, I think we need to restart containerd to consume the updated config. IIRC, that will restart rke2 process

docs/adrs/009-multus-thick-plugin.md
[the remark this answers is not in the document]
brandond: Could we just mount /run and /opt from the host into the pod? Mounting specific subdirectories per CNI feels like unnecessary complexity.

docs/adrs/009-multus-thick-plugin.md
brandond: It does not look to me like the containerd config needs to change. The thick plugin implements the CRI itself, and then passes through to containerd. So we will need to start up both containerd and the thick plugin, point kubelet/crictl at that as the container runtime endpoint, and then the thick plugin talks to containerd's CRI endpoint.

brandond: The thick plugin does not do anything with CRI no, but the dynamic network controller thank kube-virt needs does.

https://kubevirt.io/user-guide/network/hotplug_interfaces/#requirements
>  https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller
> Multus Dynamic Networks Controller: this daemon will listen to annotation changes, and trigger Multus to configure a new attachment for this pod.

thomasferrandiz: I just did a test of the dynamic networks controller + thick plugin setup.
The dnc pod needs to have the CRI socket mounted but there is no other setup to do.

Then adding and deleting interfaces from a pod following the README works by adding annotations to the pod manifest which are watched by the dynamic network controller.

brandond: Ok so it sounds like both can run in pods deployed by the chart then?

manuelbuil (approved):

brandond (approved):

PR #8492, opened 2025-07-07
Defaults to 'bare-metal' in case the connection to EC2 or GCE times out

ga4rosso:
When trying to determine the platform type a time out error was returned in case EC2 or GCE were not able to be contacted.
With this change the platform type "bare-metal" is now returned as a fallback value.

### Linked issues:
* This fix addresses https://github.com/rancher/rke2/issues/8497

### DIFF
--- a/pkg/windows/utils.go
+++ b/pkg/windows/utils.go
@@ -166,7 +166,7 @@ func platformType() (string, error) {
 	hostnameEc2Uri := "http://169.254.169.254/latest/meta-data/local-hostname"
 	ec2Resp, err := http.Get(hostnameEc2Uri)
 	if err != nil && hasTimedOut(err) {
-		logrus.Debugf("Timed out trying to get information from '%s' (EC2). Defaulting to 'bare-metal'", hostnameEc2Uri)
+		logrus.Warnf("Timed out trying to get EC2 instance metadata from %q; defaulting to 'bare-metal' Calico platform", hostnameEc2Uri)
 		return "bare-metal", nil
 	}
 	if ec2Resp != nil {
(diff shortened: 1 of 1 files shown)


pkg/windows/utils.go
```diff
@@ -163,9 +163,11 @@ func platformType() (string, error) {
 	}
 
 	// EC2
-	ec2Resp, err := http.Get("http://169.254.169.254/latest/meta-data/local-hostname")
+	hostnameEc2Uri := "http://169.254.169.254/latest/meta-data/local-hostname"
+	ec2Resp, err := http.Get(hostnameEc2Uri)
 	if err != nil && hasTimedOut(err) {
-		return "", err
+		logrus.Debugf("Timed out trying to get information from '%s' (EC2). Defaulting to 'bare-metal'", hostnameEc2Uri)
```
brandond, line 169:
```suggestion
		logrus.Warnf("Timed out trying to get EC2 instance metadata from %q; defaulting to 'bare-metal' Calico platform", hostnameEc2Uri)
```

pkg/windows/utils.go
```diff
@@ -176,14 +178,16 @@ func platformType() (string, error) {
...
+	hostnameGoogleUri := "http://metadata.google.internal/computeMetadata/v1/instance/hostname"
+	req, err := http.NewRequest("GET", hostnameGoogleUri, nil)
 	if err != nil {
 		return "", err
 	}
 	req.Header.Add("Metadata-Flavor", "Google")
 	gceResp, err := client.Do(req)
 	if err != nil && hasTimedOut(err) {
-		return "", err
+		logrus.Debugf("Timed out trying to get information from '%s' (GCE). Defaulting to 'bare-metal'", hostnameGoogleUri)
```
brandond, line 189:
```suggestion
		logrus.Warnf("Timed out trying to get GCE instance metadata from %q; defaulting to 'bare-metal' Calico platform", hostnameGoogleUri)
```

brandond (changes requested):
thanks, lgtm, but a couple nits on logging!

brandond (approved):

pkg/windows/utils.go
ga4rosso: thanks for reviewing it. Adapted as suggested.

pkg/windows/utils.go
ga4rosso: thanks for reviewing it. Adapted as suggested.

dereknola (approved):

PR #8759, opened 2025-08-12
Remove /var/log/calico in killall script

manuelbuil:
#### Proposed Changes ####

When deploying with canal, the calico container creates the directory /var/log/calico in the node. This directory does not get removed when running killall 

#### Types of Changes ####

Bugfix

#### Verification ####

Install rke2 with cni: canal, run rke2-killall.sh and verify the `/var/log/calico/` directory gets removed

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/bundle/bin/rke2-killall.sh
+++ b/bundle/bin/rke2-killall.sh
@@ -92,7 +92,7 @@ if [ -d /sys/class/net/nodelocaldns ]; then
   ip link delete nodelocaldns
 fi
 
-rm -rf /var/lib/cni/ /var/log/pods/ /var/log/containers
+rm -rf /var/lib/cni/ /var/log/pods/ /var/log/containers /var/log/calico
 
 # Remove pod-manifests files for rke2 components
 POD_MANIFESTS_DIR=${RKE2_DATA_DIR}/agent/pod-manifests


brandond:
I am confused why the killall script removes ANY logs to begin with. I guess I added that in https://github.com/rancher/rke2/pull/3821 but the k3s killall script does not do this and I don't think we have issues with the pod restart count not resetting over there. Or maybe we do and just nobody has complained about it - because it is true, the pod gets restarted NOT recreated after a killall. The change was mostly to reduce confusion for QA when reusing nodes.

manuelbuil:
> I am confused why the killall script removes ANY logs to begin with. I guess I added that in #3821 but the k3s killall script does not do this and I don't think we have issues with the pod restart count not resetting over there. Or maybe we do and just nobody has complained about it - because it is true, the pod gets restarted NOT recreated after a killall. The change was mostly to reduce confusion for QA when reusing nodes.

I thought the reason is that when you run "killall", you want to start fresh again, so removing the logs might be a good idea to avoid getting confused with old logs. TBH, I initially thought of adding the removal line in the uninstall script but then I saw the other log being removed in killall, that's why I added it there. I'll move it wherever you think it makes more sense, I don't mind

brandond:
Honestly I think that log removal should be moved to the uninstall script, but we will need to work with QA (cc @ShylajaDevadiga) to address any specific behavior they may have been relying on with regards to pod restart counts. I think that counting a pod as having been restarted after a killall is technically correct, but QA was using `rke2-killall.sh; rm -rf /var/lib/rancher/rke2` to reset nodes between tests and was surprised by seeing the pod restart count as non-zero after this.

ShylajaDevadiga:
> Honestly I think that log removal should be moved to the uninstall script, but we will need to work with QA (cc @ShylajaDevadiga) to address any specific behavior they may have been relying on with regards to pod restart counts. I think that counting a pod as having been restarted after a killall is technically correct, but QA was using `rke2-killall.sh; rm -rf /var/lib/rancher/rke2` to reset nodes between tests and was surprised by seeing the pod restart count as non-zero after this.

I agree. It's a good idea for rke2-uninstall.sh to handle complete log removal. Currently we're seeing /var/log/calico directory still remains after running rke2-uninstall.sh. It might be best for rke2-killall.sh to kill processes and process related directories while rke2-uninstall.sh to handle full cleanup so that the node can be re-usable if needed.

manuelbuil:
> > Honestly I think that log removal should be moved to the uninstall script, but we will need to work with QA (cc @ShylajaDevadiga) to address any specific behavior they may have been relying on with regards to pod restart counts. I think that counting a pod as having been restarted after a killall is technically correct, but QA was using `rke2-killall.sh; rm -rf /var/lib/rancher/rke2` to reset nodes between tests and was surprised by seeing the pod restart count as non-zero after this.
> 
> I agree. It's a good idea for rke2-uninstall.sh to handle complete log removal. Currently we're seeing /var/log/calico directory still remains after running rke2-uninstall.sh. It might be best for rke2-killall.sh to kill processes and process related directories while rke2-uninstall.sh to handle full cleanup so that the node can be re-usable if needed.

All right, I'll try to do it next week :)

bundle/bin/rke2-uninstall.sh
```diff
@@ -123,6 +123,7 @@ uninstall_remove_files()
     rm --one-file-system -rf /var/lib/kubelet || true
     rm -rf "${RKE2_DATA_DIR}" || error "Failed to remove /var/lib/rancher/rke2"
     rm -d /var/lib/rancher || true
+    rm -rf /var/lib/cni/ /var/log/pods/ /var/log/containers /var/log/calico
```
brandond, line 126: Can you put this up with /var/lib/kubelet?

brandond (changes requested):

brandond:
lgtm once rebased

PR #8766, opened 2025-08-13
rke2-uninstall.sh: add -- to rm command for safer user-supplied path and remove hardcoded path

philippebi:
#### Proposed Changes ####

Add -- to the rm command to make it safer. The RKE2_DATA_DIR variable could be user-defined (only a default is defined) and, if mistakenly set to something like '-rf /', could lead to unintended consequences.

### DIFF
--- a/bundle/bin/rke2-uninstall.sh
+++ b/bundle/bin/rke2-uninstall.sh
@@ -121,7 +121,7 @@ uninstall_remove_files()
     rm -rf /etc/cni
     rm -rf /opt/cni/bin
     rm --one-file-system -rf /var/lib/kubelet || true
-    rm -rf "${RKE2_DATA_DIR}" || error "Failed to remove /var/lib/rancher/rke2"
+    rm -rf -- "${RKE2_DATA_DIR}" || error "Failed to remove ${RKE2_DATA_DIR}"
     rm -d /var/lib/rancher || true
 
     if type fapolicyd >/dev/null 2>&1; then


manuelbuil (approved):

manuelbuil:
@philippebi you should create an issue and link it to the PR. That way QA can easily test it

bundle/bin/rke2-uninstall.sh
```diff
@@ -121,7 +121,7 @@ uninstall_remove_files()
     rm -rf /etc/cni
     rm -rf /opt/cni/bin
     rm --one-file-system -rf /var/lib/kubelet || true
-    rm -rf "${RKE2_DATA_DIR}" || error "Failed to remove /var/lib/rancher/rke2"
+    rm -rf -- "${RKE2_DATA_DIR}" || error "Failed to remove ${RKE2_DATA_DIR}"
```
brandond, line 124: > RKE2_DATA_DIR is only defaulted in the script and could be wrongly defined by a user outside of it.

I don't see how adding `--` to terminate flag parsing makes this any safer. If the user sets `RKE2_DATA_DIR=/` it will still happily `rm -rf /`. The only thing this does is prevent `rm` from confusing the data-dir arg with additional flags, if for some reason the path starts with a `-` and I'm not sure that would even work to begin with.

What actual problem does this solve?

brandond (changes requested):

philippebi:
This doesn't solve any particular issue, but it's recommended to ensure that names with dashes don't become options.

See https://www.shellcheck.net/wiki/SC2035

brandond:
@philippebi if we're going to do this just to satisfy shellcheck then we should also change all the other places that `rm` a path specified by a variable, for example: 

* https://github.com/rancher/rke2/blob/548021824bd227e7e6d844619b1673b9458f79ae/bundle/bin/rke2-uninstall.sh#L28
* https://github.com/rancher/rke2/blob/548021824bd227e7e6d844619b1673b9458f79ae/bundle/bin/rke2-uninstall.sh#L115-L117
* https://github.com/rancher/rke2/blob/548021824bd227e7e6d844619b1673b9458f79ae/bundle/bin/rke2-uninstall.sh#L143

brandond (approved):

dereknola (approved):

PR #7382, opened 2024-12-03
Add runtime classes hook and runtimes chart

vitorsavian:
#### Proposed Changes ####

- Fix for runtimes class chart not being used
- Add a hook for ownership of runtime classes that rke2 supports

#### Types of Changes ####

- Bugfix

#### Verification ####

- first run a rke2 server and see if the runtime classes are there
```bash
root@test:/home/test# kubectl get runtimeclasses --all-namespaces
NAME                  HANDLER               AGE
crun                  crun                  6m7s
nvidia                nvidia                15m
nvidia-experimental   nvidia-experimental   6m7s
```
- see the info in the runtime class 
```bash
root@test:/home/test# kubectl describe runtimeclass crun
Name:         crun
Namespace:
Labels:       app.kubernetes.io/managed-by=Helm
Annotations:  meta.helm.sh/release-name: rke2-runtimeclasses
              meta.helm.sh/release-namespace: kube-system
API Version:  node.k8s.io/v1
Handler:      crun
Kind:         RuntimeClass
Metadata:
  Creation Timestamp:  2024-12-05T15:58:35Z
  Resource Version:    1638
  UID:                 f51f3d75-228a-45e7-841f-01f22e5cf9f4
Events:                <none>
```

- delete the labels and annotations from a runtime class
```bash
kubectl patch runtimeclass crun --type='json' -p='[{"op": "remove", "path": "/metadata/labels"},{"op": "remove", "path": "/metadata/annotations"}]'
```

- then see the runtime class again
```shell
root@test:/home/test# kubectl describe runtimeclass crun
Name:         crun
Namespace:
Labels:       <none>
Annotations:  <none>
API Version:  node.k8s.io/v1
Handler:      crun
Kind:         RuntimeClass
Metadata:
  Creation Timestamp:  2024-12-05T15:58:35Z
  Resource Version:    2624
  UID:                 f51f3d75-228a-45e7-841f-01f22e5cf9f4
Events:                <none>
```
- restart rke2 and describe the runtime class, see that the labels and annotations where updated
```shell
root@test:/home/test# kubectl describe runtimeclass crun
Name:         crun
Namespace:
Labels:       app.kubernetes.io/managed-by=Helm
Annotations:  meta.helm.sh/release-name: rke2-runtimeclasses
              meta.helm.sh/release-namespace: kube-system
API Version:  node.k8s.io/v1
Handler:      crun
Kind:         RuntimeClass
Metadata:
  Creation Timestamp:  2024-12-05T15:58:35Z
[...]

PR #8915, opened 2025-09-10
Fix runtimeclass ownership takeover

brandond:
#### Proposed Changes ####

Fix runtimeclass ownership takeover

Replace relabeling of runtimeclasses to transfer ownership to the chart, with native helm `--take-ownership` support, now that helm-controller supports it.

This was revealed while troubleshooting an issue where a single etcd-only node would exit with a fatal error following a restart, if the apiserver on the control-plane node experienced a disruption while the etcd-only node is starting up. The runtimeclass fatal error does not include the actual error that occurred, which made determining root cause difficult.

https://github.com/rancher/rke2/blob/9b1d008f90792ca36c2de02bec147330e5851e7a/pkg/rke2/rc.go#L44-L46

#### Types of Changes ####

bugfix/tech debt

#### Verification ####

* Same as https://github.com/rancher/rke2/pull/7382

#### Linked Issues ####

* Seen as part of https://github.com/rancher/rke2/issues/8921 - lack of error message in runtimeclass fatal exit obscures the underlying problem

#### User-Facing Change ####
```release-note
```


manuelbuil:
Stupid question: why were we labelling runtimeclasses?

brandond:
We were doing that so that helm wouldn't complain if someone had manually added them, prior to our bundling them all in the helm chart. Adding these labels and annotations made helm believe it owned them already. We can now do that directly with helm with the --take-ownership flag.

manuelbuil:
> We were doing that so that helm wouldn't complain if someone had manually added them, prior to our bundling them all in the helm chart. Adding these labels and annotations made helm believe it owned them already. We can now do that directly with helm with the --take-ownership flag.

Perfect, thanks for the explanation

PR #9006, opened 2025-10-07
ADR about security-responder

manuelbuil:
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/010-security-responder.md
+++ b/docs/adrs/010-security-responder.md
@@ -50,6 +50,7 @@ The collected data will include the following information:
 - serverNodeCount
 - agentNodeCount
 - cni-plugin
+- ingress-controller
 - os
 - selinux
 
@@ -65,19 +66,23 @@ Example payload structure:
     "serverNodeCount": 3,
     "agentNodeCount": 2,
     "cni-plugin": "flannel",
+    "ingress-controller": "rke2-ingress-nginx",
     "os": "ubuntu",
     "selinux": "enabled"
   }
 }
 ```
 
-The `clusteruuid` is needed to differentiate between different deployments (the UUID of `kube-system`). It is completely random and does not expose privacy considerations.
+The `clusteruuid` is needed to differentiate between different deployments (the UUID of `kube-system`). It is completely random and does not expose privacy considerations. We could even consider hashing it to increase the obfuscation.
 
 ### Configuration Interface Example
 
+The security-responder is packaged using a helm chart. We can interact with it as we do with other helm charts. For example, to disable it: 
+
 ```yaml
 # /etc/rancher/rke2/config.yaml
-security-responder-enabled: true # default
+disable:
+- rke2-security-responder
 ```
 
 ## Alternatives Considered


l-mb:
This looks good to me :+1:

pgonin:
Ok for me
Which team will run the 'server' part?
Have you had some estimates on bandwidth and load numbers? (Using ballpark numbers of number of RKE2 clusters currently running)

I would add that there should be a possibility in Rancher / Rancher Prime deployment context to setup for the data to be collected by Rancher (with Rancher acting as "proxy" to needed data)

l-mb:
@pgonin We'll have EIO stand up another `upgrade-responder` deployment, they've got that fairly automated, not an issue.

RMS/Prime gathering similar data about systems under management is a very different discussion, let's have that elsewhere please.

pgonin:
Agreed, it should be a second step. To be discussed together with Rancher actually

docs/adrs/010-security-responder.md
```diff
@@ -0,0 +1,102 @@
...
+- **Default State**: Enabled by default (opt-out well documented)
+
+### Data Collection
+
+The collected data will include the following information:
+- Kubernetes version
+- clusteruuid
+- nodeCount
+- serverNodeCount
+- agentNodeCount
```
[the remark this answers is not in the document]
l-mb: Agreed.

I think the intent is to differentiate between pure control plane nodes and those that run workloads. For example, SUSE's pricing model excludes pure control plane nodes.

docs/adrs/010-security-responder.md
```diff
@@ -0,0 +1,97 @@
...
+    "serverNodeCount": 3,
+    "agentNodeCount": 2,
+    "cni-plugin": "flannel",
+    "os": "ubuntu",
+    "selinux": "enabled"
+  }
+}
+```
+
+The `clusteruuid` is needed to differentiate between different deployments (the UUID of `kube-system`). It is completely random and does not expose privacy considerations.
```
brandond: Probably not worth addressing here, but I have seen some users that build images with RKE2 already installed and started. This is a bad idea for several reasons, but it happens. When this is done, the clusters will have the same root CAs - and the same kube-system namespace UID. This will cause them to look like the same cluster to anything using that UID as a cluster UID. I don't know that there's a good way to work around this, but it is something to be aware of.

docs/adrs/010-security-responder.md
```diff
@@ -0,0 +1,95 @@
...
+- **Default State**: Enabled by default (opt-out well documented)
+
+### Data Collection
+
+The collected data will include the following information:
+- Kubernetes version
+- clusteruuid
+- serverNodeCount
+- agentNodeCount
+- cni-plugin
```
dereknola: I think collecting the ingress-controller information would be vital as well. We want to know if people are starting to use traefik over ingress-nginx (or using the `none` and rolling an external ingress). That would help inform whether we should continue with traefik integration in RKE2 or try with another ingress choice.
```suggestion
- cni-plugin
- ingress-controller
```

docs/adrs/010-security-responder.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+The `clusteruuid` is needed to differentiate between different deployments (the UUID of `kube-system`). It is completely random and does not expose privacy considerations.
+
+### Configuration Interface Example
+
+```yaml
+# /etc/rancher/rke2/config.yaml
+security-responder-enabled: true # default
```
dereknola:
```suggestion
security-responder: true # default
```

Having "enabled" in the name is redundant for a bool flag. We don't follow that style or any of our other bool flags like `embedded-registry` or `supervisor-metrics`.

dereknola (changes requested):

manuelbuil:
> nit on how to disable

Could you check again Brad? Thanks

Issue #9143, opened 2025-10-23
tests: update e2e tests to use images from the rancher org


mgfritch:
There are several e2e tests using images that can could cause test failures due to dockerhub ratelimits (or otherwise). 

Let's attempt to replace them using busybox or another rancher org alternative. 

```
git grep -in 'image:' tests/* | sed -e 's/\(.*\):.*image: \(.*\)/\2/' | sort -u
```

- [ ] gcr.io/kubernetes-e2e-test-images/dnsutils:1.3
- [ ] mbuilsuse/pstools:v0.2.0
- [x] nginx
- [x] nginx:stable-alpine
- [x] praqma/network-multitool
- [x] rancher/mirrored-pause:3.6
- [x] ranchertest/mytestcontainer

mgfritch:
Intentionally leaving these images unchanged:

- gcr.io/kubernetes-e2e-test-images/dnsutils:1.3
https://github.com/rancher/rke2/pull/9144#discussion_r2470167621

- mbuilsuse/pstools:v0.2.0
@manuelbuil, can you take a look?

manuelbuil:
@mgfritch how could we move the image to an internal rancher repo?

PR #9144, opened 2025-10-23
tests: update e2e tests to use images from the rancher org

mgfritch:
Issue: https://github.com/rancher/rke2/issues/9143

### DIFF
--- a/tests/docker/resources/volume-test.yaml
+++ b/tests/docker/resources/volume-test.yaml
@@ -19,7 +19,10 @@ metadata:
 spec:
   containers:
   - name: volume-test
-    image: rancher/mirrored-pause:3.6
+    image: rancher/mirrored-library-busybox:1.36.1
+    command:
+      - sleep
+      - infinity
     imagePullPolicy: IfNotPresent
     volumeMounts:
     - name: volv


brandond (approved):

tests/e2e/kine/kine_test.go
```diff
@@ -161,7 +161,7 @@ var _ = Describe("Verify Basic Cluster Creation with Kine", Ordered, func() {
 	It("Verifies dns access", func() {
 		_, err := tc.DeployWorkload("dnsutils.yaml")
 		Expect(err).NotTo(HaveOccurred())
-		cmd := "kubectl --kubeconfig=" + tc.KubeconfigFile + " exec -i -t dnsutils -- nslookup kubernetes.default"
+		cmd := "kubectl --kubeconfig=" + tc.KubeconfigFile + " exec -i -t dnsutils -- nslookup kubernetes.default.svc.cluster.local"
```
mgfritch: Potentially less useful as a test since the BusyBox implementation of `nslookup` does not retry with an appended search domain during a DNS`NXDOMAIN` error.

We might want to consider moving back to the `gcr.io/kubernetes-e2e-test-images/dnsutils:1.3` image or some other alternative.

tests/docker/resources/volume-test.yaml
```diff
@@ -19,7 +19,10 @@ metadata:
 spec:
   containers:
   - name: volume-test
-    image: rancher/mirrored-pause:3.6
+    image: rancher/mirrored-library-busybox:1.36.1
```
mgfritch: Either `mirrored-pause` or `mirrored-library-busybox` would be sufficient 🤷

tests/e2e/kine/kine_test.go
brandond: I think assuming that the test cluster uses the default domain suffix is probably OK?

brandond:
🚀 🚀 🚀

PR #9207, opened 2025-11-06
Improve PR Trivy Scanning Reports

dereknola:
#### Proposed Changes ####
- Switch to a more secure Tag system for the Trivy scans on  PR. RKE2 image and images lists from the PR are created in a read-only environment and then uploaded as artifacts. Then a privileged environment based on 'master' processes and comments the results. 
- Expand the images scanned beyond just the PR's `rke2-runtime` image to include the other "core" and "hardened" images. Mirrored images like Cilium, Vsphere, and Harvester are not included, as we don't have control over fixing any CVEs present in those images (and also the number of images proved too large for a single comment). 
- Produces a compact trivy report to allow a large amount of images within the `gh` CLI limit of 64K characters per comment.

#### Types of Changes ####
CI Improvements

#### Verification ####
See https://github.com/dereknola/rke2/pull/4#issuecomment-3499171481 as example of what the reports will look like. 
**Note:** You can't use the new system until this PR is merged (that's part of the security improvements).

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -12,7 +12,7 @@ awk '{print $1}' << EOF > build/images-core.txt
 EOF
 
 # use PULL_CMD=echo to print out the images without pulling them
-PULL_CMD=${PULL_CMD:-"docker image pull --quiet"}
+PULL_CMD="${PULL_CMD:-docker image pull --quiet}"
 
 xargs -n1 -t $PULL_CMD << EOF >> build/images-core.txt
     ${REGISTRY}/rancher/hardened-kubernetes:${KUBERNETES_IMAGE_TAG}


dereknola:
Got rid of the quotes around `"$PULL_CMD"` to allow the normal `docker image pull --quiet` to inject as normal.

scripts/build-images
```diff
@@ -11,7 +11,10 @@ awk '{print $1}' << EOF > build/images-core.txt
     ${REGISTRY}/${REPO}/${PROG}-runtime:${DOCKERIZED_VERSION}
 EOF
 
-xargs -n1 -t docker image pull --quiet << EOF >> build/images-core.txt
+# use PULL_CMD=echo to print out the images without pulling them
+PULL_CMD=${PULL_CMD:-"docker image pull --quiet"}
```
brandond: idk if this matters, but having the quotes inside the variable expansion seems weird?
```suggestion
PULL_CMD="${PULL_CMD:-docker image pull --quiet}"
```

dereknola: It doesn't change the output, but I can include it.

vitorsavian (approved):

Issue #9321, opened 2025-12-04
Duplicate cleanup function in rke2-uninstall.sh script


kwentine:
The [cleanup](https://github.com/rancher/rke2/blob/5b2f834f616ec02a293082f6f7376e128c90002e/bundle/bin/rke2-uninstall.sh#L145) function is defined twice in the `rke2-uninstall.sh` script.

The latter version takes heed of the `NO_COLOR` variable, so I assume it is the one to keep ?

brandond:
It looks like this snuck in when addressing merge conflicts on https://github.com/rancher/rke2/pull/6237. Keeping the second one would be correct, yet.

kwentine:
Thanks for confirming, I will open a PR.

aganesh-suse:
Closing based on: https://github.com/rancher/rke2/issues/9340

PR #9331, opened 2025-12-05
Remove redundant cleanup function definition

kwentine:
Addresses:
- #9321 

Remove harmless but redundant `cleanup` shell function definition from `rke2-uninstall.sh`.


brandond:
Thanks!

Issue #8499, opened 2025-07-10
Change images to have always the full registry in the name and allow override of the prefix


tomaluca95:
**Is your feature request related to a problem? Please describe.**

Some images now are named with rancher/ and some other are named with index.docker.io/rancher/.

This is annoying, for example, writing the registries.yaml file.

Managing an install from internal registry is a bit triky.

**Describe the solution you'd like**

All images named with the index.docker.io/rancher/ name and allow the replacement ofall images with a private registry prefix (for example oci.example.com/mirror/docker.io/rancher/ ).

**Describe alternatives you've considered**

Using registries with regexes and some triky configs

brandond:
What specifically is this breaking? This is standard image ref behavior. If the image ref does not specify a registry, docker.io is assumed. docker.io, index.docker.io, and registry-1.docker.io are all aliases for the same registry and should be handled as such.

All of this is just due to disappointing architectural choices made by Docker long before Kubernetes existed.

tomaluca95:
Now to install from a local registry I need to configure the system something like

```yaml
      mirrors:
        index.docker.io:
          disable-default-registry-endpoint: true
          endpoint:
            - "https://oci.example.com"
          rewrite:
            "^index.docker.io/rancher/(.*)": "mirror/docker.io/rancher/$1"
        docker.io:
          disable-default-registry-endpoint: true
          endpoint:
            - "https://oci.example.com"
          rewrite:
            "^rancher/(.*)": "mirror/docker.io/rancher/$1"
      configs:
        "oci.example.com":
          auth:
            username: "rke2-installer"
            password: "rke2-installer-password"
```

Seems to work but I'm not quite secure about that.

The ability to have a shared prefix that can be easily replaced can help setting up a cluster with no need to mingle with mirrors.

My idea is to be able to replace the prefix and point to a private registry that is not a docker.io mirror, only has the right images

brandond:
You shouldn't need to do that. As I said, index.docker.io and docker.io are the same thing and should be handed as such by containerd. This was regressed a bit in containerd 2.0, and I have a PR open with their project to address that:
* https://github.com/containerd/containerd/pull/12053

brandond:
Also note that different projects disagree on what to use as the backend for docker.io. containerd uses registry-1.docker.io as the endpoint for docker.io. go-containerregistry uses index.docker.io, which is where the references you're seeing come from. All three names should be interchangeable but it can be challenging to coordinate this across projects.

[...]

Issue #3227, opened 2022-08-11
Multiple registries.yaml rewrites order is not preserved in containerd config.toml


HoustonDad:
**Environmental Info:**
RKE2 Version: v1.21, v1.22, v1.23

Node(s) CPU architecture, OS, and Version: 
CentOS 7.9.2009 x86_64

Cluster Configuration:
1 Node Server/Agent combo

**Describe the bug:**
When using multiple rewrites in registries.yaml, the order of those rewrites are not preserved in the containerd configuration. This leads to RKE2 pulling from the wrong places.

**Steps To Reproduce:**

```
[root@localhost ~]# curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL=v1.23 sh -
[root@localhost ~]# cat registries.yaml
mirrors:
  quay.io:
    endpoint:
      - "https://harbor.internal.com:5000"
    rewrite:
      "strimzi/(.*)": "registry1.dso.mil/ironbank/opensource/strimzi/$1"
      "jetstack/(.*)": "registry1.dso.mil/ironbank/jetstack/$1"
      "(.*)": "quay.io/$1"
[root@localhost ~]# mkdir -p /etc/rancher/rke2
[root@localhost ~]# cat > /etc/rancher/rke2/config.yaml <<EOF
token: usercreatedtokenstring
write-kubeconfig-mode: "0640"
tls-san:
  - rke2.houstondad.com
EOF

[root@localhost ~]# cp registries.yaml /etc/rancher/rke2/
[root@localhost ~]# systemctl enable rke2-server && systemctl start rke2-server

```

**Expected behavior:**
It's expected that the order of rewrites is honored and preserved.

**Actual behavior:**
The order of rewrites appears to be put in via alphabetical order. 

**Additional context / logs:**

```
[root@localhost ~]# cat registries.yaml
mirrors:
  quay.io:
    endpoint:
      - "https://harbor.internal.com:5000"
    rewrite:
      "strimzi/(.*)": "registry1.dso.mil/ironbank/opensource/strimzi/$1"
      "jetstack/(.*)": "registry1.dso.mil/ironbank/jetstack/$1"
      "(.*)": "quay.io/$1"

[root@localhost ~]# cat /var/lib/rancher/rke2/agent/etc/containerd/config.toml
[plugins.opt]
  path = "/var/lib/rancher/rke2/agent/containerd"
[plugins.cri]
  stream_server_address = "127.0.0.1"
  stream_server_port = "10010"
[...]

PR #9404, opened 2025-12-11
Add path prefix support to `system-default-registry`

ngharo:
#### Proposed Changes ####

Extends system-default-registry to support optional repository path prefixes, allowing more flexible registry configurations.

This will make it easier to use common repository mirroring in registry software like Harbor. 

Examples for rke2-runtime image resolution:
- "example.com" → `example.com/rancher/rke2-runtime:latest`
- "example.com/docker.io" → `example.com/docker.io/rancher/rke2-runtime:latest`
- "example.com:5000/myrepo" → `example.com:5000/myrepo/rancher/rke2-runtime:latest`

This will require a change to the documentation.

#### Types of Changes ####

New feature, with backwards compatibility.

#### Linked Issues ####

* #3227 - Helps alleviate this issue when encountered for rke2 image rewrites.
* #8499

#### User-Facing Change ####

```release-note
system-default-registry now supports custom registry with repository prefix.
```

### DIFF
--- a/pkg/images/images_registry_test.go
+++ b/pkg/images/images_registry_test.go
@@ -98,7 +98,7 @@ func Test_UnitResolver_NewResolver(t *testing.T) {
 	}
 }
 
-func TestSplitRegistryAndPath(t *testing.T) {
+func Test_UnitsplitRegistryAndPath(t *testing.T) {
 	tests := []struct {
 		name         string
 		input        string


brandond:
I am against this change. Registry settings should should affect only that - the registry. If you want to rewrite or prefix paths for images, with or without also changing changing the registry, that should be done in registries.yaml, with rewrites.

Note that k3s also has a system-default-registry flag that behaves the same as rke2.

ngharo:
I do agree that the naming is bit loose when it comes to this change.

However, due issues with rewrite ordering in the linked issue, users are left without a good path forward.

> Note that k3s also has a system-default-registry flag that behaves the same as rke2.

I don't think this is true. k3s happily takes the format I'm proposing in this PR:

```
$ sudo k3s server --system-default-registry=registry.internal.com/docker.io --write-kubeconfig-mode=644 --write-kubeconfig=/tmp/k3s.yaml
```

```
$ KUBECONFIG=/tmp/k3s.yaml kubectl get pods -A -o jsonpath='{range .items[*]}{range .spec.containers[*]}{.image}{"\n"}{end}{end}' | sort -u
registry.internal.com/docker.io/rancher/klipper-helm:v0.8.0-build20230510
registry.internal.com/docker.io/rancher/klipper-lb:v0.4.3
registry.internal.com/docker.io/rancher/local-path-provisioner:v0.0.24
registry.internal.com/docker.io/rancher/mirrored-coredns-coredns:1.10.1
registry.internal.com/docker.io/rancher/mirrored-library-traefik:2.9.10
registry.internal.com/docker.io/rancher/mirrored-metrics-server:v0.6.2
```

I also should note that Rancher manager also accepts this format for `systemDefaultRegistry` and is pretty common format when used to overwrite registry values in Helm charts too.

To me, this seems like an easy win for usability with no impact to existing configurations.

brandond:
I intentionally didn't add prefix support when adding this feature, as I wanted to hold a strong line on default **registry** only affecting the **registry**, not the image name itself - what OCI spec refers to as the name, and Docker Hub calls the repository.

However, you are correct that Rancher has long muddied the water by allowing the default registry to contain a path component, and I wasn't as strict when adding this option to K3s... so I guess we might as well allow it here as well.

pkg/images/images.go
```diff
@@ -39,8 +39,9 @@ type ResolverOpt func(name.Reference) (name.Reference, error)
 
 // Resolver provides functionality to resolve an RKE2 image name to a reference.
 type Resolver struct {
-	registry  name.Registry
-	overrides map[string]name.Reference
+	registry   name.Registry
+	repoPrefix string // optional path segment to prepend to image repository
```
brandond: `repository` is a Docker Hub thing. OCI spec refers to this as the name, and that is what we should refer to it as as well. Ref:https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints


```suggestion
	namePrefix string // optional path segment to prepend to image name
```

pkg/images/images.go
```diff
@@ -144,8 +150,14 @@ func (r *Resolver) GetReference(i string, opts ...ResolverOpt) (name.Reference,
...
 	} else {
+		// Build the full prefix (registry/repoPrefix) while resolving default images
+		var prefix string
+		if r.repoPrefix != "" {
+			prefix = fmt.Sprintf("%s/%s", r.registry.Name(), r.repoPrefix)
+		}
+
 		// No override; get compile-time default
-		d, err := getDefaultImage(i)
+		d, err := getDefaultImage(i, prefix)
```
brandond: nit: rather than building this here, and passing the prefix into `getDefaultImage`, I think this might fit better into `setRegistry` - since that's the function that's supposed to be responsible for applying the overrides from system-default-registry.

brandond (changes requested):

dereknola (changes requested):
Align unit tests names with [our format guidelines](https://github.com/rancher/rke2/blob/master/developer-docs/testing.md#format)

dereknola (approved):

Issue #9434, opened 2025-12-18
[New Minor] Update k8s and k3s to v1.35.0


mgfritch:
Tracking issue for all associated PRs necessary for producing a RKE2 release-1.35 branch

mdrahman-suse:
Validated as part of v1.35.0+rke2r1 Minor release check

PR #9435, opened 2025-12-19
Update to kubernetes v1.35.0 and golang v1.25.5

mgfritch:
#### Proposed Changes ####

Update to kubernetes v1.35.0 and golang v1.25.5

- go1.25.5
- https://github.com/k3s-io/k3s/commit/a6c6cd15c0c42ec9fce21f8ad5f42aa74fddb4f2

#### Types of Changes ####

minor release

#### Verification ####

Check version

#### Linked Issues ####

https://github.com/rancher/rke2/issues/9434

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

Depends:
- [x] https://github.com/rancher/vsphere-charts/pull/118
- [x] https://github.com/rancher/rke2-charts/pull/847

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -32,10 +32,10 @@ charts:
   - version: v0.27.403
     filename: /charts/rke2-flannel.yaml
     bootstrap: true
-  - version: 1.12.100
+  - version: 1.13.000
     filename: /charts/rancher-vsphere-cpi.yaml
     bootstrap: true
-  - version: 3.5.0-rancher200
+  - version: 3.6.0-rancher100
     filename: /charts/rancher-vsphere-csi.yaml
     bootstrap: true
   - version: 0.2.1100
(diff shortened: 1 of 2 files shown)


scripts/validate-charts
brandond: I don't think this change is correct. This will allow each individual version component (major, minor, and patch) to be less than or equal to the target version. The logic here should probably be replaced with a proper semver compare but the current failure is not wrong - this SHOULD be failing at the moment, as you have not updated the vsphere charts in rke2-charts or rke2 to support 1.35.0 yet.

brandond: Its a warning because sometimes upstream projects don't add support for a new minor before we need to get a release out. We only want to block the release if the images are wrong, not because something doesn't support this minor yet.

Yes the script is ugly and should probably be converted to golang for technical correctness but it's not wrong. The chart needs to be updated. Note that this is checking CPI chart v1.12.1 but your PR to add v1.35 support bumped the chart v1.13.0: 
https://github.com/rancher/vsphere-charts/pull/118/files#diff-50ddca5e4037a6988a4f9eb12c30c2c5fa8730fb2cc7d07f1517feee49e01804R24

Pull your chart changes through to rke2-charts and this PR, and we can improve the script later.

mgfritch: Issue to track validation script improvements:  https://github.com/rancher/rke2/issues/9441

cwayne18 (approved):

PR #9511, opened 2026-01-11
Use 0600 file permissions only for CIS mode

### LINKED ISSUE #9510: Use 0600 file permissions only for CIS mode
## Summary

In `pkg/executor/staticpod/staticpod.go`, the `isAnyMode()` function applies 0600 permissions for both CIS and ETCD profile modes. This is overly restrictive - the strict 0600 permission (owner read/write only) should only apply to CIS mode, which has specific security requirements.

## Problem

- `isAnyMode()` returns true for both `ProfileModeCIS` and `ProfileModeETCD`
- ETCD mode gets 0600 permissions even though it doesn't require CIS-level security
- This creates inconsistency and overly restrictive permissions

## Proposed Solution

Use `isCISMode()` instead of `isAnyMode()` so that:
- CIS mode → 0600 (strict, as required for compliance)
- ETCD/None modes → 0644 (standard permissions)

luojiyin1987:
## Summary

Fixes #9510

Changes file permission logic in `pkg/executor/staticpod/staticpod.go` to use `isCISMode()` instead of `isAnyMode()`.

## Changes

- Replace `isAnyMode()` with `isCISMode()` for permission check
- CIS mode: uses 0600 (strict, required for compliance)
- ETCD/None modes: use 0644 (standard permissions)

## Test Plan

- Existing tests should pass
- Manual verification of manifest file permissions in different modes

### DIFF
--- a/pkg/executor/staticpod/staticpod.go
+++ b/pkg/executor/staticpod/staticpod.go
@@ -635,7 +635,7 @@ func (s *StaticPodConfig) writeTemplate(spec *podtemplate.Spec) error {
 	if err != nil {
 		return err
 	}
-	if s.ProfileMode.isAnyMode() {
+	if s.ProfileMode.isCISMode() {
 		return writeFile(manifestPath, b, 0600)
 	}
 	return writeFile(manifestPath, b, 0644)


luojiyin1987:
> A 600 permission is by design and not a mistake. See comments in the issue linked.

thank you for your answer

PR #9576, opened 2026-01-26
Fix setup-go cache key creation

dereknola:
#### Proposed Changes ####
- Align more closely with how the upstream `action/setup-go` generates their cache key. Ensure that on self-hosted runners, a `ImageOS` ENV is available, something that is expected by these actions. Deal with Windows/Linux differences in the key extensions. 
- Ensure all self-hosted runners are using the same Ubuntu24 image. This was defaulting still to ubuntu22 when not provided. 
- Upgrade existing self-hosted defined images to Ubuntu24 (from 22) 

#### Types of Changes ####
CI Bug Fixes

#### Verification ####
- Cache keys generated no longer are resulting in `setup-go-Linux-x64--go-1.25.5` and ensure when running on master (using upstreams cacheing code) no longer results in saved entries with `setup-go-Linux-x64-undefined-go-1.25.5`

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ permissions:
     id-token: write
 jobs:
   build-amd64:
-    runs-on: runs-on,runner=8cpu-linux-x64,run-id=${{ github.run_id }},image=ubuntu22-full-x64,hdd=64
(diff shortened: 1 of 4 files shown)


brandond:
Maybe sync the title with the current changes?

brandond (approved):

PR #9723, opened 2026-02-12
Change SLES 16 default installation to rpms

vitorsavian:
#### Proposed Changes ####

- Since SLES 16 was launched we have a necessity to have a rpm installation for users.

- Also we should also have a if to verify if it's sles 15 and then use tarball.

#### Types of Changes ####

- Feat

#### Verification ####

- Spin a SLES 16 OS who has the possibility to install iptables and then install rke2 with the install script
````bash
curl -sfL https://get.rke2.io | sh -
````

#### Testing ####

- No

#### Linked Issues ####

- https://github.com/rancher/rke2/issues/9722

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -102,6 +102,7 @@ check_target_ro() {
 
 # setup_env defines needed environment variables.
 setup_env() {
+    . /etc/os-release
     STORAGE_URL="https://rke2-ci-builds.s3.dualstack.us-east-1.amazonaws.com"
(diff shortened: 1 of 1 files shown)


brandond (approved):
Wait until after code freeze to merge, since this is live on commit.

manuelbuil (approved):

Issue #9728, opened 2026-02-16
Add guardrail to avoid promoting an worker node into control-plane


manuelbuil:
**Is your feature request related to a problem? Please describe.**

During 2025 at least two users accidentally turned a worker node into a control-plane node because it was forgotten that this worker node was already part of the cluster. This is possible to do because in the rke2-server systemd unit file, the following is used:
```
[Unit]
Conflicts=rke2-agent.service
```
That configuration instructs systemd to send a SIGTERM to the rke2-agent.service to shut it down so that the rke2-server can start. Therefore, it does not prevent the transformation of a worker into a control-plane.

This can be dangerous if there is only one control-plane in the cluster. In that case, when the worker becomes a control-plane node, the cluster has suddenly 2 control-plane nodes, which is suboptimal, and if the user notices this and kills the new control-plane node, the cluster will become unresponsive with the error:
```
level=info msg="Failed to test etcd connection: etcdserver: no leader"
```
because quorum is now impossible to reach.

**Describe the solution you'd like**

I would like to have an instruction in the rke2-server systemd unit file so that rke2-server can't be initiated if rke2-agent is already running. Transforming an agent into a control-plane is anyway a procedure that is not supported, so it should be prevented.

**Describe alternatives you've considered**

**Additional context**

aganesh-suse:
Closing based on: https://github.com/rancher/rke2/issues/9775

PR #9729, opened 2026-02-16
Prevent a node transform from agent/server to server/agent

manuelbuil:
#### Proposed Changes ####

This change is a guardrail to prevent the problem described in the issue. It makes it harder to start rke2-agent or rke2-server in a node that is already running the other role

#### Types of Changes ####

New feature

#### Verification ####

1 - Create an RKE2 cluster with one CP and one worker node
2 - Pretend that you forgot the worker node was already part of the node and prepare it to become an rke2-server (e.g. place the config.yaml, etc)
3 - In that worker node, run `systemctl enable --now rke2-server`
4a - Before this PR, the rke2-agent process in the node stops. The rke2-server process runs and the node becomes a control-plae
4b - By using this PR, the rke2-agent process never stops and the rke2-server process fails to start with the error:
```
Feb 16 13:25:17 ip-10-11-0-11 systemd[1]: Starting rke2-server.service - Rancher Kubernetes Engine v2 (server)...
Feb 16 13:25:18 ip-10-11-0-11 sh[111232]: 111231
Feb 16 13:25:18 ip-10-11-0-11 sh[111231]: Error: rke2-agent is running!
Feb 16 13:25:18 ip-10-11-0-11 systemd[1]: rke2-server.service: Control process exited, code=exited, status=1/FAILURE
Feb 16 13:25:18 ip-10-11-0-11 systemd[1]: rke2-server.service: Failed with result 'exit-code'.
Feb 16 13:25:18 ip-10-11-0-11 systemd[1]: Failed to start rke2-server.service - Rancher Kubernetes Engine v2 (server).
```

#### Linked Issues ####

https://github.com/rancher/rke2/issues/9728

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/bundle/lib/systemd/system/rke2-agent.service
+++ b/bundle/lib/systemd/system/rke2-agent.service
@@ -3,7 +3,6 @@ Description=Rancher Kubernetes Engine v2 (agent)
 Documentation=https://github.com/rancher/rke2#readme
 Wants=network-online.target
 After=network-online.target
-Conflicts=rke2-server.service
 
 [Install]
 WantedBy=multi-user.target
@@ -22,6 +21,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecCondition=/bin/sh -c 'if systemctl is-active --quiet rke2-server.service; then echo "Error: rke2-server is running!"; exit 1; fi'
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
 ExecStart=/usr/local/bin/rke2 agent
--- a/bundle/lib/systemd/system/rke2-server.service
+++ b/bundle/lib/systemd/system/rke2-server.service
@@ -3,7 +3,6 @@ Description=Rancher Kubernetes Engine v2 (server)
 Documentation=https://github.com/rancher/rke2#readme
 Wants=network-online.target
 After=network-online.target
-Conflicts=rke2-agent.service
 
 [Install]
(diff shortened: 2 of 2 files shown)


manuelbuil:
ing_migration test already! yuhu!

brandond:
lgtm. do we want the unit to be marked as failed, or simply not start? I think simply not starting is correct, but wanted to confirm.

> when an ExecCondition= command exits with exit code 1 through 254 (inclusive), the remaining commands are skipped and the unit is not marked as failed. However, if an ExecCondition= command exits with 255 or abnormally (e.g. timeout, killed by a signal, etc.), the unit will be considered failed (and remaining commands will be skipped).

Might also confirm this works on EL9 or whatever the most ancient linux distro we currently support is.

brandond (approved):

manuelbuil:
> lgtm. do we want the unit to be marked as failed, or simply not start? I think simply not starting is correct, but wanted to confirm.
> 
> > when an ExecCondition= command exits with exit code 1 through 254 (inclusive), the remaining commands are skipped and the unit is not marked as failed. However, if an ExecCondition= command exits with 255 or abnormally (e.g. timeout, killed by a signal, etc.), the unit will be considered failed (and remaining commands will be skipped).
> 
> Might also confirm this works on EL9 or whatever the most ancient linux distro we currently support is.

Simply not start should be good. It still provides the log, so user should be aware of what's going on
Good point about ancient distros

manuelbuil:
Works on RHEL 9,6 which is supposed to be the oldest OS we support. It uses `systemd 252`

PR #9848, opened 2026-03-02
[ADR] Add rke2-patcher ADR for review

manuelbuil:
#### Proposed Changes ####

ADR on a potential new tool

#### Types of Changes ####

New feature

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/013-rke2-patcher-tool.md
+++ b/docs/adrs/013-rke2-patcher-tool.md
@@ -1,0 +1,132 @@
+# Prime tool: rke2-patcher
+
+## Established
+
+2026-03-20
+
+## Revisit by
+
+2026-06-11
+
+## Subject
+
+Currently, when a critical CVE is identified, users must typically wait for the next scheduled RKE2 monthly release for a fix and that can take up to four weeks. This delay leaves clusters vulnerable, forcing users to implement manual workarounds or restrictive security policies. These temporary measures are often labor-intensive, costly, and disruptive to standard operations. We can improve this process by allowing users to consume patched images as soon as they are built and verified, rather than waiting for the next full release cycle.
+
+## Status
+
+Proposed
+
+## rke2-patcher tool
+
+`rke2-patcher` is a CLI that inspects and patches selected RKE2 component images.
+
+Scope and compatibility:
+
+- requires RKE2 Prime clusters with `prime: true`
+- supports two execution modes:
+  - host binary mode on a control-plane node (or equivalent environment with Kubernetes API access)
+  - in-cluster mode (as a pod in the cluster)
+
+Goals:
+
+- reduce the CVE exposure window for selected RKE2 components
+- keep operations explicit and risk-controlled
+- avoid introducing a long-running in-cluster controller
+- keep patching compatible with existing RKE2 chart customization workflows (by using `HelmChartConfig`)
+- support disconnected or restricted environments by allowing alternate registries and scanner image overrides
+
+
+### Command surface
+
+- `rke2-patcher --version`
+  - always prints CLI version
+  - attempts to print connected cluster version from Kubernetes API `/version`
+
+- `rke2-patcher --config`
+  - prints effective/default/source values for runtime configuration
+  - includes registry, scanner mode/image/namespace/timeout, and state ConfigMap coordinates
+
+- `rke2-patcher image-cve <component>`
+  - resolves the current running image for the selected component
+  - default mode is cluster scan (`RKE2_PATCHER_CVE_MODE=cluster`) using an in-cluster Trivy Job
+  - supports local mode (`RKE2_PATCHER_CVE_MODE=local`) using local scanners (`trivy`, `grype` fallback)
+  - local mode uses a shared VEX cache at `$HOME/rke2-patcher-cache/vex/rancher.openvex.json`
+
+- `rke2-patcher image-list <component> [--with-cves] [--verbose]`
+  - lists ordered release tags for the running component repository
+  - marks tags currently in use by running pods
+  - applies the same 45-day eligibility policy as `image-patch`
+  - output is split into:
+    - `eligible tags` (patchable on current cluster)
+    - `newer tags requiring RKE2 upgrade` (visible, but blocked)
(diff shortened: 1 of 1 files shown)


macedogm:
@manuelbuil amazing idea! 👏🏻

manuelbuil:
> @manuelbuil amazing idea! 👏🏻

Thanks! But I'm just the executor. The idea came from @cwayne18

docs/adrs/011-rke2-patcher-tool.md
```diff
@@ -0,0 +1,119 @@
...
+  - key: `patch-limit-state.json`
+  - namespace: `RKE2_PATCHER_CVE_NAMESPACE` (default `rke2-patcher`)
+
+### Airgap and alternate registry support
+
+- `RKE2_PATCHER_REGISTRY` controls the registry endpoint used for tag discovery in `image-list` and `image-patch`.
+- Supported formats include host-only, host:port, and URL forms (`https://`/`http://`).
+- In disconnected environments, operators can point `RKE2_PATCHER_REGISTRY` to an internal mirror registry that contains the relevant Rancher component repositories and tags.
+- For cluster CVE mode, scanner image pull source is configurable with `RKE2_PATCHER_CVE_SCANNER_IMAGE`, allowing use of an internally mirrored Trivy image.
+- If in-cluster scanner-job image pulls are not possible, operators can use local scanning mode (`RKE2_PATCHER_CVE_MODE=local`) with a local scanner binary.
```
dereknola: An example of using a local scanner would be nice, perhaps move the example of the `RKE2_PATCHER_CVE_SCANNER_IMAGE` from below up one bullet point

- "... internally mirrored Trivy image."
```
<Example>
```
- "... local scanner binary"
```
<Example>
```

docs/adrs/013-rke2-patcher-tool.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Safety and compatibility guardrails
+
+- Prime guardrail: patching is allowed only when cluster charts indicate `global.prime.enabled=true`.
+- Minor-version guardrail: refuse forward patching to newer minor lines.
+- Time-window guardrail: refuse patch target tags outside a 45-day window from cluster zero-day.
+  - goal is to avoid users staying in the same RKE2 version forever
+  - zero-day is derived from running kube-apiserver image (`rancher/hardened-kubernetes`) build date
+  - `rke2-ingress-nginx` is exempt from this date-window rule
+- Reconcile gate: if stale patch entries from a previous RKE2 version exist, forward patching is blocked until reconcile is performed.
```
dereknola, line 85: Reading this, the workflow for upgrades is somewhat unclear to me. 

So you use the patcher... then you later upgrade RKE2... then you use the patcher again to remove your customer config overrides?

dereknola (changes requested):

docs/adrs/013-rke2-patcher-tool.md
dereknola: Perhaps this is irrelevant to an ADR, but would be necessary for any documentation IMO.

docs/adrs/011-rke2-patcher-tool.md
manuelbuil: Let me add an example. The output is basically the same but in one case there is a job created in the cluster. In the other case, it just uses the "trivy" binary in the node

docs/adrs/013-rke2-patcher-tool.md
manuelbuil: Yes, your understanding is correct. And this is the reasoning: Our QA team is currently verifying RKE2 releases. Imagine we use rke2-patcher to update one image. At that point, we will be slightly drifting from the RKE2 release and hence from what QA verified. As a result, we are introducing a (small) risk and an extra (small) complexity for our support team (to reproduce, etc). Once this cluster upgrades, I would like the user to go back to a "standard RKE2 cluster". On top of that, I want to avoid users from never upgrading rke2 and just using rke2-patcher to keep their components up to date.

We could automatically detect the upgrade and then remove all the "rke2-patcher customizations" but I thought that would be more disruptive. I would expect the user, as the last step of the upgrade, to remove the customizations one by one.

Yes, I need to document all this :)

dereknola (approved):

caroline-suse-rancher:
We would love to get more eyes here, including @rancher/release-team in case there's crossover we don't see

docs/adrs/013-rke2-patcher-tool.md
briandowns: What happens when the user doesn't do that after upgrade then runs the rke2 patcher again?

docs/adrs/013-rke2-patcher-tool.md
```diff
@@ -0,0 +1,132 @@
...
+Goals:
+
+- reduce the CVE exposure window for selected RKE2 components
+- keep operations explicit and risk-controlled
+- avoid introducing a long-running in-cluster controller
+- keep patching compatible with existing RKE2 chart customization workflows (by using `HelmChartConfig`)
+- support disconnected or restricted environments by allowing alternate registries and scanner image overrides
+
+
+### Command surface
```
briandowns, line 39: Are we using the same library and style of CLI to maintain a uniform CLI UX across the products?

briandowns (commented):
couple questons

docs/adrs/013-rke2-patcher-tool.md
manuelbuil: If it tries to patch the same component, rke2-patcher will refuse to do it and will ask for reconcile. I'm testing this here https://github.com/rancher/rke2-patcher/blob/main/tests/docker/reconcile_upgrade/reconcile_upgrade_test.go#L92-L106

docs/adrs/013-rke2-patcher-tool.md
manuelbuil: Yes, I'm using `github.com/urfave/cli/v2` like rke2 and k3s

cwayne18 (approved):

dereknola (approved):

Issue #9786, opened 2026-02-26
[EPIC] Migrate to Traefik as default ingress


caroline-suse-rancher:
Coming soon, Traefik will be the default ingress of RKE2. This is as a result of ingress-nginx retiring in March 2026.

Please add sub-issues and relevant PRs here.

- #9733

manuelbuil:
We still need to change the tarballs, so that default includes Traefik as well

brandond:
Moving to to-test for v1.36.0 release. @manuelbuil I see your note about needing to add traefik to the default tarball. We should also decide when we are going to **remove** ingress-nginx from that tarball...

mdrahman-suse:
# Validation on v1.36.0-rc3+rke2r1

- New cluster (amd64 and arm64)
- Upgrade cluster (manual and SUC) (amd64 and arm64)
- Dual stack cluster
- ipv6 only cluster
- Airgap cluster (Community and Prime) (amd64 and arm64)
- Airgap cluster upgrade (Community and Prime) (manual)
- Windows agent (Non-airgap, Airgap)

brandond:
@mdrahman-suse what was the exact upgrade path that lead to you expecting to see the image change from prime3 to prime4?

Can you show the status of the rke2-ingress-nginx helm chart, deployment, and pods?

mdrahman-suse:
So I upgraded manually from v1.35.4+rke2r1 (prime release) to v1.36.0-rc3+rke2r1 (prime staging) on airgap setup. 1 bastion, 1 server, 1 agent node using tarball. The pod `rke2-ingress-nginx-admission-create-m4cz2` was in `CrashLoopBackOff`
```console
kube-system   pod/rke2-ingress-nginx-admission-create-m4cz2                             0/1     CrashLoopBackOff   5 (79s ago)   21m
```
And upon running describe on that pod showed its trying to pull/extract `kube-webhook-certgen:v1.14.5-prime4` from staging registry. Once I downloaded the `rke2-images-ingress-nginx.linux-arm64.tar.gz` and saved to `/var/lib/rancher/rke2/agent/images` (and needed a restart), the state of the pod changed to  `Running`. Hence the expectation was that the cluster would use `nginx-ingress-controller:v1.14.5-prime4` as it is bundled in `rke2-images-ingress-nginx.linux-arm64.tar.gz` instead of `nginx-ingress-controller:v1.14.5-prime3` which was installed initially as part of v1.35.4+rke2r1. 

[...]

PR #9859, opened 2026-03-03
Add `prime` configuration

dereknola:
#### Proposed Changes ####
- Adds a `prime` bool to the server configuration. This directs the cluster to pull images from the SUSE Rancher Prime registry. 
- Pulls prime specific images available only on the Prime registry. 
- Adds a docker prime test for validation (does not run on PR as it needs access to secrets)
- Deprecation notice for the `RKE_INGRESS_CONTROLLER` env var. This was a misspelling and the corrected `RKE2_XX` version is now available. Since this made it to a GA release,we need to go through our standard cli deprecation policy. 

#### Types of Changes ####
New Feature

#### Verification ####
- Runs rke2 with `prime: true`
- See server pulls images from the Prime registry instead of dockerhub

#### Linked Issues ####
* https://github.com/rancher/rke2/issues/9861
* https://github.com/rancher/rke2/issues/9786

#### User-Facing Change ####

```release-note
A new `prime` flag for use by prime customers to access the SUSE Rancher Prime registry
```

### DIFF
--- a/pkg/images/images.go
+++ b/pkg/images/images.go
@@ -22,12 +22,12 @@ const (
 	ETCD                   = "etcd-image"
 	Pause                  = "pause-image"
 	CloudControllerManager = "cloud-controller-manager-image"
-	PrimeRegistry          = "registry.rancher.com"
 )
 
 // These defaults are overridden at build time and do not need to be updated here
 var (
 	DefaultRegistry                    = name.DefaultRegistry
+	PrimeRegistry                      = ""
 	DefaultEtcdImage                   = "rancher/hardened-etcd"
 	DefaultKubernetesImage             = "rancher/hardened-kubernetes"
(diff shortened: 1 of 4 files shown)


pkg/images/images.go
```diff
@@ -22,6 +22,7 @@ const (
 	ETCD                   = "etcd-image"
 	Pause                  = "pause-image"
 	CloudControllerManager = "cloud-controller-manager-image"
+	PrimeRegistry          = "registry.rancher.com"
```
brandond: Should we set this to `docker.io` by default, and override with `-X github.com/rancher/rke2/pkg/images/PrimeRegistry=$PRIME_REGISTRY` at build time? Since we are avoiding hardcoding it anywhere else.

brandond (approved):

PR #9922, opened 2026-03-12
Fix: Additional image gets error 'cp: cannot stat '/path/rke2-images-*.tar.gz*': No such file or directory' in airgapped install

RedtapeRaptor:
#### Proposed Changes ####

Installation fails in airgapped install if you have other artifacts in the folder. 

No update to docs needed

#### Types of Changes ####

Bugfix

#### Verification ####

Add other files into the folder of INSTALL_RKE2_ARTIFACT_PATH and try to install airgapped without additional images. 

#### Testing ####

Not covered, not sure how to create a test scenario for airgapped env or unit testin for this issue.

#### Linked Issues ####

N/A not found

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

I was thinking of using either find or grep for the files and compare null array. But then we have a dependancy for find or grep. So i used bash builtin nullglob instead.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -458,12 +458,20 @@ install_airgap_tarball() {
         gzip -dc "${TMP_AIRGAP_TARBALL}" > "${INSTALL_RKE2_AGENT_IMAGES_DIR}/rke2-images.${SUFFIX}.tar"
     fi
     # Search for and install additional rke2 images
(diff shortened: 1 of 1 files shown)


RedtapeRaptor:
Hi,
I dont have a good way to do that without using --set. Wich would be kind of hard to manage over time. I am closing this PR since i dont have a good solution for /bin/sh.

brandond:
Thanks for the PR regardless. If you think of something else, please feel free to reopen!

Issue #9694, opened 2026-02-10
Tarball installer improvements for rke2-selinux


pedromfcarvalho:
**Is your feature request related to a problem? Please describe.**
Rancher uses the tarball installer to install and update rke2. Using this installer isn't as compatible as the rpm installer with respect to the rke2-selinux pakcage.

**Describe the solution you'd like**
Make the tarball installer work well with rke2-selinux.

**Additional context**
SURE-10857

aganesh-suse:
### Validated on release-1.35 branch with version v1.35.3-rc2+rke2r1
  
## Environment Details

  
*Infrastructure*
- [x] Cloud
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release  | grep PRETTY
PRETTY_NAME="Red Hat Enterprise Linux 10.1 (Coughlan)"

$ uname -m
x86_64
```
  
*Cluster Configuration:*

```
HA : 3 server / 1 agent
```

*Config.yaml:*

```
token: xxxx
write-kubeconfig-mode: "0644"
node-external-ip: 1.1.1.1
selinux: true
debug: true
```
* Was selinux enabled: true
```
$ sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33
```
## Pre-Install Steps:
1. Install container-selinux and rke2-selinux: 
```
$ command -v dnf && sudo dnf install -y container-selinux
$ command -v dnf && sudo dnf install -y https://github.com/rancher/rke2-selinux/releases/download/v0.23.testing.2/rke2-selinux-0.23-2.el10.noarch.rpm
```
## Testing Steps

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/rke2 && sudo cp config.yaml /etc/rancher/rke2
```
2. Install RKE2
```
curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION='v1.35.3-rc2+rke2r1' INSTALL_RKE2_TYPE='server' INSTALL_RKE2_METHOD=tar sh -
```
3. Start the RKE2 service
```
$ sudo systemctl enable --now rke2-server
or 
[...]

PR #9923, opened 2026-03-12
Add restorecon if selinux is installed in tar to fix wrong contexts

vitorsavian:
#### Proposed Changes ####

- There are a problem with selinux and tarball usage together, rke2 selinux policy does not apply a correct context if the installation is with tarball. The need of this change is because of this use case, restorecon automation is needed to ensure the context is correct for selinux + tarball installations.

- I saw that when you run rke2 with selinux in SLEMicro after the tarball installation the rke2 binary does not have the right context, since SLEMicro normally does not let you apply the restorecon in the /var/ folders (because of the snapshots) we need to restorecon the binary for the context be applied in the root folder of rke2.  

#### Types of Changes ####

- Enhancement/Fix

##### First verification (el10 based OS)

- Install in a RPM environment with selinux installed with tar as the default installation
```
curl -sfL https://get.rke2.io | INSTALL_RKE2_METHOD=tar  sh - 
```

- Verify if the contexts are correct after the installation (I did with a install.sh with the changes so the unconfined is expected)
```
ls -lZ /usr/local/lib/systemd/system/rke2* 2>/dev/null
ls -lZ /usr/lib/systemd/system/rke2* 2>/dev/null

// Result
-rw-r--r--. 1 root root unconfined_u:object_r:container_unit_file_t:s0  11 Mar  6 02:04 /usr/local/lib/systemd/system/rke2-agent.env
-rw-r--r--. 1 root root unconfined_u:object_r:container_unit_file_t:s0 884 Mar  6 02:04 /usr/local/lib/systemd/system/rke2-agent.service
-rw-r--r--. 1 root root unconfined_u:object_r:container_unit_file_t:s0  11 Mar  6 02:04 /usr/local/lib/systemd/system/rke2-server.env
-rw-r--r--. 1 root root unconfined_u:object_r:container_unit_file_t:s0 884 Mar  6 02:04 /usr/local/lib/systemd/system/rke2-server.service
```

##### Second Verification (SLEMicro with selinux and tar install)

- Install RKE2 with rke2-selinux already installed
```bash
localhost:/home/vsavian # sh ./install.sh
[WARN]  /usr/local is read-only or a mount point; installing to /opt/rke2
[INFO]  finding release for channel stable
[INFO]  using v1.34.5+rke2r1 as release
[INFO]  downloading checksums at https://github.com/rancher/rke2/releases/download/v1.34.5%2Brke2r1/sha256sum-amd64.txt
[INFO]  downloading tarball at https://github.com/rancher/rke2/releases/download/v1.34.5%2Brke2r1/rke2.linux-amd64.tar.gz
[INFO]  verifying tarball
[INFO]  unpacking tarball file to /opt/rke2
[INFO]  updating tarball contents to reflect install path
[INFO]  moving systemd units to /etc/systemd/system
[INFO]  install complete; you may want to run:  export PATH=$PATH:/opt/rke2/bin
[INFO]  applying correct SELinux contexts to RKE2 files
```
- Verify rke2 binary to see if the binary has the correct context and see if you set selinux to true in config 
```bash
localhost:/home/vsavian # ls -lZ /opt/rke2/bin/rke2
-rwxr-xr-x. 1 root root unconfined_u:object_r:container_runtime_exec_t:s0 125268480 Mar  5 21:07 /opt/rke2/bin/rke2

localhost:/home/vsavian # cat /etc/rancher/rke2/config.yaml
selinux: true
```
- Run rke2 and then see if the pods come up
```bash
systemctl enable rke2-server --now

export KUBECONFIG=/etc/rancher/rke2/rke2.yaml

localhost:/home/vsavian # /var/lib/rancher/rke2/bin/kubectl get nodes -A -o wide
NAME                    STATUS   ROLES                AGE    VERSION          INTERNAL-IP      EXTERNAL-IP   OS-IMAGE                            KERNEL-VERSION            CONTAINER-RUNTIME
localhost.localdomain   Ready    control-plane,etcd   119s   v1.34.5+rke2r1   192.168.18.234   <none>        SUSE Linux Enterprise Server 16.0   6.12.0-160000.5-default   containerd://2.1.5-k3s1

```

- Then you can see that the context is applied for the folders in /var/lib/rancher
```bash
localhost:/home/vsavian # ls -lZ /var/lib/rancher/rke2/
total 4
drwxr-x---. 1 root root system_u:object_r:container_var_lib_t:s0      586 Mar 17 13:04 agent
lrwxrwxrwx. 1 root root system_u:object_r:container_var_lib_t:s0       58 Mar 17 13:04 bin -> /var/lib/rancher/rke2/data/v1.34.5-rke2r1-30ba78b5a109/bin
drwxr-xr-x. 1 root root system_u:object_r:container_runtime_exec_t:s0  54 Mar 17 13:03 data
drwx------. 1 root root system_u:object_r:container_var_lib_t:s0       94 Mar 17 13:03 server
```

#### Testing ####

- We have a install script test with selinux and tarball

#### Linked Issues ####

- https://github.com/rancher/rke2/issues/9694

#### User-Facing Change ####

```release-note
Install script will now use restorecon if rke2-selinux is installed in tarball installation 
```

#### Further Comments ####

- The verification will need to be done in every OS in our support matrix (that rke2-selinux support) but specially SLEMicro and RHEL

### DIFF
--- a/.github/workflows/nightly-install.yaml
+++ b/.github/workflows/nightly-install.yaml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         channel: [stable]
-        vm: [centos-9, alma-10, opensuse-leap, ubuntu-2404, windows-2022, windows-2025]
+        vm: [centos-9, alma-10, alma-10-with-selinux-tar, opensuse-leap, ubuntu-2404, windows-2022, windows-2025]
         include:
           - {channel: latest, vm: alma-10}
           - {channel: latest, vm: ubuntu-2404}
--- a/install.sh
+++ b/install.sh
@@ -67,6 +67,10 @@ fi
 #   - INSTALL_RKE2_SKIP_FAPOLICY
 #     If set, the install script will skip adding fapolicy rules
 #     Default is not set.
+#
+#   - INSTALL_RKE2_SKIP_RESTORECON
+#     If set, the install script will skip the resorecon for selinux.
+#     Default is not set.
 
 
 # info logs the given argument at info log level.
@@ -662,6 +666,30 @@ do_install_tar() {
 
     if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
         systemctl daemon-reload
+    fi
+
+    if [ -z "${INSTALL_RKE2_SKIP_RESTORECON}" ]; then
+        do_restorecon_tar
+    fi
+}
+
+# if rke2-selinux is installed, restorecon the systemd units and the rke2 folders that have the bin
+# since they will have wrong contexts after being installed with the tarball method
+do_restorecon_tar() {
+    if command -v rpm >/dev/null 2>&1; then
+        if rpm -q --quiet rke2-selinux; then
+            info "applying correct SELinux contexts to RKE2 files"
+
+            # this is for the .service files
+            restorecon -RT 0 -i /etc/systemd/system/rke2*
+            restorecon -RT 0 -i /usr/local/lib/systemd/system/rke2*
+            restorecon -RT 0 -i /usr/lib/systemd/system/rke2*
+
+            # this one is for the bin
+            restorecon -RT 0 -i /opt/rke2
+            restorecon -RT 0 -i /usr/local/bin/rke2*
+            restorecon -RT 0 -i /usr/bin/rke2*
+        fi
     fi
 }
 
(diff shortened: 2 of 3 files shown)


manuelbuil:
According to this comment: https://suse.slack.com/archives/C02DT51PU2C/p1771950430546379?thread_ts=1771932025.125959&cid=C02DT51PU2C if the policy is available, the context should be correct. Is this comment not accurate then?

vitorsavian:
@manuelbuil for what I understood about it, it's mostly correct but when we install with tarball the files come from the tar so their context does not change if the folder does not have the correct context

if you see this file in specific here
image

they do not get the correct context because their folder are setting the context and they are not being created by the process, so this files need to have restorecon applied to them.

if you see this part of the scriptlet you will see that we do the restorecon for the `.service` and the `.env` in systemd (which works if we already have this files in the system)

https://github.com/rancher/rke2-selinux/blob/master/policy/centos10/rke2-selinux.spec#L3-L28
image

I did the restorecon the same way that we do in rke2-selinux but only in the files that need.

dereknola (approved):

brandond (changes requested):

brandond (approved):

manuelbuil (approved):

PR #10135, opened 2026-04-09
test-suite: fix vault action failure

mgfritch:
Action `rancher-eio/read-vault-secrets` fails because it is
unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable

Fixup: c3cd103

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -25,6 +25,9 @@ jobs:
   build:
     name: Build RKE2 Images and Binary
     runs-on: ${{ github.repository == 'rancher/rke2' && 'runs-on,runner=16cpu-linux-x64,image=ubuntu24-full-x64' || 'ubuntu-24.04' }}
+    permissions:
+      contents: read
+      id-token: write
     steps:
     - name: Checkout
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6


.github/workflows/test-suite.yaml
```diff
@@ -25,6 +25,9 @@ jobs:
   build:
     name: Build RKE2 Images and Binary
     runs-on: ${{ github.repository == 'rancher/rke2' && 'runs-on,runner=16cpu-linux-x64,image=ubuntu24-full-x64' || 'ubuntu-24.04' }}
+    permissions:
+      contents: read
+      id-token: write
```
brandond, line 30: I had to look this up:
> The id-token: write permission allows the GITHUB_TOKEN to make an API call to generate an OIDC token (which is an extra security token, a standard one). The OIDC token is not dangerous in itself. It is minted and signed by GitHub and includes "claims". Claims they come with includes: repo name, environment, branch, etc. A lot of context. The OIDC token is used to make extra security checks (ensure a deployment is being requested via GitHub Actions, that it targets a valid environment and a valid branch when applicable).

.github/workflows/test-suite.yaml
```diff
@@ -46,7 +49,7 @@ jobs:
     - name: Install OS Packages
       run: sudo apt-get update && sudo apt-get install -y libarchive-tools g++-mingw-w64-x86-64 gcc-mingw-w64-x86-64
     - name: "Read secrets"
-      if: github.event_name != 'pull_request' && github.repository == 'rancher/rke2'
+      #if: github.event_name != 'pull_request' && github.repository == 'rancher/rke2'
```
brandond: Either delete this or restore it

brandond (changes requested):

.github/workflows/test-suite.yaml
mgfritch: We cannot remove this line, otherwise vault will fail on every PR from a private fork (github will disable the `id-token: write` perm to protect secrets).

I've dropped the last (DNM) commit now that we've proven the vault action will succeed via this PR.

vitorsavian (approved):

brandond (approved):

PR #10280, opened 2026-04-22
Add prime configuration flag ADR

manuelbuil:
#### Proposed Changes ####

Adds an ADR summarizing the different taken decisions

#### Types of Changes ####

ADR

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/012-prime-flag-and-ingress-nginx.md
+++ b/docs/adrs/012-prime-flag-and-ingress-nginx.md
@@ -1,0 +1,78 @@
+# Introduction of "prime" configuration flag
+
+Date: 2026-04-21
+
+## Status
+
+Proposed
+
+## Context
+
+On the 12th of November, a [blog post](https://www.kubernetes.dev/blog/2025/11/12/ingress-nginx-retirement/) announced that ingress-nginx will be discontinued after March 2026. This happened during Kubecon EU and ingress-nginx will not publish any extra release.
+
+SUSE announced that RKE2 will still patch ingress-nginx for 8+ CVEs and provide support for its customers during an extended period for v1.32, v1.33, v1.34, v1.35 and v1.36 minor releases, so that they have plenty of time to migrate to Traefik. This includes the v1.35 LTS, whose support ends in December 2027.
+
+Community users will get the latest ingress-nginx image published upstream but it will never be updated.
+
+We expect both customers and community users to migrate to Traefik during the v1.36 lifecycle. Customers may get an extended opportunity in v1.37 but this is not guaranteed at the moment. 
+
+## Problem
+
+RKE2 currently lacks a mechanism to differentiate between community and customer distributions at the configuration level. There is no automated way to toggle between upstream images and SUSE-maintained images within the internal Helm charts.
+
+## Proposal
+
+We will introduce a new boolean configuration flag, prime, within the RKE2 configuration. The default value will be false. While the current focus is on ingress-nginx extended life for customers, the flag establishes a pattern for future functional deviations between community and customer distributions. As of today the following changes would occur when the flag is enabled (prime: true): 
+
+1 - Registry Redirection: The global.systemDefaultRegistry is automatically pointed to the official SUSE registry for customers
+2 - Image Tag Logic: The ingress-nginx Helm chart will switch from the standard hardened tag to the prime tag.
+3 - Ingress-nginx support in RKE2 v1.37: Ingress-nginx helm chart will remain present
+
+### ingress-nginx helm chart changes
+
+The Github Project where ingress-nginx is built for RKE2 will have two types of tags:
+* v1.14.5-hardenedN => for community, it will never be updated *
+* v1.14.5-primeN => for customers. N is a natural number and it will be incremented when new patches are added.
+
+The ingress-nginx helm chart includes two fields for the image tag:
+* [tag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L859) is for community and hence will always be v1.14.5-hardenedN
+* [primeTag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L860) is for customers and hence will be v1.14.5-primeN
+
+When enabling the "prime" flag, primeTag will be set following this [logic](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/templates/_helpers.tpl#L273-L275)
+
+
+### Default ingress controller
+
+Default ingress controller will be Traefik regardless of the "prime" flag value.
+
+Clusters upgrading to RKE2 v1.36 minor will retain their existing ingress controller, even if one was not explicitly selected. This ensures community users do not experience an unexpected swap from ingress-nginx to traefik during the minor version upgrade.
+
+### Deprecation schedule for ingress-nginx
+
+v1.36: Deprecation announcement, default ingress controller changed to traefik
+
+v1.37: The ingress-nginx helm chart will be removed for community users; customers will have very limited support
+
+v1.38: The ingress-nginx helm chart will be fully removed for all users, including customers.
+
+
+### ingress-nginx and traefik in airgap 
+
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
Thanks @manuelbuil

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+
+We will introduce a new boolean configuration flag, prime, within the RKE2 configuration. The default value will be false. While the current focus is on ingress-nginx extended life for customers, the flag establishes a pattern for future functional deviations between community and customer distributions. As of today the following changes would occur when the flag is enabled (prime: true): 
+
+1 - Registry Redirection: The global.systemDefaultRegistry is automatically pointed to the official SUSE registry for customers
+2 - Image Tag Logic: The ingress-nginx Helm chart will switch from the standard hardened tag to the prime tag.
+
+### ingress-nginx helm chart changes
+
+The Github Project where ingress-nginx is built for RKE2 will have two types of tags:
+* v1.14.5-hardened1 => for community, it will never be updated
```
mdrahman-suse: FYI we already have a `hardened2` published in GH and prime ribs

brandond: I believe this is just an example to show `-hardened` vs `-prime`, it doesn't need to be the latest. But perhaps we do want to say something less strong than "never updated" since that is clearly already untrue.

caroline-suse-rancher: This was my thought too. We say hardened1 and it will never be updated vs primeN, which will. But even now, that's not the case, and we can't ensure there won't be further hardened builds - at least for the time being. The goal will be to halt hardened builds when we can. Not sure how best to capture that here. Maybe this is fine.

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+* v1.14.5-primeN => for customers. N is a natural number and it will be incremented when new patches are added.
+
+The ingress-nginx helm chart includes two fields for the image tag:
+* [tag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L859) is for community and hence will always be v1.14.5-hardened1
```
caroline-suse-rancher: Same here - "hence will always be" is not the case.

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+* [primeTag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L860) is for customers and hence will be v1.14.5-primeN
+
+When enabling the "prime" flag, primeTag will be set following this [logic](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/templates/_helpers.tpl#L273-L275)
+
+## Use cases
+
+### RKE2 Standalone
+
+Community users can switch to customers by adding "prime: true" in their config and restarting the rke2-server in the control-plane nodes. That will trigger a reinstallation of the system charts via a "helm upgrade job" which will set the suse image registry for customers and change the ingress-nginx image if necessary.
```
caroline-suse-rancher: Maybe we should add a note here of how community users can become Prime customers. As community users will not have access to the prime ingress-nginx chart _unless_ they are customers.

docs/adrs/012-prime-flag-and-ingress-nginx.md
manuelbuil: yes, I should modify that to adjust to the reality :)

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,56 @@
...
+
+When enabling the "prime" flag, primeTag will be set following this [logic](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/templates/_helpers.tpl#L273-L275)
+
+## Use cases
+
+### RKE2 Standalone
+
+Community users can switch to customers by adding "prime: true" in their config and restarting the rke2-server in the control-plane nodes. That will trigger a reinstallation of the system charts via a "helm upgrade job" which will set the suse image registry for customers and change the ingress-nginx image if necessary. 
+
+Current customers should be already setting the suse image registry in their configuration. They'd need to swap that configuration with "prime: true" and restart the rke2-server in the control-plane nodes. That will trigger a reinstallation of the system charts via a "helm upgrade job" but as there will be no change, nothing will happen.
```
caroline-suse-rancher, line 70: We should also probably note that, for Prime customers on v1.36, they will stay on ingress-nginx standalone, unless they change the ingress in their config. Also, for community, this only applies to new clusters. Since upgrades will stay on whatever ingress they have previously configured.

mdrahman-suse:
Should we also be covering how it would work in terms of upgrade?

brandond:
~I believe we also need to include the requirement from Product that Prime also set ingress-nginx as default :(~

Edit: Got this requirement removed!

manuelbuil:
I added the "default ingress controller switch in RKE2 v1.36" part and the "ingress-nginx support in RKE2 v1.37" part as well

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,73 @@
...
+* v1.14.5-hardened1 => for community, it will never be updated *
+* v1.14.5-primeN => for customers. N is a natural number and it will be incremented when new patches are added.
+
+The ingress-nginx helm chart includes two fields for the image tag:
+* [tag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L859) is for community and hence will always be v1.14.5-hardened1
+* [primeTag](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/values.yaml#L860) is for customers and hence will be v1.14.5-primeN
+
+When enabling the "prime" flag, primeTag will be set following this [logic](https://github.com/rancher/rke2-charts/blob/main/charts/rke2-ingress-nginx/rke2-ingress-nginx/4.14.504/templates/_helpers.tpl#L273-L275)
+
+* v1.14.5-hardened1 was already updated to v1.14.5-hardened2 for unrelated reasons
```
brandond: We say hardened1 and then immediately walk that back in the next paragraph. Can we just say `-hardenedN` and that it will not be updated as part of our regular release process?

docs/adrs/012-prime-flag-and-ingress-nginx.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Default ingress controller
+
+Default ingress controller will be Traefik regardless of the "prime" flag value.
+
+Clusters upgrading to RKE2 v1.36 minor will retain their existing ingress controller, even if one was not explicitly selected. This ensures community users do not experience an unexpected swap from ingress-nginx to traefik during the minor version upgrade.
+
+### Deprecation schedule for ingress-nginx
+
+v1.36: Deprecation announcement, default ingress controller swap
```
brandond:
```suggestion
v1.36: Deprecation announcement, default ingress controller changed to traefik
```

docs/adrs/012-prime-flag-and-ingress-nginx.md
manuelbuil: yeah, that's a good point

brandond (approved):

docs/adrs/012-prime-flag-and-ingress-nginx.md
caroline-suse-rancher: Above is no longer true, other than the point about upgrades :)

caroline-suse-rancher (approved):

docs/adrs/012-prime-flag-and-ingress-nginx.md
manuelbuil: good point

docs/adrs/012-prime-flag-and-ingress-nginx.md
manuelbuil: I wrote the following in the "Default ingress controller" section

```
### Default ingress controller

Default ingress controller will be Traefik regardless of the "prime" flag value.

Clusters upgrading to RKE2 v1.36 minor will retain their existing ingress controller, even if one was not explicitly selected. This ensures community users do not experience an unexpected swap from ingress-nginx to traefik during the minor version upgrade.
```

caroline-suse-rancher (approved):
Looks great!

Issue #10424, opened 2026-05-19
Support startup probe for etcd static pod


alexander-demicev:
**Describe the bug**

Etcd static pod is created without a kubelet startup probe. On clusters with a multi GB etcd datastore the in-memory index
rebuild can take longer than the about 90 s liveness budget. Without a startup probe to defer liveness, the kubelet kills etcd
mid-rebuild, it restarts, and the cluster never comes up. Logs:

```
level=info msg="Failed to test etcd connection: failed to get etcd status: context deadline exceeded"
Liveness probe failed: Get "http://127.0.0.1:2381/health?serializable=true": dial tcp: connection refused
Container etcd failed liveness probe, will be restarted
```

`--control-plane-probe-configuration etcd-startup-failure-threshold=N` is a no-op for etcd today because the startup slot isn't wired up.

**Solution**

Add startup values to etcd config. Same target as the existing liveness probe.  Startup probes are one shot and liveness with its 90s ceiling takes over after they pass.

aganesh-suse:
Closing based on: https://github.com/rancher/rke2/issues/10473

PR #10425, opened 2026-05-19
Set startup probes for etcd

alexander-demicev:
#### Proposed Changes ####

Add StartupPort, StartupPath, StartupSchemeSame target to etcd pod config, same as the existing liveness probe.
 
`etcd-startup-*` keys are already parsed by rke2 but have no effect for etcd today. With this change they become functional:
- `etcd-startup-initial-delay-seconds`
- `etcd-startup-timeout-seconds`
- `etcd-startup-failure-threshold`
- `etcd-startup-period-seconds`

#### Verification ####

Start a server with all four `etcd-startup-*` keys overridden, for
example via `/etc/rancher/rke2/config.yaml`:

```yaml
control-plane-probe-configuration:
  - etcd-startup-failure-threshold=60
  - etcd-startup-initial-delay-seconds=20
  - etcd-startup-timeout-seconds=25
  - etcd-startup-period-seconds=15
```

After rke2 starts, inspect `/var/lib/rancher/rke2/agent/pod-manifests/etcd.yaml`, the
startup probe should reflect every override:

```yaml
startupProbe:
  failureThreshold: 60
  httpGet:
    host: localhost
    path: /health?serializable=true
    port: 2381
    scheme: HTTP
  initialDelaySeconds: 20
  periodSeconds: 15
  successThreshold: 1
  timeoutSeconds: 25
```

#### Linked Issues ####

https://github.com/rancher/rke2/issues/10424

#### User-Facing Change ####

```release-note
The etcd static pod now ships with a kubelet startup probe. This makes `etcd-startup-*` keys functional.
```


brandond:
All I see is that you've populated the defaults; how does this address your reported inability to manually configure startup checks via `control-plane-probe-configuration`? I guess since there is no probe, setting the threshold and periods does nothing.

The current defaults were taken from kubeadm; if we are adding startup checks to rke2 that kubeadm does not have I would want to confirm that they do not have a good reason for omitting them.

brandond (approved):

alexander-demicev:
@brandond `control-plane-probe-configuration` values get propagated fine but we need the probe to be set. Looking at kubeadm code https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/etcd/local.go#L225, they have the startup probe but it relies on `/readyz`, I added `/health?serializable=true` for RKE2 because If I'm not mistaken the `/readyz` needs a quorum to pass.

furkatgofurov7 (approved):
LGTM

manuelbuil (approved):

brandond (approved):

alexander-demicev:
@manuelbuil @brandond can this PR be merged? I can also create back-ports if needed

manuelbuil:
> @manuelbuil @brandond can this PR be merged? I can also create back-ports if needed

yes! Let me do that now. If you could create the backports, that'd be great! You need one issue per release as well

PR #10426, opened 2026-05-19
Set klipper-helm registry correctly when prime

manuelbuil:
#### Proposed Changes ####

When setting `prime: true` all RKE2 system images were correctly pointing to the prime registry, except for `klipper-helm`. This PR fixes that and `klipper-helm` is now pointing to the prime registry as well.

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy rke2 with `prime: true` and without setting the `system-default-registry` configuration flag. Verify that the `helm-install-*` images are now using an image coming from the prime registry. 

#### Linked Issues ####

https://github.com/rancher/rke2/issues/10427

#### User-Facing Change ####

```release-note
When enabling prime configuration, the system-default-registry flag points to the correct susre prime registry automatically and hence there is no need to set it explicitly anymore
```

### DIFF
--- a/pkg/rke2/rke2_linux.go
+++ b/pkg/rke2/rke2_linux.go
@@ -34,6 +34,8 @@ func initExecutor(clx *cli.Context, cfg rke2cli.Config, isServer bool) (executor
 	// resolver's default registry will get updated later when bootstrapping
 	if !clx.IsSet("system-default-registry") && clx.Bool("prime") {
 		cfg.Images.SystemDefaultRegistry = images.PrimeRegistry
+		// By setting system-default-registry, we ensure that K3s sets the klipper-helm registry correctly
+		clx.Set("system-default-registry", images.PrimeRegistry)
 	} else {
 		cfg.Images.SystemDefaultRegistry = clx.String("system-default-registry")
 	}
--- a/tests/docker/prime/prime_test.go
+++ b/tests/docker/prime/prime_test.go
@@ -32,7 +32,7 @@ var _ = Describe("Prime Tests", Ordered, func() {
 			var err error
 			tc, err = docker.NewTestConfig()
 			Expect(err).NotTo(HaveOccurred())
-			tc.ServerYaml = "prime: true\n"
+			tc.ServerYaml = "prime: true\ningress-controller: ingress-nginx"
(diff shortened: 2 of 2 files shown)


brandond (approved):

pkg/rke2/rke2_linux.go
```diff
@@ -34,6 +34,9 @@ func initExecutor(clx *cli.Context, cfg rke2cli.Config, isServer bool) (executor
 	// resolver's default registry will get updated later when bootstrapping
 	if !clx.IsSet("system-default-registry") && clx.Bool("prime") {
 		cfg.Images.SystemDefaultRegistry = images.PrimeRegistry
+		// K3s populates config.Control.SystemDefaultRegistry from cmds.ServerConfig.
+		// We need it to set klipper-helm registry correctly
+		cmds.ServerConfig.SystemDefaultRegistry = images.PrimeRegistry
 	} else {
 		cfg.Images.SystemDefaultRegistry = clx.String("system-default-registry")
```
brandond, line 40: Right, this path was working because K3s was ALSO parsing the system-default-registry flag on its own, into `cmds.ServerConfig.SystemDefaultRegistry`.

Should we maybe just/also do `clx.Set("system-default-registry", images.PrimeRegistry)` in case there is anywhere else where the CLI flag is directly checked?

.github/workflows/test-suite.yaml
```diff
@@ -151,7 +151,6 @@ jobs:
         run: |
           sudo modprobe br_netfilter overlay
       - name: "Read secrets"
-        if: github.event_name == 'push'
```
brandond, line 154: No. PRs aren't allowed to read secrets, on purpose. This test can only run on merge (push) to a release branch.

brandond (changes requested):

.github/workflows/test-suite.yaml
manuelbuil: I was just trying to see if the test worked but there are other issues. I reverted that

pkg/rke2/rke2_linux.go
manuelbuil: yes, that also works and helps in case there are other cases. Let me push a change

brandond (approved):

vitorsavian (approved):

Issue #10360, opened 2026-05-11
rke2/multus+cilium: Workloads are scheduled too early


max06:
**Environmental Info:**
RKE2 Version: v1.32.13+rke2r1 (via rancher)

`Node(s) CPU architecture, OS, and Version: Linux srv-01-01-09.a1.redacted 6.12.85+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.85-1 (2026-04-30) x86_64 GNU/Linux`

Cluster Configuration: 3 servers, 1-9 agents

**Describe the bug:**

Creating a new rke2 cluster on rancher/elemental via rancher using Multus+Cilium schedules workloads with wrong Pod-IPs, due to nodes being marked as ready before the cilium installation finished. 

Very likely directly related to Issue #5333.

**Steps To Reproduce:**

- Installed RKE2: 
  - Target hosts are running custom rancher/elemental images based on debian 13, with podman and a lot of other packages installed, registered with the elemental operator in rancher.
  - On Cluster creation (selecting multus+cilium), rancher provisions rancher-system-agent on target nodes
  - Rancher-system-agent fetches rke2 installation instructions from rancher plan, together with bootstrap manifests and executes
  - multus and cilium are being installed by helm controller
  - multus finishes first, removes NoSchedule-taints from node
  - workload (like cattle-cluster-agent, coredns, ingress-controller) is deployed, while cilium installation is still running
  - multus assigns IP 10.88.0.x to pods (using cni-podman as bridge, not configured by us)
  - cilium installation finishes, cilium operator cleans up coredns-deployment due to unmanagedPodWatcher defaults (selector only matches kubedns)
  - all other pods scheduled before cilium finished remain unmanaged

**Expected behavior:**

The node should keep its NoSchedule-Taint until multus and cilium are finished.

**Actual behavior:**

The NoSchedule-Taint is removed after multus finished, allowing pods to be scheduled and started before cilium is ready.

**Additional context / logs:**

Issue #5333 is directly related to this. Re-enabling `setNodeTaints` taints each new node joining a cluster immediately with `node.cilium.io/agent-not-ready (NoSchedule)`, only removing the taint after the cilium installation finished on that node. I was able to confirm that all pods (except some helm-jobs, that shouldn't matter) got a Pod-IP from cilium instead of multus. 

[...]

PR #10500, opened 2026-06-01
Set bootstrap to false for rke2-multus

thomasferrandiz:
This will prevent Multus from starting before the main CNI and creating a race condition when this CNI is slow to start (e.g. Cilium).

Issue: https://github.com/rancher/rke2/issues/10360

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -29,7 +29,7 @@ charts:
     bootstrap: false
   - version: v4.2.413
     filename: /charts/rke2-multus.yaml
-    bootstrap: true
+    bootstrap: false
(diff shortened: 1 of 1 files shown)


manuelbuil (approved):
I forgot why we set it to true in the first place... maybe just because it was a CNI plugin? I am wondering if by switching this, we will encounter other issues. I can't think of any

mgfritch (approved):

thomasferrandiz:
> I forgot why we set it to true in the first place... maybe just because it was a CNI plugin? I am wondering if by switching this, we will encounter other issues. I can't think of any

I did some basic tests and everything looked fine

PR #10501, opened 2026-06-01
ADR about control-plane critical config validation

manuelbuil:
#### Proposed Changes ####

This ADR proposes a new mechanism to avoid control-plane configuration being different in HA clusters

#### Types of Changes ####

New feature

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/014-control-plane-critical-config-validation.md
+++ b/docs/adrs/014-control-plane-critical-config-validation.md
@@ -1,0 +1,89 @@
+# Control-Plane Critical Config Validation
+
+Date: 2026-06-01
+
+## Subject
+Given HA control-plane requirements, when a new RKE2 server joins an existing cluster, then we validate specific critical configuration before allowing the node to proceed.
+
+## Status
+Proposed
+
+## Context
+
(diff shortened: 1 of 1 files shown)


docs/adrs/014-control-plane-critical-config-validation.md
```diff
@@ -0,0 +1,78 @@
...
+Given HA control-plane requirements, when a new RKE2 server joins an existing cluster, then we validate specific critical configuration before allowing the node to proceed.
+
+## Status
+Proposed
+
+## Context
+
+RKE2 requires some server settings to be consistent across all control-plane nodes (for example, CNI, ingress-controller, system-default-registry (or prime)... selection). If these values diverge, cluster behavior can become non-deterministic and difficult to troubleshoot. For example, our support team has already detected this in a couple of users while migrating the ingress controller. If one control-plane sets traefik but another one sets ingress-nginx, the cluster enters a never ending loop of ingresses installing and uninstalling.
+
+K3s already has a join-time validation flow for `CriticalControlArgs` through the existing `/v1-<program>/config` endpoint and compare logic in cluster bootstrap. If that config is not the same, the new control-plane node fails to bootstrap very early with a self-explanatory error.
```
brandond, line 15: I will note that there are a bunch of caveats to the current logic that we may want to address if we are going to expand on it:
1. Config comparison requires that a `server:` address be set in the node's config.
2. The server compares its config to the server it is joining, and exits with an error if it does not match. This means that validation is entirely client-side.
3. Validation checks only the single server it is joining against; not any others.
4. Once a server has joined the cluster, the server address does not need to be set in the config as long as the node has the etcd role, so we cannot ensure that there is even a server to validate against.
5. Once a server has joined the cluster, config validation is skipped if the server cannot be contacted, to allow for cold-starts when other cluster members are not yet up.

Issue #7974, opened 2025-03-27
Asset rke2.windows-amd64.tar.gz containing both binaries. rke2.exe and rke2.


Najzer0:
As I am working on our own SBOM I've noticed that the windows tarball also includes the linux binary.
I dont think this is necc. for windows nodes and the tar.gz.

**Environmental Info:**
RKE2 Version: 
found this on 1.31.6, but looking at the current stable (.7) and filesize, this seems to be the case most of the time.
https://github.com/rancher/rke2/releases/download/v1.31.6%2Brke2r1/rke2.windows-amd64.tar.gz

Node(s) CPU architecture, OS, and Version: 
n.a.

Cluster Configuration:
n.a.

**Describe the bug:**
tar.gz bigger than required

**Steps To Reproduce:**
- Installed RKE2:  n.a.

**Expected behavior:**
.tar.gz only containing windows binaries

**Actual behavior:**
having linux + windows

**Additional context / logs:**

PR #10611, opened 2026-06-12
changed: exclude linux from windows tarball

bashilias:
#### Proposed Changes ####

Excludes Linux files from the Windows tarball (`rke2.windows-amd64.tar.gz`).

Both package-bundle and package-windows-bundle stage into the same dist/bundle directory. The Linux tarball already excludes Windows files (.exe and .ps), but the Windows tarball had no excludes.

#### Types of Changes ####

Bugfix

#### Verification ####

1. run a full package build (/scripts/package.
2. check the Windows tarball; tar -tzvf dist/artifacts/rke2.windows-amd64.tar.gz

#### Linked Issues ####
#7974 

#### User-Facing Change ####
```release-note
The Windows tarball (rke2.windows-amd64.tar.gz) no longer includes the Linux rke2 binary and other Linux only files

### DIFF
--- a/scripts/package-windows-bundle
+++ b/scripts/package-windows-bundle
@@ -25,5 +25,9 @@ mkdir -p dist/artifacts
 
 ### (make the tarball)
 if [ -z "${PACKAGE_SKIP_TARBALL}" ]; then
-    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle --exclude bin/${PROG} --exclude 'bin/*.sh' --exclude lib --exclude share/${PROG}/${PROG}-cis-sysctl.conf $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
+    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle \
+        bin/${PROG}.exe \
+        bin/rke2-uninstall.ps1 \
+        share/${PROG}/LICENSE.txt \
+        share/${PROG}-windows
 fi
--- a/scripts/package-windows-bundle
+++ b/scripts/package-windows-bundle
@@ -25,5 +25,5 @@ mkdir -p dist/artifacts
 
 ### (make the tarball)
 if [ -z "${PACKAGE_SKIP_TARBALL}" ]; then
-    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle --exclude bin/${PROG} --exclude '*.sh' --exclude lib --exclude share/${PROG}/${PROG}-cis-sysctl.conf $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bashilias** [comment]
Hi @caroline-suse-rancher I don't think anyone is being notified.

**Comment by caroline-suse-rancher** [comment]
Hey @bashilias, thanks for the contribution! I will get some eyes on it. Just be aware, we are in code freeze currently, so while we may review it, we will not merge until that is lifted.

File: scripts/package-windows-bundle [hunk 1]
```diff
@@ -25,5 +25,5 @@ mkdir -p dist/artifacts
 
 ### (make the tarball)
 if [ -z "${PACKAGE_SKIP_TARBALL}" ]; then
-    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
+    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle --exclude bin/${PROG} --exclude '*.sh' --exclude lib --exclude share/${PROG}/${PROG}-cis-sysctl.conf $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
```
manuelbuil: rke2-killall.sh and rke2-uninstall.sh are part of the bin/ directory. Shouldn't it be " --exclude='bin/*.sh'"?

manuelbuil (changes requested):

scripts/package-windows-bundle
bashilias: Thanks for the review, `--exclude '*.sh' ` without a slash matches against the basename in gnu tar, so it correctly excludes bin/rke2-killall.sh and bin/rke2-uninstall.sh regardless of their path. But ` --exclude='bin/*.sh'` is more explicit.

manuelbuil (approved):

scripts/package-windows-bundle
```diff
@@ -25,5 +25,5 @@ mkdir -p dist/artifacts
 
 ### (make the tarball)
 if [ -z "${PACKAGE_SKIP_TARBALL}" ]; then
-    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
+    tar -czf dist/artifacts/${PROG}.windows-${ARCH}.tar.gz -C dist/bundle --exclude bin/${PROG} --exclude 'bin/*.sh' --exclude lib --exclude share/${PROG}/${PROG}-cis-sysctl.conf $(find dist/bundle -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
```
brandond: Rather than excluding a bunch of crap, might it be easier to replace the find (which was copied from the linux tarball script) with a specific list of things we want to include for Windows? Or just modify the find to only grab the stuff we want?

brandond (changes requested):

scripts/package-windows-bundle
bashilias: I copied the linux tar and added an exclude. But you're right including is much cleaner..

brandond (approved):

manuelbuil (approved):

PR #10641, opened 2026-06-19
fix: avoid fatal logger crash in install.ps1

### LINKED ISSUE #6852: `install.ps1` references `$suffix` before its defined.
**Environmental Info:**
RKE2 Version: 
master

Node(s) CPU architecture, OS, and Version: 
> Probably ALL

- Windows Server 2019 Standard Edition
- Intel x86/64 architecture

Cluster Configuration:

**Describe the bug:**
If you install rke2 without the `Containers` feature for windows installed you'll get the following error:

```
[ERROR] Windows feature: 'Containers' is not installed. Please run: Install-WindowsFeature -Name Containers
Confirm-WindowsFeatures : The variable '$suffix' cannot be retrieved because it has not been set.
At C:\Users\Administrator\install.ps1:670 char:1
+ Confirm-WindowsFeatures -RequiredFeatures @("Containers")
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (suffix:String) [Confirm-WindowsFeatures], RuntimeException
    + FullyQualifiedErrorId : VariableIsUndefined,Confirm-WindowsFeatures
```

This is because the Powershell script `install.ps1` uses the variable `$suffix` without initializing it [here](https://github.com/rancher/rke2/blob/master/install.ps1#L97)

**Steps To Reproduce:**
- Ensure windows feature `Container` is not present
- run `./install.ps1` 

**Expected behavior:**
Error message should print without powershell `InvalidOperation` error.

**Actual behavior:**
powershell `InvalidOperation` error.

**Additional context / logs:**
None

immanuwell:
#### Proposed Changes ####

`install.ps1` can trip over its own fatal logger before it prints the real error. This predefines `$suffix` and makes `Write-FatalLog` build a best effort download suffix straight from `PROCESSOR_ARCHITECTURE`, so early failures just log and exit, no extra nonsense.

#### Types of Changes ####

Bugfix

#### Verification ####

1. On Windows, make sure the `Containers` feature is not installed.
2. Run `./install.ps1`.
3. Before this patch, the script prints the expected `[ERROR]` line and then faceplants with `The variable '$suffix' cannot be retrieved because it has not been set.`
4. With this patch, it prints the intended `[ERROR]` and `[ALT]` lines, then exits cleanly.

#### Testing ####

Reproduced the strict mode failure and checked the fixed logger behavior with `pwsh` in `mcr.microsoft.com/powershell:lts-7.4-alpine-3.20`. I did not run a full Windows install here.

#### Linked Issues ####

Fixes #6852

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####

This also avoids calling `Get-ArchitectureInfo` from inside `Write-FatalLog`, which keeps the error path a lot less fragile.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -84,6 +84,7 @@ param (
 
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
+$suffix = ""
 
 function Write-InfoLog() {
     Write-Output "[INFO] $($args -join " ")"
@@ -100,10 +101,12 @@ function Write-DebugLog() {
 # fatal logs the given argument at fatal log level.
 function Write-FatalLog() {
     Write-Output "[ERROR] $($args -join " ")"
-    if ([string]::IsNullOrEmpty($suffix)) {
-        $archInfo = Get-ArchitectureInfo
-        $suffix = $archInfo.Suffix
-        Write-Output "[ALT] Please visit 'https://github.com/rancher/rke2/releases' directly and download the latest rke2.$suffix.tar.gz"
+    $downloadSuffix = $suffix
+    if ([string]::IsNullOrEmpty($downloadSuffix) -and $env:PROCESSOR_ARCHITECTURE) {
+        $downloadSuffix = "windows-$($env:PROCESSOR_ARCHITECTURE.ToLower())"
+    }
+    if (-not [string]::IsNullOrEmpty($downloadSuffix)) {
+        Write-Output "[ALT] Please visit 'https://github.com/rancher/rke2/releases' directly and download the latest rke2.$downloadSuffix.tar.gz"
     }
     exit 1
 }


dereknola (changes requested):
The original install does need to change, but this fix is just creating a new problem. Now the output looks like
```
PS C:\Users\vagrant> .\install.ps1                                                                              
[ERROR] Windows feature: 'Containers' is not installed. Please run: Install-WindowsFeature -Name Containers     
[ALT] Please visit 'https://github.com/rancher/rke2/releases' directly and download the latest rke2.windows-amd64.tar.gz
```

But.... the true issue is that Containers is not yet installed. The entire `[Alt]` line is a missdirection, and would confuse the user. We should just skip the alt line if suffix is empty.

PR #10646, opened 2026-06-22
[1.33] Bump HCP 0.2.12

w13915984028:
#### Linked Issues ####

Harvester issue https://github.com/harvester/harvester/issues/10068

rke2-chart is via https://github.com/rancher/rke2-charts/pull/1043, to be approved and generated

#### User-Facing Change ####

```release-note

```

There is no change on any default behaviour.

Add some new features/bug fix:

1. Support secret based cloud-config (when user configure it explicitly) https://github.com/harvester/harvester/issues/9503
2. Support configure `extraArgs` (when user configure it explicitly) https://github.com/harvester/harvester/issues/9852

1. Fix dual-stack/dual-ip(cluster is on ipv4 only, but node is booted with both ipv4, ipv6) IP reporting https://github.com/harvester/harvester/issues/7275

### DIFF
--- a/charts/chart_versions.yaml
+++ b/charts/chart_versions.yaml
@@ -39,7 +39,7 @@ charts:
   - version: 3.5.0-rancher200
     filename: /charts/rancher-vsphere-csi.yaml
     bootstrap: true
-  - version: 0.2.1100
+  - version: 0.2.1200
     filename: /charts/harvester-cloud-provider.yaml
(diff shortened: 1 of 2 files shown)


matttrach (approved):

caroline-suse-rancher:
We are currently frozen, I will come back and merge when freeze is lifted.

PR #10681, opened 2026-07-02
test: add new nightly tests for CNIs

thomasferrandiz:
#### Proposed Changes ####

This PR adds a nightly test run to validate various CNI configurations (iptables and nftables).
This should help us to officially support running rke2 in nftables-only mode.

#### Types of Changes ####

New tests

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/nightly-cni.yaml
+++ b/.github/workflows/nightly-cni.yaml
@@ -1,0 +1,107 @@
+name: Nightly CNI
+on:
+  schedule:
+    - cron: "0 3 * * 1-5"
+  workflow_dispatch: {}
+  pull_request:
+    paths:
+      - '.github/workflows/nightly-cni.yaml'
+      - 'tests/e2e/cni/**'
+
+jobs:
+  e2e:
+    name: "Nightly CNI Tests"
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        cni: [flannel, canal, calico, cilium]
+        filter: [iptables,nftables]
+        exclude:
+          - {cni: cilium, filter: nftables}
+      max-parallel: 3
+    defaults:
+      run:
+        working-directory: tests/e2e/cni
+    env:
+      INSTALL_RKE2_CHANNEL: latest
+    steps:
+    - name: "Checkout"
+      # https://github.com/actions/checkout/releases
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with: {fetch-depth: 1}
+    - name: Remove Unnecessary Tools
+      run: |
+        sudo rm -rf \
+          /home/linuxbrew \
+          /home/packer \
+          /opt/az \
+          /opt/microsoft \
+          /usr/lib/firefox \
+          /usr/lib/google-cloud-sdk \
+          /usr/local/julia* \
+          /usr/local/share/boost \
+          /usr/local/share/chromium \
+          /usr/local/share/powershell \
+          /usr/share/az* \
+          /usr/share/dotnet \
+          /usr/share/miniconda \
+          /usr/share/swift
+        df -khl
+    - name: Set up vagrant and libvirt
+      uses: ./.github/actions/vagrant-setup
+    - name: Vagrant R/W Cache
+      if: github.ref == 'refs/heads/master'
+      # https://github.com/actions/cache/releases
+      uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
+      with:
+        path: |
+            ~/.vagrant.d/boxes
+        key: vagrant-box-ubuntu-2404
+    - name: Vagrant Read Cache
+      if: github.ref != 'refs/heads/master'
+      # https://github.com/actions/cache/releases
+      uses: actions/cache/restore@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
+      with:
+        path: |
+            ~/.vagrant.d/boxes
+        key: vagrant-box-ubuntu-2404
+    - name: "Vagrant Plugin(s)"
+      env:
+        VAGRANT_RKE2_VERSION: "0.2.0"
+        VAGRANT_RKE2_CHECKSUM: "b91dbeee2e7f35d4849ff149ce3d4bdb35f1e594b5a889c48b28a2fce346277d"
+        VAGRANT_RELOAD_VERSION: "0.0.1"
+        VAGRANT_RELOAD_CHECKSUM: "aa03d5e52737586bd5dab740c8bc61690e63c75c0d96383d9bea7520fa05be6a"
+        VAGRANT_SCP_VERSION: "0.5.9"
+        VAGRANT_SCP_CHECKSUM: "e3adda6c6a059f10e9edfa90a8f08892493010f176a18360a5d92d92847cf329"
+      run: |
+        for plugin in "vagrant-rke2:${VAGRANT_RKE2_VERSION}:${VAGRANT_RKE2_CHECKSUM}" \
+                      "vagrant-reload:${VAGRANT_RELOAD_VERSION}:${VAGRANT_RELOAD_CHECKSUM}" \
+                      "vagrant-scp:${VAGRANT_SCP_VERSION}:${VAGRANT_SCP_CHECKSUM}"; do
+          name="${plugin%%:*}"
+          rest="${plugin#*:}"
+          version="${rest%%:*}"
+          checksum="${rest##*:}"
+          curl -fsSL "https://rubygems.org/gems/${name}-${version}.gem" -o "${name}.gem"
+          echo "${checksum}  ${name}.gem" | sha256sum -c
+          vagrant plugin install "${name}.gem"
+          rm -f "${name}.gem"
+        done
+    
+    - name: Install Go
+      uses: ./.github/actions/setup-go
+    - name: Install Kubectl
+      env:
+        KUBECTL_VERSION: "v1.34.6"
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly-cni.yaml
```diff
@@ -0,0 +1,108 @@
+name: Nightly CNI
```
dereknola, line 1: since its triggers on the same cron, just combine this with the nightly install workflow.

.github/workflows/nightly-cni.yaml
```diff
@@ -0,0 +1,108 @@
...
+on:
+  schedule:
+    - cron: "0 0 * * 1-5"
+  workflow_dispatch: {}
+  push:
+    paths:
+      - '.github/workflows/nightly-cni.yaml'
+      - 'tests/cni/**'
```
dereknola: Nit: I don't know if we care to test this on Push, since its going to be tested already on PR.

dereknola (changes requested):

tests/e2e/testutils.go
```diff
@@ -402,8 +402,14 @@ func CreateLocalMixedCluster(nodeOS string, serverCount, linuxAgentCount, window
 }
 
 func (config TestConfig) DeployWorkload(workload string) (string, error) {
-	resourceDir := "../resource_files"
+	var resourceDir string
+	resourceDir = "../resource_files"
 	files, err := os.ReadDir(resourceDir)
+	if err != nil {
+		resourceDir = "../e2e/resource_files"
+		fmt.Println("Unable to read resource manifest file for ", workload)
```
dereknola: All the e2e tests should be launched and have the same relative access to the resource_files directory. When are you seeing this fail?

tests/e2e/scripts/create_cni_manifest.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
dereknola, line 1: Move this file to `/tests/e2e/scripts/` to keep everything consistent

dereknola (changes requested):

tests/e2e/testutils.go
thomasferrandiz: it's because I didn't put the `cni` directory inside the `e2e` one.
Now that it's moved this is not needed

.github/workflows/nightly-cni.yaml
thomasferrandiz: I was thinking that I should maybe set it to a different cron schedule so that it doesn't overload the CI?

.github/workflows/nightly-cni.yaml
thomasferrandiz: I forgot to remove it after testing the workflow

.github/workflows/nightly-cni.yaml
```diff
@@ -0,0 +1,104 @@
...
+          - {cni: cilium, filter: nftables}
+      max-parallel: 3
+    defaults:
+      run:
+        working-directory: tests/e2e/cni
+    env:
+      INSTALL_RKE2_CHANNEL: latest
+    steps:
+    - name: "Checkout"
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
matttrach: checkout is on version 7 now: 
https://github.com/actions/checkout/commit/9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
I suggest adding a comment with a link to the releases page to make it easier:

```suggestion
      # https://github.com/actions/checkout/releases
      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```

.github/workflows/nightly-cni.yaml
```diff
@@ -0,0 +1,104 @@
...
+          /usr/share/az* \
+          /usr/share/dotnet \
+          /usr/share/miniconda \
+          /usr/share/swift
+        df -khl
+    - name: Set up vagrant and libvirt
+      uses: ./.github/actions/vagrant-setup
+    - name: Vagrant R/W Cache
+      if: github.ref == 'refs/heads/master'
+      uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5
```
matttrach:
```suggestion
      # https://github.com/actions/cache/releases
      uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
```

.github/workflows/nightly-cni.yaml
```diff
@@ -0,0 +1,104 @@
...
+      with:
+        path: |
+            ~/.vagrant.d/boxes
+        key: vagrant-box-ubuntu-2404
+    - name: Vagrant Read Cache
+      if: github.ref != 'refs/heads/master'
+      uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5
```
matttrach:
```suggestion
      # https://github.com/actions/cache/releases
      uses: actions/cache/restore@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
```

matttrach (approved):

.github/workflows/nightly-cni.yaml
brandond: it'll just queue if we are short on runners, should be fine.

tests/e2e/cni/Vagrantfile
```diff
@@ -0,0 +1,112 @@
...
+      rke2.installer_url = 'file:///home/vagrant/install.sh'
+      rke2.env = %W[ #{external_env} INSTALL_RKE2_TYPE=server]
+      rke2.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
+      rke2.config = <<~YAML
+        write-kubeconfig-mode: '0644'
+        node-external-ip: #{node_ip4},#{node_ip6}
+        node-ip: #{node_ip4},#{node_ip6}
+        token: vagrant-rke2
+        kube-proxy-arg:
+          - proxy-mode=#{proxy_mode}
```
brandond, line 80: Should we also be removing iptables from the host for nftables, and vice versa - to make sure that the CNI works when only the selected filter backend is available? Otherwise things might be falling back to iptables without us knowing about it.

thomasferrandiz: good point, thanks

dereknola (approved):

tests/e2e/cni/Vagrantfile
brandond: In addition to removing the userspace tools, might also be good to unload and blacklist the kernel modules. Otherwise privileged containers can still use them, if they have the CLI tools available within the image.

Should be able to just `echo blacklist ip_tables >> /etc/modprobe.d/blacklist-iptables.conf` and then reboot - you could also try `rmmod ip_tables` after blacklisting but getting all the various modules removed in the correct order once they're loaded might be challenging; just blacklisting and then rebooting might be simpler.

brandond (approved):

tests/e2e/cni/cni_test.go
```diff
@@ -104,12 +104,21 @@ var _ = Describe("Verify DualStack in "+*cni+", "+*filter+" configuration", Orde
...
+					calicoCmd = "nft list ruleset | grep 'cali-'"
+					flannelCmd = "nft list ruleset | grep 'flannel'"
 				} else {
-					cmd = "iptables -L | grep -E 'cali-|FLANNEL'"
+					calicoCmd = "iptables -L | grep 'cali-'"
+					flannelCmd = "iptables -L | grep 'FLANNEL'"
 				}
-
+				res, err := node.RunCmdOnNode(calicoCmd)
+				GinkgoWriter.Println("canal calico filtering backend check on", node.Name, ":", res)
```
brandond: The idiomatic way to dump progress within a gingkgo node is `By()`

https://github.com/rancher/rke2/blob/dcd092797ccf3b782e8b2269d614c4be95801aa2/tests/e2e/multus/multus_test.go#L46-L47

https://github.com/rancher/rke2/blob/dcd092797ccf3b782e8b2269d614c4be95801aa2/tests/docker/ing_migration/ing_migration_test.go#L118-L136

brandond: A bunch of the existing e2e code also just uses `fmt.Println` but ideally we would move away from that and wire a proper logger up to the test writer :/

tests/e2e/cni/cni_test.go
```diff
@@ -0,0 +1,296 @@
...
+		cmd := "kubectl get ingress ds-ingress --kubeconfig=" + tc.KubeconfigFile + " -o jsonpath=\"{.spec.rules[*].host}\""
+		hostName, err := e2e.RunCommand(cmd)
+		Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+		nodeIPs, err := e2e.GetNodeIPs(tc.KubeconfigFile)
+		Expect(err).NotTo(HaveOccurred(), "failed cmd: "+cmd)
+		for _, node := range nodeIPs {
+			cmd := fmt.Sprintf("curl  --header host:%s http://%s/name.html", hostName, node.Ipv4)
+			Eventually(func() (string, error) {
+				return e2e.RunCommand(cmd)
+			}, "60s", "5s").Should(ContainSubstring("ds-clusterip-pod"), "failed cmd: "+cmd)
```
brandond: FYI you can simplify all of these `Eventually(func() (string, error) {})` tests - Eventually passes the first return value to the matcher, and expects all other return values (ie the error) to be nil, and you can pass the args in via `WithArguments`.

```suggestion
		for _, node := range nodeIPs {
			cmd := fmt.Sprintf("curl  --header host:%s http://%s/name.html", hostName, node.Ipv4)
			Eventually(e2e.RunCommand, "60s", "5s").WithArguments(cmd).Should(ContainSubstring("ds-clusterip-pod"), "failed cmd: "+cmd)
```

thomasferrandiz:
The fix for the last failing test is available in the master branch but not yet in the latest release tested by the nightly tests.

PR #10684, opened 2026-07-06
Add updateCLI policy to bump ingress-nginx prime image tag

cwayne18:
## What

Adds an updateCLI policy that automatically opens a PR to bump `INGRESS_NGINX_PRIME_TAG` in `scripts/build-images` whenever a newer `-primeN` ingress-nginx image (a CVE rebuild pushed to the prime registry) is tagged in [`rancher/ingress-nginx`](https://github.com/rancher/ingress-nginx) for the nginx base version currently pinned here.

## Why

When a CVE is found in the ingress-nginx base image, a new release is tagged from the `hardened-nginx-1.x.x` branch and the image is pushed to the prime registry (`registry.rancher.com`) with a `-primeN` suffix. Today the corresponding bump of `INGRESS_NGINX_PRIME_TAG` is manual — this automates it.

## How

- `updatecli/scripts/get-ingress-nginx-prime.sh` (source): reads the currently pinned `INGRESS_NGINX_PRIME_TAG`, derives the nginx base version (e.g. `v1.15.1`), and returns the highest `-primeN` git tag available for that base in `rancher/ingress-nginx` (numeric, double-digit safe).
- `updatecli/updatecli.d/ingress-nginx-prime.yml` (manifest): a `file` target that updates `INGRESS_NGINX_PRIME_TAG` in `scripts/build-images` and opens the PR.

Because the base version is derived from the file itself, this only ever advances the prime (CVE rebuild) suffix — it never silently jumps the nginx minor/patch line, which remains a deliberate human change.

The new manifest is picked up automatically by the existing scheduled `.github/workflows/updatecli.yml` — no workflow changes required.

## Validation

Ran `updatecli diff` against `rancher/rke2@master`: the source resolved the latest tag and the target produced the expected change:
```
-INGRESS_NGINX_PRIME_TAG=v1.15.1-prime2
+INGRESS_NGINX_PRIME_TAG=v1.15.1-prime5
```


brandond (approved):

updatecli/scripts/get-ingress-nginx-prime.sh
```diff
@@ -0,0 +1,33 @@
...
+
+current=$(sed -nr 's/^INGRESS_NGINX_PRIME_TAG=(.*)$/\1/p' "${BUILD_IMAGES_FILE}")
+if [ -z "${current}" ]; then
+    echo "failed to read INGRESS_NGINX_PRIME_TAG from ${BUILD_IMAGES_FILE}" >&2
+    exit 1
+fi
+
+base=${current%-prime*}
+
+latest=$(git ls-remote --tags --refs "${INGRESS_NGINX_REPO}" "${base}-prime*" \
```
mgfritch: what happens if the tag has been marked as a pre-release?

updatecli/updatecli.d/ingress-nginx-prime.yml
```diff
@@ -0,0 +1,58 @@
...
+      branch: master
+
+sources:
+  prime:
+    name: "Get latest ingress-nginx prime tag"
+    kind: "shell"
+    spec:
+      command: bash ./updatecli/scripts/get-ingress-nginx-prime.sh
+      environments:
+        - name: PATH
```
mgfritch: you could also avoid exec'ing a shell script by using a updatecli [`githubrelease`](https://www.updatecli.io/docs/plugins/resource/github_release/)

mgfritch (changes requested):
also, what about the [chart version](https://github.com/rancher/rke2/blob/aea5181e0c830c8a0b0c0ec59f74e7ee335bbe71/charts/chart_versions.yaml#L18)?

cwayne18:
Thanks @mgfritch — addressed all three:

- **Pre-release tags** & **avoiding the shell script**: replaced the `get-ingress-nginx-prime.sh` source with a native updatecli `githubrelease` source using `typefilter: { release: true, draft: false, prerelease: false }`, so drafts/pre-releases are excluded and no shell exec is needed for discovery.
- **Chart version**: added a `helmchart` source plus a `yaml` target that bumps the `rke2-ingress-nginx` entry in `charts/chart_versions.yaml`. It is guarded by a `yaml` condition on the entry filename so the positional key can never edit the wrong chart if the file is reordered (it safely skips instead).

The prime image tag pattern is intentionally pinned to the current `v1.15.1` hardened line so a nginx minor/patch bump stays a deliberate human change.

updatecli/updatecli.d/ingress-nginx-prime.yml
```diff
@@ -0,0 +1,106 @@
...
+    kind: "yaml"
+    disablesourceinput: true
+    spec:
+      file: charts/chart_versions.yaml
+      # Guard the positional key below: if the entry ever moves, the condition
+      # fails and the chart-version target is safely skipped instead of editing
+      # the wrong chart.
+      key: '$.charts[5].filename'
+      value: /charts/rke2-ingress-nginx.yaml
+
```
mgfritch:
```suggestion
verifyPrimeInChartValues:
    name: "Verify prime image is contained in the source chart's values.yaml"
    kind: "helmchart"
    disablesourceinput: true
    spec:
      url: "https://rancher.github.io/rke2-charts"
      name: "rke2-ingress-nginx"
      version: '{{ source "chart" }}'
      file: "values.yaml"
      key: "primeTag"
      value: '{{ source "prime" }}'
      
```

PR checks should fail if the chart version doesn't actually contain the desired image, but add a conditional to prevent UpdateCLI from creating the PR anyway?

mgfritch (approved):
We also have [this](https://github.com/rancher/rke2/blob/aea5181e0c830c8a0b0c0ec59f74e7ee335bbe71/updatecli/scripts/update_chart_and_images_cni.sh#L37) and [this](https://github.com/rancher/rke2/blob/aea5181e0c830c8a0b0c0ec59f74e7ee335bbe71/updatecli/scripts/update_chart_and_images.sh#L35) for extracting the image version(s) from the chart. 

At some point it would be nice to consolidate on a general pattern for ensuring the chart/image bump PRs remain in sync.

PR #10696, opened 2026-07-07
Add PRIME versions of the vsphere images

dereknola:
#### Proposed Changes ####
- Update to latest CPI/CSI charts which now have PRIME only hardened versions. 

#### Verification ####
- Deploy rke2 on vsphere with the `cloud-controller-name: vsphere` and `prime: true`
- See that the images deployed for the csi and cpi verisons are `hardened-XXXXX` instead of the `mirrored-sig-storage-XXXXX`

#### Testing ####
We have no CI that tests vsphere, so this will need to be handled by QA

#### Linked Issues ####
https://github.com/rancher/rke2/issues/10718

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/scripts/build-images
+++ b/scripts/build-images
@@ -118,8 +118,8 @@ EOF
     else
 xargs -n1 -t $PULL_CMD << EOF > $BUILD_DIR/images-vsphere.txt
     ${REGISTRY}/rancher/mirrored-cloud-provider-vsphere:v1.36.0
-    ${REGISTRY}/rancher/mirrored-cloud-provider-vsphere-csi-release-driver:v3.6.0
-    ${REGISTRY}/rancher/mirrored-cloud-provider-vsphere-csi-release-syncer:v3.6.0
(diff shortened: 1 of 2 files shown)


cwayne18 (approved):

scripts/build-images
```diff
@@ -113,6 +127,8 @@ xargs -n1 -t $PULL_CMD << EOF > $BUILD_DIR/images-vsphere.txt
     ${REGISTRY}/rancher/mirrored-sig-storage-csi-provisioner:v4.0.1
     ${REGISTRY}/rancher/hardened-csi-snapshotter:v8.6.0-build20260608
```
brandond, line 128: Doesn't this new chart release bump the non-prime versions of these as well? Or is that still WIP?

dereknola: Good catch Brad

cwayne18 (approved):

brandond (approved):

PR #10810, opened 2026-07-15
Fix busy-wait loop in HNS network creation polling

cwayne18:
## Problem

During Windows node network setup, the loop that waits for the Calico HNS network to become available after creation had no sleep between iterations. This caused a tight busy-wait that burns 100% CPU for up to 5 minutes while polling `hcsshim.GetHNSNetworkByName`.

## Fix

Add a 1-second `time.Sleep` between polling attempts, converting the busy-wait into a reasonable polling loop.

## Impact

Eliminates a CPU spike on Windows nodes during Calico network initialization. The network is still detected within 1 second of becoming available, so there is no meaningful delay added to the setup process.

### DIFF
--- a/developer-docs/upgrading_kubernetes.md
+++ b/developer-docs/upgrading_kubernetes.md
@@ -38,6 +38,13 @@ This will take a few minutes for [CI](https://drone-pr.rancher.io/rancher/image-
 
 ## Update RKE2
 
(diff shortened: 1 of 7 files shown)


brandond:
There seems to be some other stuff mixed in to this PR?

PR #10907, opened 2026-07-22
updatecli: gate vsphere PRs on the chart version needing an update

mgfritch:
The vsphere-cpi/vsphere-csi pipelines had no conditions block, so their targets always ran. When only the scripts/build-images file target changed (e.g. a mirrored image tag drifted) but charts/chart_versions.yaml did not, updatecli still opened a PR with no chart version bump (see #10906).

Add a shell condition that runs validate_version.sh, mirroring the CNI pipelines. It compares the resolved chart source version against the version pinned in charts/chart_versions.yaml and exits non-zero ("has the latest version") when they match. A failing condition makes updatecli skip the targets and open no PR, so a pull request is only created when the chart version in charts/chart_versions.yaml would actually change.

Verified with mikefarah yq v4.53.3: validate_version.sh exits 1 for the current version (rancher-vsphere-csi 3.7.2-rancher300, rancher-vsphere-cpi 1.15.200) and exits 0 for a newer version, so no PR is created unless the chart version bumps.

#### User-Facing Change ####

```release-note

```


mgfritch:
Disallows creation of PRs that only bump images, but do not change the chart

For example: https://github.com/rancher/rke2/pull/10905

brandond (approved):

vitorsavian (approved):

PR #10921, opened 2026-07-23
Add ADR for helm v4 SSA failure handling

brandond:
#### Proposed Changes ####

Add ADR for helm v4 SSA failure handling

Covers changes from:
* https://github.com/rancher/rke2/pull/10890
* https://github.com/k3s-io/helm-controller/pull/341

#### Types of Changes ####

Docs

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/docs/adrs/014-helmchart-failureaction-change.md
+++ b/docs/adrs/014-helmchart-failureaction-change.md
@@ -1,0 +1,57 @@
+# Change Default FailureAction to `retry` for Bundled Charts
+
+Date: 2026-07-23
+
+## Status
+
+Accepted
+
+## Context
+
+RKE2 includes a bundled Helm controller that uses a container image that bundles the helm CLI tools to install, upgrade, and uninstall Helm charts.
+
+The image used by the Helm controller recently moved from bundling v3 releases of the Helm CLI, to v4. This was due to Helm v3 reaching end-of-life status, which made it no longer eligible for security updates.
+
+One of the changes included in Helm v4 is to use [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management) to effect changes to resources managed by Helm.
+One side-effect of this is that Helm now respects field ownership, and any change made to field set by Helm, cannot be changed by another entity, without causing conflicts that move the Chart release into a `failed` state.
+
+#### Example 1:
+1. A Chart's template and values specify a Deployment with 1 replica
+2. The user manually scales the Deployment (ex: `kubectl scale deployment EXAMPLE-DEPLOYMENT --replicas=3`)
+3. The user makes a change to the chart version or values, triggering a `helm upgrade` operation
+4. The upgrade fails and creates a new release version with `failed` status, due to `kubectl scale` now owning the Deployment's `.spec.replicas` field.
+
+#### Example 2:
+1. A Chart deploys a Custom Resource, and a Operator that manages that custom resource (ex: Calico's `Installation` resource)
+2. The Operator runs, and updates the Custom Resource to modify fields based on the running state of the cluster.
+3. The user makes a change to the chart version or values that do not affect the Custom Resource, triggering a `helm upgrade` operation.
+4. The upgrade fails and creates a new release version with `failed` status, due to the operator now owning some of the fields on the Custom Resource.
+
+Historically, the only things that would cause the chart release to move into a `failed` state were incompatible changes to Kubernetes resources that required delete and recreate of a resource, or unexpected early termination of a Helm upgrade Pod.
+With the new behavior of Helm v4, failure is much more likely.
+
+When the Chart moves into a `failed` state, the default failure policy is `reinstall`, which uninstalls and reinstalls the chart to restore it to a known-good state.
+While this was effective in ensuring that charts were consistently installed and upgraded, there were scenarios where this could cause unexpected outages or data loss.
+For this reason, many users have deployed HelmChartConfigs that override the failure policy to `abort`, which simply leaves the chart in a failed state, awaiting external action from the cluster administrator to resolve the issue.
+
+The Helm controller recently added a new field (`.spec.forceConflicts`) and a new failure policy (`.spec.failurePolicy: retry`) to allow Helm to retry the operation and reclaim management of the conflicting fields.
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


docs/adrs/014-helmchart-failureaction-change.md
```diff
@@ -0,0 +1,57 @@
...
+3. The user makes a change to the chart version or values that do not affect the Custom Resource, triggering a `helm upgrade` operation.
+4. The upgrade fails and creates a new release version with `failed` status, due to the operator now owning some of the fields on the Custom Resource.
+
+Historically, the only things that would cause the chart release to move into a `failed` state were incompatible changes to Kubernetes resources that required delete and recreate of a resource, or unexpected early termination of a Helm upgrade Pod.
+With the new behavior of Helm v4, failure is much more likely.
+
+When the Chart moves into a `failed` state, the default failure policy is `reinstall`, which uninstalls and reinstalls the chart to restore it to a known-good state.
+While this was effective in ensuring that charts were consistently installed and upgraded, there were scenarios where this could cause unexpected outages or data loss.
+For this reason, many users have deployed HelmChartConfigs that override the failure policy to `abort`, which simply leaves the chart in a failed state, awaiting external action from the cluster administrator to resolve the issue.
```
manuelbuil, line 35: This concerns me a bit, though I might be missing some historical context. I consider the `reinstall` policy too intrusive and dangerous because of the uninstallation step. However, since it has been around almost since the start of K3s, there must have been good reasons for picking it. My main concern is that by changing a long-standing default, we might uncover or recreate the problems reinstall was originally designed to solve.

I'd love to understand if things have changed over time making it no longer necessary. For example, maybe a few years ago Helm charts frequently ended up in an unrecoverable state that required a full reinstallation, but chart quality has improved and this is now rare. You were probably not around either when this decision was taken but perhaps you know?

docs/adrs/014-helmchart-failureaction-change.md
```diff
@@ -0,0 +1,57 @@
...
+
+The Helm controller recently added a new field (`.spec.forceConflicts`) and a new failure policy (`.spec.failurePolicy: retry`) to allow Helm to retry the operation and reclaim management of the conflicting fields.
+
+## Decision
+
+1. Helm charts bundled with RKE2 will set `.spec.forceConflicts: true` to restore the previous behavior of ignoring field management conflicts.
```
manuelbuil, line 41: Even though internals change, from a user perspective, it feels like setting forceConflicts to true makes Helm v4 behave like Helm v3, right? Or is there any potential problem we should keep an eye on? I'm thinking on some kind of operator vs helm infinite conflict, but I guess should have already seen that with v3

docs/adrs/014-helmchart-failureaction-change.md
```diff
@@ -0,0 +1,57 @@
...
+If the user has overridden the policy to `abort`, the Helm controller will not retry the upgrade with `--force-conflicts` set, nor will it uninstall and reinstall.
+The administrator will need to manually correct the failure before the chart can be upgraded or new values applied.
+
+Any HelmChart resources deployed by users will continue to see the legacy behavior of the default failure policy and Helm v4 field management conflicts.
+
+## Consequences
+
+1. Helm charts bundled with RKE2 will ignore conflicts in server-side apply field management, by use of Helm v4's `--force-conflicts` flag. This re-aligns with the default behavior of prior releases of RKE2 that used Helm v3.
+2. Helm charts bundled with RKE2 will retry the upgrade when the latest helm release is `failed` state, or was moved to `failed` state due to being left in a `pending` state due to interrupted upgrade operation.
+   Retry operations will be visible to users by continuous creation of Events reporting the attempt, attached to the HelmChart resource, emitted by helm-controller. Users will also observe the Helm Job and Pods being recreated to retry the upgrade.
```
manuelbuil, line 56: We are not capping this, right? It will be retried forever until the cluster admin fixes the problem

manuelbuil (commented):
Thanks for the ADR!

docs/adrs/014-helmchart-failureaction-change.md
brandond: Up until I added `abort` there was no such things as policies. If the chart failed, we uninstalled and reinstalled it. This behavior goes back to the very first releases of helm-controller and klipper-helm that Darren wrote years ago. At the time I don't think he was very concerned about the side effects of this, as the only thing we were using it for was deploying Traefik in K3s. He just wanted the chart to get updated eventually and didn't really care how that happened.

With RKE2 we started using it a lot more, and using it to deploy charts that do not behave well when uninstalled and reinstalled - for example, if a chart includes CRDs, then we'll uninstall the CRDs which results in all resources of that type getting deleted. Uninstalling the CNI chart may leave the cluster network broken. And so on. That lead to the `abort` policy which let us somewhat punt on the problem, by just refusing to deal with it in favor of letting the administrator figure it out.

Previously it took some unusual interruption to the update process, or a conflicting change that the apiserver does not allow to put the release into failed state. Now that helm v4 has added SSA support we have a MUCH more common case for releases to go into the failed state - any conflict in field ownership will do it. Now that it's more common, we need to actually handle it in a reasonable, automated fashion. That is what `retry` attempts to do.

docs/adrs/014-helmchart-failureaction-change.md
brandond: SSA with force-conflicts, or just disabling SSA entirely, makes it behave much like helm v4 yes.

That said I don't want to turn it off by default, because when used correctly it protects against users trying to reconfigure things out from under Helm, and then getting surprised when Helm undoes their changes next time they upgrade.

docs/adrs/014-helmchart-failureaction-change.md
brandond: That is the current design, correct. The job RetryPolicy has always capped failed container restarts at 1000 which is effectively forever as well, so the churn here is not new.

docs/adrs/014-helmchart-failureaction-change.md
manuelbuil: thanks for the explanation Brad

manuelbuil (approved):

brandond:
Merging, as this ship has sailed.

PR #10936, opened 2026-07-24
e2e: use server node IP for Calico eBPF kubernetes endpoint

thomasferrandiz:
#### Proposed Changes ####

The Calico eBPF e2e test manifest hardcoded `localhost` as the `kubernetesServiceEndpoint.host`, which is incorrect when Calico eBPF is configured to bypass kube-proxy and connect to the Kubernetes API server directly. The endpoint should be the actual IP address of the server node so that Calico can reach the API server over the network rather than loopback.

This makes the e2e test match our documentation on using Calico with eBPF.

The node's IPv4 address is passed as an argument from the Vagrantfile provisioner to the manifest script, and used in place of `localhost`.

#### Types of Changes ####

Bug fix in e2e test infrastructure.

#### Verification ####

Run the Calico eBPF e2e test suite:
```
cd tests/e2e/calico_ebpf
go test -v -timeout=60m ./... -nodeOS=bento/ubuntu-24.04 -serverCount=1 -agentCount=1
```

Verify that `rke2-calico-config.yaml` on the server node contains `host: 10.10.10.100` instead of `host: localhost`.

#### Testing ####

Covered by the existing Calico eBPF e2e test suite (`tests/e2e/calico_ebpf/calico_ebpf_test.go`).

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

The Vagrantfile already computes each node's IP as `#{NETWORK4_PREFIX}.#{100+node_num}`, so server-0 gets `10.10.10.100`. The same argument-passing pattern is already used by the `configure_second_interface.sh` provisioner in the same Vagrantfile.

### DIFF
--- a/tests/e2e/scripts/calico_ebpf_manifest.sh
+++ b/tests/e2e/scripts/calico_ebpf_manifest.sh
@@ -3,6 +3,8 @@
 # Set Calico parameters to use the eBPF dataplane instead of iptables
 mkdir -p /var/lib/rancher/rke2/server/manifests
 
+NODE_IP=${1}
+
 echo "Creating calico chart"
 echo "apiVersion: helm.cattle.io/v1
 kind: HelmChartConfig
@@ -20,4 +22,4 @@ spec:
         kubeProxyManagement: Enabled
         linuxDataplane: BPF
     kubernetesServiceEndpoint:
-      host: localhost" > /var/lib/rancher/rke2/server/manifests/rke2-calico-config.yaml
+      host: ${NODE_IP}" > /var/lib/rancher/rke2/server/manifests/rke2-calico-config.yaml
--- a/tests/e2e/calico_ebpf/Vagrantfile
+++ b/tests/e2e/calico_ebpf/Vagrantfile
@@ -32,7 +32,7 @@ def provision(vm, roles, role_num, node_num)
   
   scripts_location = Dir.exist?("./scripts") ? "./scripts" : "../scripts" 
   vm.provision "Configure second interface", type: "shell", path: scripts_location + "/configure_second_interface.sh", args: [ node_ip4, node_ip6, node_ip6_gw, vm.box ]
-  vm.provision "Create Calico Manifest", type: "shell", path: scripts_location +  "/calico_ebpf_manifest.sh"
+  vm.provision "Create Calico Manifest", type: "shell", path: scripts_location +  "/calico_ebpf_manifest.sh", args: [ node_ip4 ]
 
   install_type = getInstallType(vm, RELEASE_VERSION, GITHUB_BRANCH)
   


tests/e2e/scripts/calico_ebpf_manifest.sh
```diff
@@ -20,4 +22,4 @@ spec:
         kubeProxyManagement: Enabled
         linuxDataplane: BPF
     kubernetesServiceEndpoint:
-      host: localhost" > /var/lib/rancher/rke2/server/manifests/rke2-calico-config.yaml
+      host: ${NODE_IP}" > /var/lib/rancher/rke2/server/manifests/rke2-calico-config.yaml
```
brandond, line 25: Do we really tell people to hardcode the IP of a single node in the chart values? That is terrible, we must not do that. Won't this break the calico for the whole cluster if that one node is down?

Why can't we use `localhost:6443`? This is guaranteed to be available on both servers and agents - if the node is not a control-plane node, there's a supervisor load-balancer listening on that port that forwards to an apiserver. This is build in to RKE2 and does not rely on kube-proxy.

brandond (changes requested):

tests/e2e/scripts/calico_ebpf_manifest.sh
thomasferrandiz: We've had reports from customers (SURE-11652 for example) that using `localhost` doesn't work in that case.
The Calico pod `calico-kube-controllers` is not using the host network so it doesn't resolve `localhost:6643` as the load-balancer endpoint.

That's also why I added the doc about using HAProxy for when users need HA with Calico eBPF.

brandond: Hmm. These are all ugly solutions that we should come up with a better answer for.

Can we modify calico-kube-controllers so that it does run with host network? Or does it need to be in the cluster for some reason?

If running with host network is not desirable for some reason, we should use [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#available-fields) to inject the IP of the node the pod is running on (`status.hostIP`), instead of hardcoding a single IP in the values.

thomasferrandiz: We can't easily modify calico-kube-controllers because it's deployed by the Tigera Operator so we don't control the daemonset configuration.
I'll check if I can find a way to use `status.hostIP` in the KubernetesEndPoint configmap.

thomasferrandiz: It looks like we can't do that either without changing the operator itself.
We could always open an issue upstream but there was https://github.com/projectcalico/calico/issues/9141 already.

Their opinion seems to be that it's OK to have a non-redundant connection to the API server because the downtime would be short anyway.

Maybe kube-vip would be a better fit for this than Haproxy?

brandond: This seems like a really poor design in general... either you need to make one of your nodes a single point of failure for the whole cluster, or you need to stand up your own LB in front of the apiserver?

If it would just look at the Kubernetes Endpoints/EndpointSlice instead of the service ClusterIP, it would get the IPs of all the control-plane nodes, and could connect to them directly without having to do anything at all.

thomasferrandiz: I agree and we could try to submit a PR upstream but it doesn't look like they care.
eBPF is most likely there to check a box compared to Cilium it's not their main data plane.

PR #10937, opened 2026-07-24
Add cluster load balancer e2e test to nightly CNI workflow

thomasferrandiz:
#### Proposed Changes ####

Adds an e2e test that reproduces the external load balancer topology documented at https://docs.rke2.io/networking/cluster-loadbalancer. That setup is an HA embedded-etcd cluster fronted by an HAProxy + Keepalived virtual IP (VIP), where the VIP serves as the fixed registration address (`9345`) and API server endpoint (`6443`). Servers beyond the first, and all agents, join the cluster through the VIP rather than a concrete server IP. Until now nothing in CI exercised this registration-through-a-load-balancer path.

The new `tests/e2e/clusterloadbalancer` suite stands up 1 LB + 3 HA servers + 1 agent (reduced from the doc's 8-node example to fit the GitHub runner). Key design points:

- **Ordered bring-up.** The load balancer comes up first so the VIP exists, then the cluster-init server, then the remaining servers and agent register through the VIP. Nodes are provisioned sequentially so each joins fully before the next starts, avoiding etcd learner races.
- **VIP addressing.** Nodes start at `.101` so the VIP `.100` never collides with a node's primary IP. `tls-san` includes the VIP so the API server cert is valid for it.
- **HAProxy + Keepalived** are configured by a new `scripts/loadbalancer_setup.sh` (TCP backends for all servers on 9345/6443, a single MASTER keepalived instance for the VIP).
- **eBPF variant.** `scripts/calico_ebpf_manifest.sh` gained an optional `kubernetesServiceEndpoint.host` argument (defaulting to `localhost`, so the existing `calico_ebpf` test is unaffected) so the eBPF dataplane run points at the VIP with kube-proxy disabled, matching the doc's eBPF section.

The suite verifies HAProxy/Keepalived are active with the VIP assigned, that joining nodes reference the VIP in their config, that the cluster becomes Ready, that the API server is reachable through a VIP-based kubeconfig on 6443, a basic ClusterIP workload, and (eBPF only) that no `KUBE-SVC` iptables rules remain.

No product code is changed; this is test-only plus a nightly workflow addition. It does not require a documentation update.

#### Types of Changes ####

New test / CI coverage.

#### Verification ####

The full run requires libvirt + Vagrant and is exercised by the new `e2e-loadbalancer` job in the nightly CNI workflow (matrix: `canal`/iptables and `calico`/ebpf). Static checks performed locally: `gofmt` clean, `go vet ./tests/e2e/clusterloadbalancer/` passes, `bash -n` on both scripts, `ruby -c` on the Vagrantfile, and YAML load of the workflow.

#### Testing ####

This PR is itself e2e test coverage. It is wired into `.github/workflows/nightly-cni.yaml` and also runs on pull requests that touch the relevant paths.

#### Linked Issues ####

N/A

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####

The topology is intentionally reduced (1 LB, single keepalived instance) versus the doc's fully-redundant example; HA of the load balancer itself is out of scope for CI. The main runtime risk is the memory budget of 5 VMs on the GitHub runner, so the LB node is given minimal memory.

### DIFF
--- a/.github/workflows/nightly-cni.yaml
+++ b/.github/workflows/nightly-cni.yaml
@@ -9,6 +9,7 @@ on:
       - 'tests/e2e/cni/**'
       - 'tests/e2e/clusterloadbalancer/**'
       - 'tests/e2e/scripts/loadbalancer_setup.sh'
+      - 'tests/e2e/scripts/calico_ebpf_manifest.sh'
 
 jobs:
   e2e:
@@ -201,4 +202,3 @@ jobs:
       run: |
         E2E_CNI=${{ matrix.cni }} E2E_DATAPLANE=${{ matrix.dataplane }} \
         go test -timeout=75m ./clusterloadbalancer_test.go -ci -cni=${{ matrix.cni }} -dataplane=${{ matrix.dataplane }} -ginkgo.v -test.v
-
--- a/.github/workflows/nightly-cni.yaml
+++ b/.github/workflows/nightly-cni.yaml
@@ -7,6 +7,8 @@ on:
     paths:
       - '.github/workflows/nightly-cni.yaml'
       - 'tests/e2e/cni/**'
+      - 'tests/e2e/clusterloadbalancer/**'
+      - 'tests/e2e/scripts/loadbalancer_setup.sh'
 
 jobs:
   e2e:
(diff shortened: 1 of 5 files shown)


brandond (approved):

brandond:
approach looks good, lgtm when it passes checks

thomasferrandiz:
The failing nightly test will be fixed once a new latest version is released.

matttrach (approved):

Issue #10955, opened 2026-07-28
Leftover nft tables and blackhole routes


manuelbuil:
**Environmental Info:**
RKE2 Version: 

rke2 v1.36.2

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

1 server. CNI plugin: canal

**Describe the bug:**

When using cni plugin Canal, after running rke2-uninstall.sh, there are some leftovers which can be problematic

**Leftover 1**

there is one calico nft table left in the node:

$> sudo nft list tables
table arp calico-arp
table ip mangle
table ip raw
table ip filter
table ip nat
table ip6 mangle
table ip6 raw
table ip6 filter
table ip6 nat

**Leftover 2**

There is a blackhole route left in the routing table of the node:
$> ip r
...
blackhole 10.42.1.0/24
...

$> ip -6 r
...
blackhole 2001:cafe:42:1::/64 dev lo metric 1024 pref medium
...

**Steps To Reproduce:**

- Installed RKE2: 

Deploy RKE2 v1.36.2 with CNI plugin canal and then run `rke2-uninstall.sh`. After the uninstallation is completed:

1 - run `sudo nft list tables`, you will see a table `arp calico-arp`
2 - run `ip r` and `ip -6 r`, you will see the `blackhole` routes

**Expected behavior:**

No calico-arp table exists and no blackhole route exist

**Actual behavior:**

Calico-arp table and blackhole route exist

**Additional context / logs:**

PR #10956, opened 2026-07-28
Remove leftover nft tables and blackhole route from canal

manuelbuil:
#### Proposed Changes ####

This PR removes one dangerous leftovers from the canal CNI plugin installation and documents one:
1 - calico nft tables (arp calico-arp)
2 - flannel blackole route (both ipv4 and ipv6)

For the first one, this PR adds a loop over calico tables, even though there is currently only one: "arp calico-arp". Note that when installing with CNI plugin Calico, it is also left there.

For the second one, we need to identify the blackhole route created by Flannel. For that, we pick the net from the flannel interface and the mask from the route to other networks.

#### Types of Changes ####

Bugfix

#### Verification ####

1 - Deploy RKE2 multi-node (it can be HA or 1 server + N workers) with cni plugin canal
2 - Run `sudo nft list tables` and you will see the table "arp calico-arp"
3 - Run `ip r` and `ip -6 r` and you will see the blackhole route
4 - Uninstall rke2 with `rke2-uninstall.sh`
5 - Run `sudo nft list tables` and you will not see the table "arp calico-arp"
6 - Run `ip r` and `ip -6 r` and you will not see the blackhole route

#### Linked Issues ####

https://github.com/rancher/rke2/issues/10955

#### User-Facing Change ####

```release-note
Remove calico leftover nft tables and blackhole routes when uninstalling rke2
```

### DIFF
--- a/bundle/bin/rke2-killall.sh
+++ b/bundle/bin/rke2-killall.sh
@@ -69,6 +69,27 @@ ip link show 2>/dev/null | grep 'master cni0' | while read ignore iface ignore; 
     iface=${iface%%@*}
     [ -z "$iface" ] || ip link delete $iface
 done
+
+# --- Blackhole Canal route cleanup ---
+MASK_V4=$(ip -4 route show dev flannel.1 2>/dev/null | grep -oP '/\K[0-9]+' | head -n1)
+if [ -n "$MASK_V4" ]; then
+  LOCAL_IP_V4=$(ip -4 addr show dev flannel.1 2>/dev/null | awk '/inet / {print $2}' | cut -d/ -f1)
+  if [ -n "$LOCAL_IP_V4" ]; then
+    LOCAL_CIDR_V4="${LOCAL_IP_V4}/${MASK_V4}"
+    ip route del blackhole "$LOCAL_CIDR_V4" 2>/dev/null || true
+  fi
+fi
+
+MASK_V6=$(ip -6 route show dev flannel-v6.1 2>/dev/null | grep -oP '/\K[0-9]+' | head -n1)
+if [ -n "$MASK_V6" ]; then
+  LOCAL_IP_V6=$(ip -6 addr show dev flannel-v6.1 scope global 2>/dev/null | awk '/inet6 / {print $2}' | cut -d/ -f1)
+  if [ -n "$LOCAL_IP_V6" ]; then
+    LOCAL_CIDR_V6="${LOCAL_IP_V6}/${MASK_V6}"
+    ip -6 route del blackhole "$LOCAL_CIDR_V6" 2>/dev/null || true
+  fi
+fi
+
+# Delete interfaces
 ip link delete cni0
 ip link delete flannel.1
 ip link delete flannel.4096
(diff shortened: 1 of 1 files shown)


rbrtbnfgl (approved):

matttrach (approved):

bundle/bin/rke2-killall.sh
```diff
@@ -106,9 +108,18 @@ rm -f "${POD_MANIFESTS_DIR}/etcd.yaml" \
 iptables-save | grep -v KUBE- | grep -v CNI- | grep -v cali- | grep -v cali: | grep -v CILIUM_ | grep -v flannel | iptables-restore
 ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -v cali- | grep -v cali: | grep -v CILIUM_ | grep -v flannel | ip6tables-restore
 
+# Remove leftover calico nft tables
+nft list tables | grep calico | while read -r _ family name; do
```
brandond, line 131: do we need to `sudo nft list tables` command too, or is sudo only required to delete?

bundle/bin/rke2-killall.sh
```diff
@@ -106,9 +108,18 @@ rm -f "${POD_MANIFESTS_DIR}/etcd.yaml" \
...
 
-echo 'If this cluster was upgraded from an older release of the Canal CNI, you may need to manually remove some flannel iptables rules:'
+echo 'If this cluster was upgraded from an older release of the Canal CNI plugin, you may need to manually remove some flannel iptables rules:'
 echo -e '\texport cluster_cidr=YOUR-CLUSTER-CIDR'
 echo -e '\tiptables -D POSTROUTING -s $cluster_cidr -j MASQUERADE --random-fully'
 echo -e '\tiptables -D POSTROUTING ! -s $cluster_cidr -d  -j MASQUERADE --random-fully'
+
+echo 'If this cluster was using the Canal CNI Plugin, you need to manually remove a blackhole route:'
+echo -e '\tip route del blackhole "$ipv4_node_cidr="'
+echo -e '\tip -6 route del blackhole "$ipv6_node_cidr="'
```
brandond: is there no good way to identify these routes and safely remove them? Is it associated with a particular interface or anything else that might suggest it's for canal?

bundle/bin/rke2-killall.sh
manuelbuil: we need sudo for both, good catch! But because rke2-killall.sh must be run as root, we should remove it in both

bundle/bin/rke2-killall.sh
manuelbuil: Flannel interface gives us the network. However, the problem is how to identify the mask, which I thought it was not possible to calculate without querying kube-api. But I have just had an idea that might work (checking `ip route` command, since all nodes share the same mask)

manuelbuil: it works but not for one-node cluster since there won't be any route... I'll add a message for that one

matttrach (approved):

bundle/bin/rke2-killall.sh
brandond: if there's no route then what is there to clean up?

manuelbuil: Sorry, I was too cryptic. If there is only one node in the cluster, there won't be any flannel route to other nodes, so the approach to fetch the netmask using those routes doesn't work. Unfortunately, the blackhole route is still there, even if there is just one-node cluster. In canal, we need that blackhole route even for one-node cluster, otherwise traffic to non-existing pods escapes the node

brandond: I think we have enough info to find and remove the blackhole route even on single-node clusters, don't we? Its for the local CIDR, so it'll match the address of the flannel interface.

```console
root@rke2-server-001:/# kubectl get node
NAME                          STATUS   ROLES                AGE     VERSION
rke2-server-001.example.com   Ready    control-plane,etcd   8m25s   v1.36.2+rke2r1

root@rke2-server-001:/# ip addr show dev flannel.1 scope global
5: flannel.1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UNKNOWN group default
    link/ether e6:38:41:1f:69:b4 brd ff:ff:ff:ff:ff:ff
    inet 10.42.0.0/32 scope global flannel.1
       valid_lft forever preferred_lft forever

root@rke2-server-001:/# ip route show type blackhole
blackhole 10.42.0.0/24
```

thomasferrandiz (approved):

rbrtbnfgl (approved):

brandond (approved):

bundle/bin/rke2-killall.sh
manuelbuil: even if we know `10.42.0.0` is the address, we don't know what's the mask. It typically is `/24` but if the user changes it, it can be something different and we don't know for sure in one-node clusters :(

PR #10997, opened 2026-08-05
tests: Fix Nightly CNI Tests (canal, nftables)

mgfritch:
#### Proposed Changes ####

The nightly [Nightly CNI Tests (canal, nftables)](https://github.com/rancher/rke2/actions/runs/29820697846/job/88602400508#logs) tests are failing 

#### Types of Changes ####

The "Checks filtering backend" spec's canal nftables branch greps for
lowercase 'flannel' in the ruleset, but under canal, flannel's rules are
created via the iptables-nft compat layer as UPPERCASE FLANNEL-* chains.
    
Fixup for https://github.com/rancher/rke2/pull/10681

#### Verification ####

The [Nightly CNI ](https://github.com/rancher/rke2/actions/workflows/nightly-cni.yaml) tests succeed. 

#### Testing ####

yes

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/cni/cni_test.go
+++ b/tests/e2e/cni/cni_test.go
@@ -147,7 +147,7 @@ var _ = Describe("Verify DualStack in "+*cni+", "+*filter+" configuration", Orde
 				var calicoCmd, flannelCmd string
 				if *filter == "nftables" {
 					calicoCmd = "nft list ruleset | grep 'cali-'"
-					flannelCmd = "nft list ruleset | grep 'flannel'"
+					flannelCmd = "nft list ruleset | grep 'FLANNEL'"
 				} else {
 					calicoCmd = "iptables -L | grep 'cali-'"
 					flannelCmd = "iptables -L -t nat| grep 'FLANNEL'"


tests/e2e/cni/cni_test.go
```diff
@@ -147,7 +147,7 @@ var _ = Describe("Verify DualStack in "+*cni+", "+*filter+" configuration", Orde
 				var calicoCmd, flannelCmd string
 				if *filter == "nftables" {
 					calicoCmd = "nft list ruleset | grep 'cali-'"
-					flannelCmd = "nft list ruleset | grep 'flannel'"
+					flannelCmd = "nft list ruleset | grep 'FLANNEL'"
```
thomasferrandiz, line 150: The correct check here is to look for 'flannel', not 'FLANNEL'.
When nftables is used by flannel, the table names are in lower case.

The issue here is that the canal chart wasn't setting flannel's config correctly so the flannel part never used nftables and we ended up with the upper-case 'FLANNEL' table name from the conversion layer.

This should be solved when the 'latest' channel points to the July release where the chart is fixed.

thomasferrandiz (changes requested):

tests/e2e/cni/cni_test.go
mgfritch: ah ok, this was resolved by: https://github.com/rancher/rke2-charts/pull/1081

But the nightly tests (and PR checks) use `latest` from the channel server, which will cause these to fail until a new latest release that includes the new chart is tagged.

maybe these tests should run against the main (master) branch instead?

mgfritch: https://github.com/rancher/rke2/pull/11000

mgfritch:
close in favor of https://github.com/rancher/rke2/pull/11000

PR #11000, opened 2026-08-05
nightly-cni: test against master build instead of latest release

thomasferrandiz:
#### Proposed Changes ####

The nightly CNI workflow was installing RKE2 from the published `latest` release channel,
which means it tested a released version rather than the code on the master branch. This
change makes all three job groups (CNI, cluster load balancer, kube-vip) build RKE2 from
the checked-out source and install that locally built binary and images into the test VMs.

The approach follows the same pattern already used in `test-suite.yaml`: a `build` job
compiles RKE2 (`make package-bundle` + `make package-image-runtime`), uploads the artifacts,
and each e2e job downloads them before running the test with `-local`.

Changes:
- `nightly-cni.yaml`: adds a `build` job; all three e2e jobs now `needs: build`, drop
  `INSTALL_RKE2_CHANNEL: latest`, download+place the artifacts, and pass `-local` to
  `go test`.
- `tests/e2e/testutils.go`: exports `SCPRke2Artifacts` so test packages outside the
  `e2e` package can copy artifacts to VMs.
- `tests/e2e/clusterloadbalancer/clusterloadbalancer_test.go`: adds `-local` flag;
  in local mode the cluster is brought up without provisioning first, artifacts are
  scoped to the RKE2 nodes (not the HAProxy/Keepalived lb node), and provisioning
  follows the required lb-0 -> server-0 -> rest order.
- `tests/e2e/kubevip/kubevip_test.go`: adds `-local` flag; in local mode all nodes are
  brought up without provisioning, artifacts are copied to all nodes, then provisioning
  follows server-0 -> waitForVIP -> rest order so kube-vip can claim the VIP before
  the other nodes join.

#### Types of Changes ####

Test/CI infrastructure improvement -- no change to RKE2 runtime behavior.

#### Verification ####

The workflow can be triggered manually via `workflow_dispatch`. A PR that touches
`.github/workflows/nightly-cni.yaml` or any of the three test suites will also
exercise the new build + local-install path.

#### Testing ####

The Go test packages compile cleanly (`go build ./tests/e2e/...`). The `-local` path
in cni_test.go was already exercised by `test-suite.yaml`; the same pattern is now
applied consistently to the LB and kube-vip suites.

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

The `clusterloadbalancer` and `kubevip` suites use custom cluster bring-up functions
(not the generic `CreateLocalCluster` helper) because they have strict node ordering
requirements (VIP must exist before followers join). The `-local` path in each suite
preserves that ordering while inserting the `--no-provision` boot + scp + provision
sequence that the local install requires.

### DIFF
--- a/.github/workflows/nightly-cni.yaml
+++ b/.github/workflows/nightly-cni.yaml
@@ -14,8 +14,57 @@ on:
       - 'tests/e2e/scripts/calico_ebpf_manifest.sh'
 
 jobs:
+  build:
+    name: Build RKE2 Images and Binary
+    runs-on: ${{ github.repository == 'rancher/rke2' && 'runs-on,runner=16cpu-linux-x64,image=ubuntu24-full-x64' || 'ubuntu-24.04' }}
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+    - name: Checkout
+      # https://github.com/actions/checkout/releases
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        fetch-depth: 1
+    - name: Find Go Version for Build
+      id: go-finder
+      run: |
+        GOOS=linux GOARCH=amd64 . ./scripts/version.sh
+        set +x
+        VERSION_GOLANG=$(echo $VERSION_GOLANG | sed 's/go//')
+        echo "VERSION_GOLANG=${VERSION_GOLANG}" >> "$GITHUB_OUTPUT"
+    - name: Install Go
+      uses: ./.github/actions/setup-go
+      with:
+        go-version: ${{ steps.go-finder.outputs.VERSION_GOLANG }}
+    - name: Set up Docker Buildx
+      # https://github.com/docker/setup-buildx-action/releases
+      uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+    - name: Install OS Packages
+      run: sudo apt-get update && sudo apt-get install -y libarchive-tools g++-mingw-w64-x86-64 gcc-mingw-w64-x86-64
+    - name: "Read secrets"
+      if: github.event_name != 'pull_request' && github.repository == 'rancher/rke2'
+      uses: rancher-eio/read-vault-secrets@7282bf97898cd1c16c89f837e0bb442e6d384c89 # v3
+      with:
+        secrets: |
+          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials registry | PRIME_REGISTRY ;
+    - name: Build RKE2 Binary and Compressed Runtime Image
+      run: |
+        make package-bundle
+        make package-image-runtime
+        cp ./bin/rke2 ./build/images/rke2
+        cp ./dist/artifacts/rke2.*-amd64.tar.gz ./build/images/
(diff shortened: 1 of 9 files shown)


tests/e2e/testutils.go
```diff
@@ -253,6 +253,13 @@ func scpRKE2Artifacts(nodes []VagrantNode) error {
 	return nil
 }
 
+// SCPRke2Artifacts copies the locally built RKE2 binary, tarball, and image archive
+// to each node so they can be installed without fetching from a release channel.
+// It delegates to scpRKE2Artifacts and is exported for use in external test suites.
+func SCPRke2Artifacts(nodes []VagrantNode) error {
```
dereknola: Just expose the original function.

tests/e2e/cni/cni_test.go
```diff
@@ -119,11 +119,22 @@ var _ = Describe("Verify DualStack in "+*cni+", "+*filter+" configuration", Orde
 		}, "180s", "5s").Should(Succeed())
 	})
 
-	It("Waits for traefik daemonset readiness", func() {
-		By("waiting for traefik daemonset readiness")
+	It("Waits for the bundled ingress controller daemonset readiness", func() {
```
dereknola, line 122: This should always be traefik for the latest versions of RKE2, we should want to ensure this holds true. Backports will just need to be modified to look for ingress-nginx.

mgfritch (approved):
awesome, thanks!

.github/workflows/nightly-cni.yaml
```diff
@@ -14,8 +14,57 @@ on:
       - 'tests/e2e/scripts/calico_ebpf_manifest.sh'
 
 jobs:
+  build:
```
dereknola, line 17: We should consolidate this in a future PR between this testing and the test-suite.yml. We have a build.yml workflow, but I don't believe its being utilized.

dereknola (changes requested):

tests/e2e/testutils.go
thomasferrandiz: Done, renamed the original function directly.

tests/e2e/cni/cni_test.go
thomasferrandiz: That's what I was doing when testing with the latest release but the test was failing on master becasue ingress-nginx was deployed but not traefik.
Is that expected or it is an issue in the build?

.github/workflows/nightly-cni.yaml
thomasferrandiz: Agreed, consolidating the build jobs and making use of build.yml would be a good follow-up. I will open a separate PR for that.

tests/e2e/cni/cni_test.go
dereknola: We should have traefik deployed on master. Only PRIME and older releases should be deploying ingress-nginx by default. If your seeing this, then there is a bug.

thomasferrandiz: Good point. Reverted to only waiting for rke2-traefik -- the ingress-nginx fallback was based on a wrong assumption. Backports that ship ingress-nginx can adjust the daemonset name there.

thomasferrandiz:
@dereknola It should be fixed now. The vagrant setup was using the 'stable' release of rke2 instead of the local build from master.

dereknola (approved):

Issue #11010, opened 2026-08-07
Recent windows tar-balls/assets hold docker.io/rancher/mirrored-pause:i-was-a-digest instead of real pause-image version


Najzer0:
**Environmental Info:**
RKE2 Version: 
Noticed with v1.35.7+rke2r1 but just cherry picked a recent other drop and its the same in their manifest.json

**Describe the bug:**
we need to receive the pause-image version to inject it into our windows nodes via config.yaml "pause-image: value".
As its not part of the manifest any more.. thats kinda broken ;-)

brandond:
Where exactly are you looking? Windows image lists look ok...
https://github.com/rancher/rke2/releases/tag/v1.35.7%2Brke2r1
> https://github.com/rancher/rke2/releases/download/v1.35.7%2Brke2r1/rke2-images.windows-amd64.txt
>>```
>> docker.io/rancher/mirrored-pause:3.10.2
>>```

Why would you be unpacking the whole image tarball to introspect the manifest when the info you need is in the list artifact?

Najzer0:
ah, thank you for your input brandon. Its just we have this laying on the designated (to be installed windows) nodes anyway and looking into the manifest for the pause version is easy lifting, compared to having another .txt as part of our airgap drop.

It just recently started behaving differently, so thats why I spoke up

brandond:
It changed because with the new pause version there are no longer distinct tags for each windows OS version. See: 
* https://github.com/rancher/rke2/commit/1e1d9b25c94b35cf36f9f3bd64dbbaa8ff66a727
* https://github.com/rancher/rke2/commit/f7aaea90ce3c463fd0bd5975ca3da038d77c7bd0

Are you sure you still need this logic?

Najzer0:
> It changed because with the new pause version there are no longer distinct tags for each windows OS version. See:
> 
> * [1e1d9b2](https://github.com/rancher/rke2/commit/1e1d9b25c94b35cf36f9f3bd64dbbaa8ff66a727)
> * [f7aaea9](https://github.com/rancher/rke2/commit/f7aaea90ce3c463fd0bd5975ca3da038d77c7bd0)
> 
> Are you sure you still need this logic?

I noticed the "issue" when creating sbom for our internal component tracking and saw "i-was-a-digest" (skopeo + trivy) instead of version numbers for pause. I think its not as uncommon, so others will most likely trip over this.
[...]

PR #11114, opened 2026-08-21
[master] Fix 'no such image' in Release CI

rafaelbreno:
#### Proposed Changes ####

Backports:
- #11113 
- #11112  
- #11111  

`scripts/package-windows-images` was rewritten in #11028 to move from `crane pull` to `docker pull` / `docker image save`, in order to preserve the `mirrored-pause` tag reference in the airgap tarball (see #11010, where digest-based pinning caused SBOM and `pause-image:` config regressions for Windows nodes).

The rewrite introduced two bugs that broke the `create-draft-release` workflow:

1. `${PROG}-runtime` (rke2-runtime) was never pulled before `docker image save` ran, causing:
   ```
   Error response from daemon: No such image: rancher/rke2-runtime:v1.34.11-rc1-rke2r1-windows-amd64
   ```

2. Once the missing pull was added, a second latent bug surfaced. All three Windows `os.version` variants of `mirrored-pause` share a single tag (`mirrored-pause:3.10.2`) in the registry manifest list. Pulling all three locally left the local image store unable to disambiguate which manifest `docker image save --platform 'windows(BUILD)/amd64'` should select, causing:
   ```
   Error response from daemon: no suitable export target found: multiple manifests found for platform windows(10.0.17763.7919)/amd64
   ```

Additionally, the `rke2-windows-2025-amd64-images.tar` block was resolving against the same `os.version` (`10.0.20348.4294`) as the `ltsc2022` block instead of its own (`10.0.26100.6899`), so it was silently bundling the wrong `mirrored-pause` variant.

This PR:
- Adds the missing `docker pull` for `rke2-runtime` before the save step.
- Resolves each Windows variant's `mirrored-pause` digest explicitly via `docker buildx imagetools inspect`, then `docker tag`s the local `mirrored-pause:${PAUSE_VERSION}` reference to point at exactly one digest immediately before each `docker image save` call. This avoids relying on `docker image save --platform` to disambiguate, while still preserving the human-readable tag in the resulting tarball (per the fix in #11010).
- Fixes the 2025 build to resolve against `10.0.26100.6899` instead of reusing the ltsc2022 build number.

#### Types of Changes ####

Bugfix.

#### Verification ####

Reproduced locally against tag `v1.34.11-rc1+rke2r1`, with `containerd-snapshotter` enabled in the local Docker daemon to match the CI runner config in `docker/setup-docker-action`.

1. Confirmed original failure (`No such image: rke2-runtime`) reproduces with the unmodified script.
2. Confirmed the ambiguous-manifest failure reproduces once the missing pull is added back in, isolating it as a second, independent bug.
3. Confirmed the manifest list served by the registry is unambiguous (`docker manifest inspect`), ruling out a registry-side issue and confirming the ambiguity is local-store only.
4. Applied the fix and confirmed all three `docker image save` calls succeed:
   ```
   tar -xOf build/images/rke2-windows-1809-amd64-images.tar manifest.json | jq '.[].RepoTags'
   tar -xOf build/images/rke2-windows-ltsc2022-amd64-images.tar manifest.json | jq '.[].RepoTags'
   tar -xOf build/images/rke2-windows-2025-amd64-images.tar manifest.json | jq '.[].RepoTags'
   ```
   All three tarballs preserve `rancher/mirrored-pause:3.10.2` as a tag reference, not a digest.
5. Confirmed the three tarballs bundle genuinely distinct `mirrored-pause` content (not the same digest three times) by comparing config blob digests across all three tarballs, which resolved to three distinct SHAs.

#### Testing ####

Manual verification as described above. This script runs inside `create-draft-release` in CI and is exercised by fork CI on this branch prior to merge.

#### Linked Issues ####

Fixes a regression introduced by #10942.
Fixes a bug introduced by #11028
Related: #11010

#### Further Comments ####

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,9 @@ jobs:
         version: type=image,tag=29
         daemon-config: '{"features":{"containerd-snapshotter":true}}'
         set-host: true
+    - name: Set up Containerd
+      run: |
+        echo CONTAINERD_ADDRESS=$(docker info -f '{{.Containerd.Address}}') >> "$GITHUB_ENV"
     - name: Checkout code
       uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
     - name: "Read secrets"
@@ -62,6 +65,9 @@ jobs:
         version: type=image,tag=29
         daemon-config: '{"features":{"containerd-snapshotter":true}}'
         set-host: true
+    - name: Set up Containerd
+      run: |
+        echo CONTAINERD_ADDRESS=$(docker info -f '{{.Containerd.Address}}') >> "$GITHUB_ENV"
     - name: Checkout code
       uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
     - name: "Read secrets"
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,6 +19,9 @@ jobs:
         version: type=image,tag=29
         daemon-config: '{"features":{"containerd-snapshotter":true}}'
         set-host: true
+    - name: Set up Containerd
+      run: |
+        echo CONTAINERD_ADDRESS=$(docker info -f '{{.Containerd.Address}}') >> "$GITHUB_ENV"
     - name: Checkout code
       uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
     - name: Build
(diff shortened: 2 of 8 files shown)


brandond (changes requested):
No, don't do this.

If this step is failing, it is because you are not running `./scripts/build-windows-images` first in the same job. The whole point is that the images are built locally and just saved out, instead of pushing them to a remote registry and then pulling+saving them back out.

Run `make package-windows-images` and it will run both scripts for you in the correct order.

Edit: I see that the release CI breaks up the build-and-push and tarball+upload steps across multiple jobs so that they can be retried independently. I think we can make that work...

brandond (approved):

briandowns (approved):

Issue #11081, opened 2026-08-19
More improvements for tarball installer with rke2-selinux


rancher-max:
**Is your feature request related to a problem? Please describe.**

The restorecon from the install script covers too wide of a surface area when rke2 was installed via the tarball method.

**Describe the solution you'd like**

Have the restorecon skip files that no longer exist.

**Describe alternatives you've considered**

- Do not attempt the restorecon at all and have users handle this manually themselves
- Skip the snapshots directly entirely, although this might have other consequences for files that _do_ exist.

**Additional context**

Current internal Jira: SURE-11906
Previous issue that put this in: https://github.com/rancher/rke2/issues/9694
Specific function on install script: https://github.com/rancher/rke2/blob/master/install.sh#L676-L699

rancher-max:
Putting some additional details and suggestions here:

1. `restorecon` actually does already ignore files that do not exist via the `-i` flag. This issue generally only presents itself in highly active environments and only happens because it's running while rke2 is running. See log snippet below.
```
Aug  4 19:10:54 xxxxxx rancher-system-agent[1489]: time="2026-08-04T19:10:54+03:00" level=info msg="[e259d7d6c2535a2d7c96fdab85b2b652be2a945a716dd02b608c3cc5009333ad_0:stderr]: restorecon: Could not read /var/lib/rancher/rke2/agent/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1063: No such file or directory."
Aug  4 19:10:55 xxxxxx rancher-system-agent[1489]: time="2026-08-04T19:10:55+03:00" level=info msg="[e259d7d6c2535a2d7c96fdab85b2b652be2a945a716dd02b608c3cc5009333ad_0:stderr]: restorecon: Could not set context for /var/lib/rancher/rke2/agent/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1039:  No such file or directory"
```
2. This presents itself mostly in rancher due to the way it all works with rancher-system-agent. It's possible that this could be updated in that use case to include the `INSTALL_RKE2_SKIP_RESTORECON` env var similar to how some other env vars are set in the snippet below, but that also may not achieve the same end results where users might _want_ the restorecon.
```
[...]

PR #11137, opened 2026-08-28
Update install script to warn instead of exit on restorecon errors

rancher-max:
#### Proposed Changes ####

This PR aims to update the recently added restorecon function so that if any restorecon command were to fail and cause the install script to exit nonzero, it will instead give a warning but continue with the rest of the install script and provide exit status 0.

#### Types of Changes ####

Enhancement/Fix

#### Verification ####

Should be able to follow the same steps as in https://github.com/rancher/rke2/issues/9694#issuecomment-4130457049 but add another test: while the install script is running, add and remove files to any of the folders. This should've previously caused a failure, but should only output a warning moving forward.

I'll add a comment to this PR with some tests I'll run myself before merge since this goes live upon merging.

#### Testing ####

General functionality covered by existing nightly install test running on `alma-10-with-selinux-tar`. 

#### Linked Issues ####

https://github.com/rancher/rke2/issues/11081

#### User-Facing Change ####

```release-note
When rke2-selinux is installed in tarball, restorecon operations will warn on error instead of fail
```

#### Further Comments ####

- The verification should be done across every OS in our support matrix that we support rke2-selinux with, specifically SLE Micro and *EL-like distros.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -679,20 +679,23 @@ do_restorecon_tar() {
     if command -v rpm >/dev/null 2>&1; then
         if rpm -q --quiet rke2-selinux; then
             info "applying correct SELinux contexts to RKE2 files"
+            (
+              set +e
 
-            # this is for the .service files
-            err_msg=$(restorecon -R -i /etc/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /etc/systemd/system/rke2*: ${err_msg}"
-            err_msg=$(restorecon -R -i /usr/local/lib/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /usr/local/lib/systemd/system/rke2: ${err_msg}"
-            err_msg=$(restorecon -R -i /usr/lib/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /usr/lib/systemd/system/rke2: ${err_msg}"
+              # this is for the .service files
+              restorecon -R -i /etc/systemd/system/rke2*
+              restorecon -R -i /usr/local/lib/systemd/system/rke2*
+              restorecon -R -i /usr/lib/systemd/system/rke2*
 
-            # this one is for the bin
-            err_msg=$(restorecon -R -i /opt/rke2 2>&1) || warn "Failed to run restorecon on /opt/rke2: ${err_msg}"
-            err_msg=$(restorecon -R -i /usr/local/bin/rke2* 2>&1) || warn "Failed to run restorecon on /usr/local/bin/rke2*: ${err_msg}"
-            err_msg=$(restorecon -R -i /usr/bin/rke2* 2>&1) || warn "Failed to run restorecon on /usr/bin/rke2*: ${err_msg}"
+              # this one is for the bin
+              restorecon -R -i /opt/rke2
+              restorecon -R -i /usr/local/bin/rke2*
+              restorecon -R -i /usr/bin/rke2*
 
-            if [ -n "${INSTALL_RKE2_ARTIFACT_PATH}" ]; then
-                err_msg=$(restorecon -R -i /var/lib/rancher/rke2 2>&1) || warn "Failed to run restorecon on /var/lib/rancher/rke2: ${err_msg}"
-            fi
+              if [ -n "${INSTALL_RKE2_ARTIFACT_PATH}" ]; then
+                  restorecon -R -i /var/lib/rancher/rke2
+              fi
+            ) || true
         fi
     fi
 }
--- a/install.sh
+++ b/install.sh
@@ -681,17 +681,17 @@ do_restorecon_tar() {
             info "applying correct SELinux contexts to RKE2 files"
 
             # this is for the .service files
-            restorecon -R -i /etc/systemd/system/rke2*
-            restorecon -R -i /usr/local/lib/systemd/system/rke2*
-            restorecon -R -i /usr/lib/systemd/system/rke2*
+            err_msg=$(restorecon -R -i /etc/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /etc/systemd/system/rke2*: ${err_msg}"
+            err_msg=$(restorecon -R -i /usr/local/lib/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /usr/local/lib/systemd/system/rke2: ${err_msg}"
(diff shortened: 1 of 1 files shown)


vitorsavian (approved):

caroline-suse-rancher (approved):

install.sh
```diff
@@ -681,17 +681,17 @@ do_restorecon_tar() {
             info "applying correct SELinux contexts to RKE2 files"
 
             # this is for the .service files
-            restorecon -R -i /etc/systemd/system/rke2*
-            restorecon -R -i /usr/local/lib/systemd/system/rke2*
-            restorecon -R -i /usr/lib/systemd/system/rke2*
+            err_msg=$(restorecon -R -i /etc/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /etc/systemd/system/rke2*: ${err_msg}"
+            err_msg=$(restorecon -R -i /usr/local/lib/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /usr/local/lib/systemd/system/rke2: ${err_msg}"
+            err_msg=$(restorecon -R -i /usr/lib/systemd/system/rke2* 2>&1) || warn "Failed to run restorecon on /usr/lib/systemd/system/rke2: ${err_msg}"
```
brandond: Rather than capturing the text from each of these and printing a warning... why not just run them in a subshell and disable exit-on-error within that block? We're not adding much with the capture-and-warn wrapper that the command's output doesn't already handle, and it's a fair bit less readable.
```suggestion
            (
                set +e
                restorecon -R -i /etc/systemd/system/rke2*
                restorecon -R -i /usr/local/lib/systemd/system/rke2*
                restorecon -R -i /usr/lib/systemd/system/rke2*
             )
```

brandond (changes requested):

install.sh
rancher-max: I thought about doing it like this actually (with the full block that was changed) and including ` || true` at the end so that the subshell doesn't potentially get a nonzero exitcode. I'll change it to that since that's the preference 👍

brandond (approved):

caroline-suse-rancher (approved):
Much prettier :)

rancher-max:
<details>
  <summary><b>Test Script. Click to Expand...</b></summary>

  _Disclaimer: I used AI to create this script to validate these changes locally_


  ```sh
  #!/bin/sh
  set -eu
  
  TARGET="${1:-./install.sh}"
  
  extract_do_restorecon_tar() {
      sed -n '/^[[:space:]]*do_restorecon_tar()[[:space:]]*{/,/^[[:space:]]*}[[:space:]]*$/p' "$1"
  }
  
  info() { echo "[INFO] $*"; }
  warn() { echo "[WARN] $*" >&2; }
  
  rpm() {
      if [ "${1:-}" = "-q" ] && [ "${2:-}" = "--quiet" ] && [ "${3:-}" = "rke2-selinux" ]; then
          return 0
      fi
      return 1
  }
  
  restorecon() {
      echo "restorecon invoked: $*" >&2
      case "$*" in
          *"/usr/local/lib/systemd/system/rke2*"*)
              echo "simulated restorecon error: /usr/local/lib/systemd/system/rke2* context lookup failed" >&2
              return 1
              ;;
          *"/opt/rke2"*)
              echo "simulated restorecon error: /opt/rke2 resource busy" >&2
              return 1
              ;;
      esac
      return 0
  }
  
  [ -f "$TARGET" ] || {
      echo "Target not found: $TARGET" >&2
      exit 2
  }
  
  FN="$(extract_do_restorecon_tar "$TARGET")"
  [ -n "$FN" ] || {
      echo "Could not extract do_restorecon_tar from $TARGET" >&2
      exit 3
  }
  
  eval "$FN"
  
  INSTALL_RKE2_ARTIFACT_PATH=/tmp/fake-artifacts
  
  echo "=== Running do_restorecon_tar under parent set -e ==="
  set +e
  (
      set -e
      do_restorecon_tar
      echo "FUNCTION_COMPLETED"
  )
  rc=$?
  set -e
  
  echo "FUNCTION_EXIT_CODE=$rc"
  if [ "$rc" -eq 0 ]; then
      echo "RESULT: continued after restorecon failures"
  else
      echo "RESULT: aborted early due to non-zero return"
  fi

  ```

</details>

Note that `install.sh` points to get.rke2.io and `new-install.sh` points to the new version from this PR.
```sh
$ sh /tmp/test-restorecon-behavior.sh ./install.sh 
=== Running do_restorecon_tar under parent set -e ===
[INFO] applying correct SELinux contexts to RKE2 files
restorecon invoked: -R -i /etc/systemd/system/rke2*
restorecon invoked: -R -i /usr/local/lib/systemd/system/rke2-agent.env /usr/local/lib/systemd/system/rke2-agent.service /usr/local/lib/systemd/system/rke2-server.env /usr/local/lib/systemd/system/rke2-server.service
restorecon invoked: -R -i /usr/lib/systemd/system/rke2*
restorecon invoked: -R -i /opt/rke2
simulated restorecon error: /opt/rke2 resource busy
FUNCTION_EXIT_CODE=1
RESULT: aborted early due to non-zero return

$ sh /tmp/test-restorecon-behavior.sh ./new-install.sh 
=== Running do_restorecon_tar under parent set -e ===
[INFO] applying correct SELinux contexts to RKE2 files
restorecon invoked: -R -i /etc/systemd/system/rke2*
restorecon invoked: -R -i /usr/local/lib/systemd/system/rke2-agent.env /usr/local/lib/systemd/system/rke2-agent.service /usr/local/lib/systemd/system/rke2-server.env /usr/local/lib/systemd/system/rke2-server.service
restorecon invoked: -R -i /usr/lib/systemd/system/rke2*
restorecon invoked: -R -i /opt/rke2
simulated restorecon error: /opt/rke2 resource busy
restorecon invoked: -R -i /usr/local/bin/rke2 /usr/local/bin/rke2-killall.sh /usr/local/bin/rke2-uninstall.sh
restorecon invoked: -R -i /usr/bin/rke2*
restorecon invoked: -R -i /var/lib/rancher/rke2
FUNCTION_COMPLETED
FUNCTION_EXIT_CODE=0
RESULT: continued after restorecon failures
```

I will allow you all to merge when / if you all want to put this change in 🙂

brandond:
Should not require any backports, since the install script is always served from master branch.

PR #11138, opened 2026-08-30
Add bash completions to install.sh

megabreit:
#### Proposed Changes ####

Add a feature to activate rke2 shell completions at install time.

#### Types of Changes ####

New feature

#### Verification ####

Set the variable INSTALL_RKE2_COMPLETION=<bash,zsh> on the install command line and check for the RKE2 completion files ~/.bashrc or ~/.zshrc

#### Testing ####

Not yet implemented. It's creating completion entries in ~/.bashrc or ~/.zshrc at install time

#### Linked Issues ####

None

#### User-Facing Change ####

```release-note
Adds the variable INSTALL_RKE2_COMPLETION to the installer to add shell completions for the root user
```

#### Further Comments ####

This is for first discussion and asking for assistance and comments. It shows a skeleton of what should be achieved.
It's incomplete and probably needs various improvements. So let's discuss:

Missing things/questions:

- Are 4 variables too much? Is one better to install all completions at once, but without the possibility to choose? What are the preferred ways of adding such a feature?
- SELinux and fapolicy. Is this required or needed for the completion files?
- add this feature to the RPM package (is this needed, or optional? Can RKE2 be installed without the installer, just with RPM?)
- update completion files at RKE2 upgrade at RPM upgrade
- removal of the completion files at RPM uninstall (is it a supported way to uninstall the RPM without running rke2-uninstall.sh?)
- add even more completions? kubectl and crictl completions are candidates -> not possible

Would you accept/support this feature in general?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -689,6 +689,7 @@ do_restorecon_tar() {
             restorecon -R -i /opt/rke2
             restorecon -R -i /usr/local/bin/rke2*
             restorecon -R -i /usr/bin/rke2*
+	    # TODO: Add the shell completion files here too?
 
             if [ -n "${INSTALL_RKE2_ARTIFACT_PATH}" ]; then
                 restorecon -R -i /var/lib/rancher/rke2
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -707,13 +708,37 @@ allow perm=any all : dir=/opt/cni/
...
             fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
             systemctl restart fapolicyd
         fi
     fi
 }
 
+do_add_completion() {
+    COMPLETION_DIR="/etc/bash_completion.d"
+    if [ ! -z "${INSTALL_RKE2_COMPLETION_ALL}" ]; then
+        INSTALL_RKE2_COMPLETION_RKE2=1
```
dereknola: 4 separate ENV VARS is too much. I would take a single INSTALL_RKE2_COMPLETION=<bash/zsh> and assume you want all the completions for that shell type.

dereknola (changes requested):

install.sh
megabreit: Ok. I would also add the crictl and kubectl binary completion in this case.

dereknola: You still haven't address the bash/zsh option. You need to not rely on a `1/0` but a string as the value of `bash/zsh`. And you should remove the TODOs around RPMs. Lets just focus on installing tarball only.

install.sh
```diff
@@ -707,13 +708,37 @@ allow perm=any all : dir=/opt/cni/
...
+	# TODO: Add the shell completion files here too?
         if [ -z "${INSTALL_RKE2_SKIP_RELOAD}" ]; then
             fagenrules --load || fatal "failed to load rke2 fapolicyd rules"
             systemctl restart fapolicyd
         fi
     fi
 }
 
+do_add_completion() {
+    COMPLETION_DIR="/etc/bash_completion.d"
```
dereknola: Lets just make these installed under `/etc/rancher/rke2`. The uninstall already handles everything in that directory.

dereknola (changes requested):

install.sh
megabreit: I can do that. But then I still need to link the files into /etc/bash_completion.d or /usr/local/share/zsh/site-functions to make them work for all users... the uninstall then needs to remove the links instead.

install.sh
megabreit: TIL: kubectl and crictl completions are out of scope because the binaries are not yet present when the installer is finished.

install.sh
megabreit: I added zsh and kept the associated completion directories for now. Maybe adding the hypothetical path /etc/rancher/rke2/completions/bash in /etc/bash_completion would work. But I don't know any similar mechanism for zsh to achieve the same. Copy the completion file to their intended directories seems logical.

install.sh
megabreit: Done.

install.sh
dereknola: Why don't you just use the `-i` flag? That's why we wrote it, to make it easy to install into the users bashrc for sourcing. 

We don't have an easy uninstall, but that could be added in a future PR, or you could manually sed replace the lines in the bashrc on uninstall.

megabreit: The actual intention was to make it work for all users, not just root. With this at least `kubectl` could be used by normal users (if /etc/rancher/rke2/rke2.yaml has the correct read permission). `crictl` and `ctr` are out of scope, they can't be used with normal non-root permissions. But creating the completion only for root would also work. It's probably used mostly by root anyway... and the MVP for this PR :-)

The removal of the completion is also no big issue. `rke2` sets markers around the completion, so for bash a simple `sed -i '/# >> rke2 command completion (start)/,/# >> rke2 command completion (end)/d' ~/.bashrc` would do the trick. Or the `rke2 completion` command could add a `-u` option to be able to remove it itself. But that's beyond my coding abilities.

I'll change the code to just work with the root profiles.

megabreit: @dereknola I changed the installation and removal for the root profiles. Because the `rke2 completion -i`  does not check if the completion is already present and adds it multiple times, the removal has to deal with multiple blocks. The removal only works correctly when the markers are on separate lines. This is why I added newlines in front and after the added block.
The only possible issue I see: If the completion ever changes, it will not be updated anymore. I could always try to remove it before adding it though...

dereknola (approved):

dereknola:
You'll need to rebase off master to resolve the conflicts, and then I'll reapprove

dereknola (approved):

brandond (approved):

Issue #8628, opened 2025-07-16
Move multus/whereabouts CRDs to -crd chart


brandond:
* *From https://github.com/rancher/rke2/issues/5586#issuecomment-3079683608*

The multus whereabouts subchart contains CRDs that don't get installed if you update the multus chart values to enable whereabouts AFTER the initial installation.

We need to move the whereabouts CRDs out of the chart's crds/ folder into a -crd chart. This will likely require adding support for the `--take-ownership` helm cli flag to helm-controller to avoid having helm freak out when the resources move.

caroline-suse-rancher:
@brandond does this last PR solve this issue, or more needs to be done?

brandond:
With that feature in the k3s/rke2 helm-controller we can now move resources from one chart to another, so we're unblocked on that. Someone still needs to do the actual thing that we discussed doing here.

thomasferrandiz:
I did some tests and this  doesn't work with the way we currently generate charts.
Since the `rke2-whereabouts` chart is itself a dependency of `rke2-multus`, the script does not generate another sub-chart for the whereabouts crds so we end up in the same situation as before.

I guess we could include whereabouts directly in the multus chart so that the CRDs are always installed when multus is deployed.
We can also keep whereabouts separate but install it by default since the alternative is to use `host-local` which is not useful in real-world scenarios.

mdrahman-suse:
@caroline-suse-rancher is this ready to Test?

thomasferrandiz:
The migration needs to take place step-by-step:
In 1.36, we [added](https://github.com/rancher/rke2-charts/pull/948) a 'keep' annotation to the CRDs to make sure it's not accidentally deleted during an upgrade.
In 1.37, we will be able to migrate the CRDs to their own sub-chart.

endawkins:
@mgfritch  I have tried to validate the issue but am still seeing issues where the CRDs are not showing as installed:

Here are my steps:
1. Install RKE2 with thick multus and whereabouts enabled
2. Check to verify that the whereabouts CRDs are installed.

```
[...]

PR #11166, opened 2026-09-04
Disable rke2-multus-crd chart when multus is not selected

mgfritch:
#### Proposed Changes ####

Disable rke2-multus-crd chart when multus is not selected

#### Types of Changes ####

Fixup: 4b865f2adba843f44a57b27031306f951e66cbbb

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/8628
* https://github.com/rancher/rke2/issues/11169

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -226,6 +226,7 @@ func validateCNI(clx *cli.Context) {
 				return nil, errors.New("multus must be used alongside another primary cni selection")
 			}
 			clx.Set("disable", "rke2-multus")
+			clx.Set("disable", "rke2-multus-crd")
 		case 2:
 			if values[0] == "multus" {
 				values = values[1:]


brandond (approved):

pkg/cli/cmds/server.go
```diff
@@ -226,6 +226,7 @@ func validateCNI(clx *cli.Context) {
 				return nil, errors.New("multus must be used alongside another primary cni selection")
 			}
 			clx.Set("disable", "rke2-multus")
+			clx.Set("disable", "rke2-multus-crd")
```
brandond, line 229: I am curious if this is actually necessary? There is some special handling for CRD charts further down this file. The intent was that disabling a chart would also disable its crd chart.

https://github.com/rancher/rke2/blob/f1bd72501fef1adeb7d42d0ab85c13e990df1225/pkg/cli/cmds/server.go#L261-L278

Perhaps this is getting bypassed because we're adding the multus disable to the CLI flag that runs?

mdrahman-suse (approved):

pkg/cli/cmds/server.go
mgfritch: The handling applies to charts derived from `CNIItems`

https://github.com/rancher/rke2/blob/f1bd72501fef1adeb7d42d0ab85c13e990df1225/pkg/cli/types.go#L10

But multus is not listed there since it is a special case where it must be paired with another CNI.

PR #11174, opened 2026-09-08
Include gateway-api test

manuelbuil:
#### Proposed Changes ####

This PR adds a gateway-api test. It verifies:

1 - Gateway-api CRDs are installed
2 - Gateway-api CRDs are owned by gateway-api-crd chart
3 - Traefik accepts a configuration change (enabling gateway-api)
4 - Traefik is able to handle gateway-api configuration (HTTPRoute, Gateway, Gatewayclass and ~TCPRoute~)

TCPRoute does not work well with Traefik 3.7. We must wait for 3.8

#### Types of Changes ####

Test

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/docker/gatewayapi/gatewayapi_test.go
+++ b/tests/docker/gatewayapi/gatewayapi_test.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"flag"
+	"strings"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/rke2/tests"
+	"github.com/rancher/rke2/tests/docker"
+)
+
+var (
+	serverCount = flag.Int("serverCount", 1, "number of server nodes")
+	ci          = flag.Bool("ci", false, "running on CI, force cleanup")
+
+	tc *docker.TestConfig
+)
+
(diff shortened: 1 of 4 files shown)


brandond (approved):

tests/docker/gatewayapi/gatewayapi_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+var _ = Describe("Gateway API Tests", Ordered, func() {
+	Context("Setup Cluster", func() {
+		It("should provision servers with Traefik Gateway API enabled", func() {
+			var err error
+			tc, err = docker.NewTestConfig(GinkgoTB())
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(tc.ProvisionServers(*serverCount)).To(Succeed())
```
brandond, line 34: I don't think this step actually waits for the server to be up before returning, so later steps that expect the apiserver to be up or the admin kubeconfig to exist will fail.

brandond (changes requested):

tests/docker/gatewayapi/gatewayapi_test.go
manuelbuil: yes

brandond (approved):

vitorsavian (approved):

PR #11272, opened 2026-09-17
Add ADR for Kata containers

manuelbuil:
#### Proposed Changes ####

This PR adds an ADR about Kata containers integration

#### Types of Changes ####

New feature

#### User-Facing Change ####

```release-note

```


docs/adrs/015-kata-containers-support.md
```diff
@@ -0,0 +1,61 @@
...
+
+Describe in our rke2-docs documentation how to install Kata containers using kata-deploy and add some sections about prerequisites, how to verify it works and how to debug. In the documentation, we should show the HelmChart with the few options that we would support initially.
+
+### (Option 2) Integration using rke2-charts
+
+Add kata-deploy to our rke2-charts as rke2-kata-deploy. Include a new flag in RKE2 that would install the kata-deploy helm chart.
+
+### Other options
+
+There is a process that automatically searches for runtimes, that RKE2 inherits from K3s. We could install the Kata Container binaries and reuse that proecss to set the containerd configuration and the runtimeClass. Given that we have kata-deploy, this option seems like reinventing the wheel and more work for us.
```
brandond, line 39: we should decide on which one of these we are going to do before accepting the ADR.

Since installing directly from the upstream chart seems simple enough - it explicitly supports rke2 already - that seems like the simplest option for now. This would also align with how we "support" the nvidia operator.

If we have customers that want to see deeper integration and specific versions of the chart tied to and tested with each RKE2 release, then bundling the chart (and requiring users to enable it with HelmChartConfig) seems like the correct path.

manuelbuil: 100% agree. I have the docs ready and I run a PoC on how the integration would look like. I'll update the ADR in a sec. As a summary, I would divide this into two phases: 
1 - Docs (same as we do with nvidia gpu operator)
2 - Integation in rke2 (dormant HelmCart)

I'd stay on 1 as long as no customer requires a deeper integration. To be honest, the integration does not bring a lot of benefits.

dereknola (approved):

